Issue #14, opened 2019-01-15
resample_*: Resampling methods


m-mohr:
The resampling methods `resample_spatial` and `resample_temporal` are still very experimental and I honestly couldn't fully remember how we designed them in the VITO sprint. We probably need to check and describe better how the `target` parameter works. We also need to review whether the resampling methods work as expected and are complete, how upsampling could work (especially temporally) and whether the spatial resampling really should link to gdalwarp.

lforesta:
> We probably need to check and describe better how the `target` parameter works

I agree, at the moment I don't know exactly how to create it, maybe with the processes create_raster_cube + add_dimension applied as many times as the dimensions needed? So it seems a user needs to explicitly generate an empty raster-cube in advance of using the resample processes.
Particularly for resample_spatial, maybe it would be easier from a user perspective to specify a target spatial resolution/sampling desired and let the back-end modify the existing raster-cube or create a new one.

m-mohr:
Telco conclusions from today:
We need a new method for resampling based on actual parameters such as the resolution (see gdalwarp). We'll merge it with the new warp_projection methods and could call it resample_spatial. The other process we have at the moment, which spatially resamples from a given datacube to another datacube will be renamed. I think (after telco thoughts), this could be generalized to also allow temporal and other resampling from one data cube to another data cube. In this case the name could simply be resample_cube, resample_target or so.

m-mohr:
Note from @claxn in the warp_projection PR (#28):
> Just a small comment from my side: In my opinion it would be convenient to specify a generic function warp(raster_cube_source, raster_cube_target, resampling_method). As you have mentioned in the meeting today, the datacube object should have all spatial related properties (e.g. extent, projection, pixel spacing...). If the extent is different, you will get the same result as with filter_bbox, if you you want to resample the data (e.g. downsampling 10m->500m), then you change the spatial resolution of the target datacube and if you want to reproject, then you change the projection system of the target datacube. Then we would have only one function instead of multiple. Moreover, a function clone_raster_cube (i.e. copying all properties from a raster cube and creating an empty one) could be helpful in case a user only has to change one property instead of specifying all spatial attributes.

m-mohr:
@claxn It sounds like you are only speaking about spatial warping, so it doesn't affect or allow temporal dimensions for example. Foes the term warping imply spatial?

I like the idea to add a clone_cube process, see #32.

[...]

PR #28, opened 2019-02-11
Added process: warp_projection

### LINKED ISSUE #22: Warp projection - missing process?
I think we currently don't have a process that warps a projection. SInce this is a fairly common operation to do, I would propose to add it to the process list.

lforesta:
I'm not sure which of the existing categories is suitable, I tentatively added a category 'spatial'.
Closes #22

### DIFF
--- a/filter_bbox.json
+++ b/filter_bbox.json
@@ -58,10 +58,23 @@
             }
         },
         "crs": {
-            "description": "Coordinate reference system. EPSG codes must be supported. In addition, proj4 strings should be supported by back-ends. Whenever possible, it is recommended to use EPSG codes instead of proj4 strings. Defaults to EPSG:4326 unless the client explicitly requests a different coordinate reference system.",
+            "description": "Coordinate reference system of the parameters specified as [EPSG](http://www.epsg.org) code or [PROJ](https://proj4.org) definition. Whenever possible, it is recommended to use EPSG codes instead of PROJ definitions. Defaults to `4326` (EPSG code 4326) unless the client explicitly requests a different coordinate reference system.",
             "schema": {
-                "type": "string",
-                "default": "EPSG:4326"
+                "oneOf": [
+                    {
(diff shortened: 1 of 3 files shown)


m-mohr:
Great, thank you. I made some tweaks, but are also not sure about the categories yet. I put it into cubes and aggregate (to be combined with related operations such as resample and aggregate), but the category aggregate needs probably another name as warping is not really an aggregation method. Anybody ware of a better name?

jdries:
looks good, about the description:
Warps the projection of a data cube to the target one
I would say:
Warps the data cube to the target projection

lforesta:
> Warps the data cube to the target projection
I agree, much better than what I wrote.

m-mohr:
Thanks, changed. Also improved the definition of the projection parameter.

claxn:
Just a small comment from my side: In my opinion it would be convenient to specify a generic function `warp(raster_cube_source, raster_cube_target, resampling_method)`. As you have mentioned in the meeting today, the datacube object should have all spatial related properties (e.g. extent, projection, pixel spacing...). If the extent is different, you will get the same result as with `filter_bbox`, if you you want to resample the data (e.g. downsampling 10m->500m), then you change the spatial resolution of the target datacube and if you want to reproject, then you change the projection system of the target datacube. Then we would have only one function instead of multiple. Moreover, a function `clone_raster_cube` (i.e. copying all properties from a raster cube and creating an empty one) could be helpful in case a user only has to change one property instead of specifying all spatial attributes.

mkadunc:
Do we need to better define what kind of datacube is returned? I guess it will have new geospatial axes, while all other axes will remain the same. Do we need to prescribe the coordinates (offset, spacing, extent) of the new axes, or do we allow backend to determine those in a proprietary way?

m-mohr:
We better discuss this in a non-closed issue/PR such as #14.

Issue #81, opened 2019-09-12
Reduce with multiple return values


claxn:
@m-mohr and I struggled finding a clear definition of what happens with the dimension when a reducer returns multiple values. In my opinion the dimension over which is being reduced is either replaced if no `target_dimension`  is given or a new dimension is created. The requested dimension should then contain the outputs of the reducer. E.g. for a datacube with the dimensions `x,y,t,b` and a reducer `extrema` with a target_dimension `p` applied on `t` you should get an updated datacube with `x,y,p,b` as the new dimensions and `min` and `max` as values/indizes along the new dimension `p`. Do other backends agree with this approach or have you implemented it already in a different way?

This brings me to my second question. At the end of the process graph you can save the computed results. Continuing with the example from above it will be necessary to reduce dimension `p` in the end if you want to visualise the `min` OR `max` value for different bands along `b`. But if you want to finally reduce over the bands to be able to visualise the `min` or `max` for one band,  what should happen with the datatype of the remaining dimension? I would suggest that if no dimension with the datatype `band` is available anymore the last dimension gets automatically labelled with the datatype `band`.

m-mohr:
So this seems to be mostly a documentation update.
One thing to clarify: How do we determine the dimension values (min and max in the example above)? extrema returns an array with the min, max values, but other processes have a different semantic.

claxn:
@m-mohr : If the return values of the process are clearly defined I would go for a strict labelling, e.g. `{"min", "max"}`. If not, I would use a range of integer values equal to the length of the returned values.

What I am still not sure about: 
In the end, when you want to save the result, what dimensions are you allowed to specify to visualise the result? I would say it is allowed to specify any dimension, as long as the result is unique, i.e. the data cube has been filtered before. How should we handle (the) spatial dimension(s)?

m-mohr:
@claxn What does "strict labelling" mean? Is the return value of `extrema` an object instead of an array? That is problematic, we don't have processes to work with object. Or do we pass dimension values (min, max) in the reduce process?

I don't get the last issue, but it seems to be not related to this issue and should be discussed separately.

claxn:
[...]

Issue #57, opened 2019-04-26
reduce: required parameter dimension after an optional one


m-mohr:
The `reduce` process has 5 parameters: data, reducer, dimension, target_dimension and binary.
data and dimension are required, dimension and the other two are not. In languages with non-names parameters (JS, PHP, ...) this leads to problems as you have an required parameter after an optional one. That doesn't make any sense, so we may either switch the order and place dimension before reducer (which wouldn't be consistent with other, similar processes, I think) or we need to think about another solution.

m-mohr:
Swapped order of reducer and dimension. Sacrificed consistency over usability.

m-mohr:
Will be reverted, reducer will be required and the null reducer is replaced by the drop_dimension process.

Issue #54, opened 2019-04-05
Adding element to a dimension / keep dimension after reduce


lforesta:
When doing some band mathematics and creating a new band, one can use reduce() and specify the needed formula as in [this example](https://open-eo.github.io/openeo-api/processgraphs/). However reduce() will remove that dimension (spectral in this case).

What if I want to add this new band to the ones already existing in the cube, in this case changing the cardinality of the spectral dimension? Is there a way to do this with the existing processes?

@m-mohr suggested to merge the output of the reduce operation to the original cube, however the newly created cube will not have the dimension over which it was reduced. Following this idea, one could use reduce() and then use add_dimension to re-introduce the lost dimension, then use merge_cubes()?

Is there a better way?

m-mohr:
Yes, after checking the process descriptions I would say it could be solved like this (pseudo code):
```
cube = load_collection(...) // loads a data cube with dimensions x, y, t and spectral for example
// Do filtering etc.
new_band = reduce(cube, callback, 'spectral');
new_band = add_dimension(new_band, 'spectral', 'band_name', 'bands')
cube = merge_cubes(cube, new_band)
// Export data cube etc.
```

Strictly speaking, you don't want to reduce the data / remove a dimension, so reduce doesn't feel like the right process here. So I'm thinking whether apply_dimension with dimension set to spectral could also be a solution. It's not quite covered by the description, but would also seem logical to me.

lforesta:
with apply_dimension the cardinality must not change, so I think the option reduce/add_dimension/merge_cubes is the way to go currently
It seems counter-intuitive to having to reduce a dimension and then add it again and then merge, so maybe in the future we could consider a wrapper-process for these three operations, unless someone has a better approach for this.

m-mohr:
> with apply_dimension the cardinality must not change

Yes, that's what I meant by "It's not quite covered by the description". So in the current version

> the option reduce/add_dimension/merge_cubes is the way to go currently

The question is, how should be be in future? What would you suggest? What process should cover such a use case? A changed apply_dimension? A new process? Name? Description? ...

claxn:
This was discussed some time ago and I am not sure whether this makes sense or if I got something wrong, but wouldn't such a concatenation of the processes above be easier if we don't remove the dimension, unless explicitly told by the user? I don't see any drawback in keeping dimensions with `length(dim) == 1`.

claxn:
[...]

Issue #50, opened 2019-03-01
Rename dimension values?


m-mohr:
In a discussion in #43, we discussed a use case where two cubes are merged. The issue is: What happens if there are collisions e.g. for the band names? Both data cubes could have a band called "B1" and in this case it should be possible to rename one of them before so that there is no collision, right? Could be part of the bigger discussion in #30. Potential names: rename_dimension_value or rename_label.

PR #89, opened 2019-11-18
Simplify reduce, add drop_dimension + rename_labels

m-mohr:
First draft of a reduce rework. Tackles #81, #54 and #57 and #50.


drop_dimension.json
```diff
@@ -0,0 +1,44 @@
+{
+    "id": "drop_dimension",
+    "summary": "Removes a dimension",
+    "description": "Drops a dimension from the data cube.\n\nDropping a dimension only works on dimensions with a single dimension value left, otherwise the process fails with a `TooManyDimensionValues` error. Dimension values can be reduced to a single value with a filter such as ``filter_bands()`` or the ``reduce()`` process. If a dimension with the specified name does not exist, the process fails with a `DimensionExists` error.",
```
soxofaan: > process fails with a `DimensionExists` error

should be `DimensionNotAvailable` as specified lower?

reduce.json
```diff
@@ -60,17 +54,18 @@
...
         },
         "target_dimension": {
-            "description": "The name of the target dimension. Only required if the reducer returns multiple values, otherwise ignored. By default creates a new dimension with the specified name and the type `other` (see ``add_dimension()``). If a dimension with the specified name exists, the dimension is replaced, but keeps the original type.",
+            "description": "The name of the target dimension or `null` to drop the dimension (default).\n\nIf a target dimension is specified, a new dimension with the specified name and the type `other` (see ``add_dimension()``) is created and the old dimension is dropped. If a dimension with the specified name exists, the dimension is replaced, but keeps the original type.\n\nThe resulting target dimension has as many dimension labels as the reducer returned values. For example, a reducer returning a single scalar results in a dimension with a single label, a reducer returning an array with four elements results in a dimension with four labels.\n\nIf a reducer returns multiple value, this parameter is required to be a string. Otherwise the process produces the error `TargetDimensionMissing`.",
```
soxofaan: > If a reducer returns multiple value,

... values

rename_labels.json
```diff
@@ -0,0 +1,161 @@
+{
+    "id": "rename_labels",
+    "summary": "Rename dimension labels",
+    "description": "Renames the labels of a dimension in the data cube. Afterwards, the dimension can be referenced with the new name.",
```
soxofaan: > Afterwards, the dimension can be referenced with the new name.
 
this sounds a bit like the dimension name itself is/can be changed. I think this second sentence is not really necessary.

rename_labels.json
```diff
@@ -0,0 +1,161 @@
...
+        },
+        "dimension": {
+            "description": "The name of the dimension to rename the labels for.",
+            "schema": {
+                "type": "string"
+            },
+            "required": true
+        },
+        "labels": {
+            "description": "The new names for the labels.\n\nIf an **array** is passed, the dimension labels in the data cube are expected to be enumerated so that the dimension labels map to the array indices and get replaced. If the dimension labels are not enumerated, a `LabelsNotEnumerated` is thrown. The number of dimension labels must be equal to the number of array elements, otherwise the error `LabelMismatch` is thrown.\n\nIf an **object** is passed, the object defines a (potentially incomplete) mapping of the source dimension labels to the target dimension labels. If one of the source dimension labels doesn't exist, a `LabelNotAvailable` error is thrown. If a target dimension label exists in the data cube, a `LabelExists` error is thrown.",
```
soxofaan: > dimension labels in the data cube are expected to be enumerated so that the dimension labels map to the array indices and get replaced. If the dimension labels are not enumerated, a `LabelsNotEnumerated` is thrown.

I'm not sure what "enumeration" means in this context.

> If a target dimension label exists

 ... label already exists ...

drop_dimension.json
m-mohr: Thanks, that was too much "copy&paste" from add_dimension.

reduce.json
m-mohr: Fixed

rename_labels.json
m-mohr: Too much copy&paste from rename_dimension. Removed the sentence from both processes.

rename_labels.json
m-mohr: enumerated means zero-based numbering (0,1,2,3,...).
Improved the text accordingly.

reduce.json
```diff
@@ -1,15 +1,15 @@
...
+    "description": "Applies a reducer to a data cube dimension by collapsing all the input values along the specified dimension into an output value computed by the reducer.\n\nBy default the dimension is dropped, but this behaviour can be influenced with the parameter `target_dimension` The parameter must be specified, if a reducer such as ``extrema()`` returns multiple values.\n\nBy default, a reducer must accept an array as input. The process can also work on two values by setting the parameter `binary` to `true`. The reducer doesn't get executed on a single value.\n\nNominal values can be reduced too, but need to be mapped. For example date strings to numeric timestamps since 1970 etc.",
```
mkadunc: - I don't think the part about nominal values needing to be mapped is necessary — the reducer can easily be a function that operates on nominal values, and it can return either nominal or numeric (in case of array input).

- "The reducer doesn't get executed on a single value" seems out-of-place. I'm starting to think that we should move the 'binary' reduce into a new process (e.g. called ['fold'](https://en.wikipedia.org/wiki/Fold_(higher-order_function)) ) and describe the mechanics there (I know this is not the best place to start this proposal - will raise it in the meeting today)

```suggestion
    "description": "Applies a reducer to a data cube dimension by collapsing all the input values along the specified dimension into an output value computed by the reducer.\n\nBy default the dimension is dropped, but this behaviour can be influenced with the parameter `target_dimension` The parameter must be specified, if a reducer such as ``extrema()`` returns multiple values.\n\nBy default, a reducer must accept an array as input. The process can also work on two values by setting the parameter `binary` to `true`. The reducer doesn't get executed on a single value.",
```

m-mohr: Thanks, @mkadunc. I commited your suggestions for point 1. For point 2 I opened a separate issue #94 as we haven't discussed it today.

Issue #85, opened 2019-10-28
Separate binary operators (add, multiply...) from reducers (sum, product...) and revise definitions


mkadunc:
When implementing the reducer processes — `sum`, `subtract`, `multiply`, `divide`, `and`, `or` and `xor` — we realized that both the description of the processes and their usage would benefit from separating the binary case from the aggregate (reducer) case. We have a couple of arguments:

- `ignore_nodata` as default is intuitive and useful in the aggregate case (`a1 + a2 + a3 ...`). However, when evaluating an expression with two distinct values (e.g. `B08 + B04`), as a user I expect a `no_data` input to yield a `no_data` output, and I believe this should be the behaviour for all binary operators and that we don't even need to have the `ignore_nodata` flag in the binary case.

- `divide` and `subtract` should not be reducers — in both, the first element is special and treating repeated division (or subtraction) as an aggregate operation will confuse many users (there's the additional problem of how `no_data` should be ignored when it's the first value in the array) - we've decided against positional arguments in openEO, and we're forcing them back through this weird division and subtraction processes

- `xor` as a reducer should either be removed, or renamed to `oneTrue` or `oneHot` or similar. Repeated application of `⊻` to a list of booleans gives `true` iff the list contains an odd number of `true` values (e.g. `(T ⊻ T) ⊻ T = F ⊻ T = T`), which is different from the current openEO xor definition. This discrepancy could unnecessarily cause problems in implementation of the processes and in their usage.

- results of boolean operations are not well defined when all elements are `no_data` (the text suggests that `and[no_data]` should return `FALSE` which is inconsistent with both the requirement that an empty array should yield `no_data`, and with maths, which suggests that `and[] = TRUE` to satisfy associativity)

- `sum` and `multiply` needlessly require at least two elements, and throw an exception otherwise - the aggregate functions should be useful in a dynamic process graph where the user shouldn't have to guarantee that a certain number of elements will be present (data availability is not predictable, especially in e.g. cloud-filtered time series) - I believe that `no_data` result would be the best compromise for an empty array (mathematically, `sum[]` could also be 0, but if we make it consistent and define `multiply[] ==1` that might be counter-intuitive)

I suggest the following names for the above-mentioned processes:

| binary | reducer  |
|--------|----------|
| and     | [all](https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/all) |
| or       | [any](https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/any) |
[...]

PR #90, opened 2019-11-21
Separate binary operators from reducers and revise definitions

m-mohr:
Fixes issue #85


mkadunc:
Should we name parameters of binary operations consistently? Might be helpful for higher-order functions that would take them as callbacks (fold/binary reduce, sort etc.)...

Right now we use x,y or a,b

m-mohr:
@mkadunc Good point. I thought I had another convention for that, but seems it was always x and y so changed all a and b parameters to x and y.

or.json
```diff
@@ -1,39 +1,38 @@
...
+    "summary": "Logical OR",
+    "description": "Checks if **at least one** of the values is true. Evaluates parameter `a` before `b` and stops once the outcome is unambiguous. If a component is `null`, the result will be `null` if the outcome is ambiguous.\n\n**Truth table:**\n\n```\na \\ b || null | false | true\n----- || ---- | ----- | ----\nnull  || null | null  | true\nfalse || null | false | true\ntrue  || true | true  | true\n```",
```
mkadunc: a,b -> x,y

```suggestion
    "description": "Checks if **at least one** of the values is true. Evaluates parameter `x` before `y` and stops once the outcome is unambiguous. If a component is `null`, the result will be `null` if the outcome is ambiguous.\n\n**Truth table:**\n\n```\na \\ b || null | false | true\n----- || ---- | ----- | ----\nnull  || null | null  | true\nfalse || null | false | true\ntrue  || true | true  | true\n```",
```

subtract.json
```diff
@@ -1,79 +1,64 @@
 {
     "id": "subtract",
-    "summary": "Subtraction of a sequence of numbers",
-    "description": "Takes the first element of a sequential array of numbers and subtracts all other elements from it.\n\nThe computations should follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935) whenever the processing environment supports it. Otherwise an exception must the thrown for incomputable results.\n\nBy default no-data values are ignored. Setting `ignore_nodata` to `false` considers no-data values so that `null` is returned if any element is such a value.",
+    "summary": "Subtraction of two numbers",
+    "description": "Subtracts the second value from the first value and returns the computed result.\n\nNo-data values are taken into account so that `null` is returned if any element is such a value.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it.",
```
mkadunc: I suggest to remove position of parameters from the description

```suggestion
    "description": "Subtracts `y` from `x` and returns the computed result (`x - y`).\n\nNo-data values are taken into account so that `null` is returned if any element is such a value.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it.",
```

subtract.json
```diff
@@ -1,79 +1,64 @@
...
+                    "null"
+                ]
             },
-            "default": true
+            "required": true
         }
     },
     "returns": {
-        "description": "The computed result of the sequence of numbers.",
+        "description": "The computed result the two numbers.",
```
mkadunc:
```suggestion
        "description": "The computed result.",
```

xor.json
```diff
@@ -1,39 +1,38 @@
...
+    "summary": "Logical XOR (exclusive or)",
+    "description": "Checks if **exactly one** of the values is true. Evaluates parameter `a` before `b` and stops once the outcome is unambiguous. If a component is `null`, the result will be `null` if the outcome is ambiguous.\n\n**Truth table:**\n\n```\na \\ b || null | false | true\n----- || ---- | ----- | -----\nnull  || null | null  | null\nfalse || null | false | true\ntrue  || null | true  | false\n```",
```
mkadunc: a,b -> x,y
stopping when the outcome is unambiguous doesn't make sense in this case

```suggestion
    "description": "Checks if **exactly one** of the values is true. If a component is `null`, the result will be `null`\n\n**Truth table:**\n\n```\na \\ b || null | false | true\n----- || ---- | ----- | -----\nnull  || null | null  | null\nfalse || null | false | true\ntrue  || null | true  | false\n```",
```

m-mohr:
Thank you @mkadunc. I addressed all your points.

Issue #91, opened 2019-11-21
Make labels available in reduce, apply_dimension etc.


m-mohr:
We pass only the data to the callbacks in these functions: - `aggregate_polygon`, `aggregate_temporal`, `apply_dimension`, `merge_cubes`, `reduce`, `resample_cube_temporal`. It is useful to also have the labels available, e.g. for the client band math "magic" or more advanced timeseries analysis. We should make the labels available for each value. Could be achieved either with an additional parameter or something like a labeled array data type.

m-mohr:
Seems to be useful and needs to be explored:

1. Whether back-ends can actually provide the data (rasdaman may not be able to do it)
2. How to pass the data to the reducer

m-mohr:
Telco: It seems useful, let's explore it.

m-mohr:
## Idea

1. Define a data type "assoc-array" or "ordered-object" (i.e. a OrderedDict in Python, an associative array in PHP, , Map in JS, not sure about Java). Keys are dimension labels, values are pixel values. There's not JSON equivalent for this, but I don't think this is an issue.
    Example (PHP): `$data = ["a" => 123, "b" => 567]`
2. Allow easy access to it by extending the from_argument object with an index. This avoids heavy use of `array_element` or a similar process.
    Example: `{from_argument:  "data", index: "a"}` to access a in data.
3. Additionally, either allow array_element to be used on this data type (and objects?) or define separate processes.

## Example process graph

Changes:  https://gist.github.com/m-mohr/ec69ca2fc27a003aa3bd78a8e4b512da/revisions 

### Before

```json
{
  "dc": {
    "process_id": "load_collection",
    "description": "Loading the data; The order of the specified bands is important for the following reduce operation.",
    "arguments": {
      "id": "Sentinel-2",
      "spatial_extent": {
        "west": 16.1,
        "east": 16.6,
        "north": 48.6,
        "south": 47.2
      },
      "temporal_extent": ["2018-01-01", "2018-02-01"],
      "bands": ["B08", "B04", "B02"]
    }
  },
  "evi": {
    "process_id": "reduce",
    "description": "Compute the EVI. Formula: 2.5 * (NIR - RED) / (1 + NIR + 6*RED + -7.5*BLUE)",
    "arguments": {
      "data": {"from_node": "dc"},
      "dimension": "spectral",
      "reducer": {
        "callback": {
          "nir": {
            "process_id": "array_element",
            "arguments": {
[...]

PR #97, opened 2019-12-02
WIP: histogram reducer process with discrete buckets

bossie:


m-mohr:
Thank you @bossie  for coming up with a process description for the histogram process.

For me there's room for improvement as it's not clear how exactly the process works and how the bins are computed.

For example, what would be the result for the following input?
* `[-0.6, -0,4, 0,3, 0.8]` (i.e. what happens if the step is not 1?)
* `[0, 55, 255]` (i.e. what happens with gaps?)

In addition, I think it's problematic to return a array of objects. We don't have processes to work on objects and therefore it will be impossible to get values out of the objects. We probably need a pure array approach or improvements to the processes. Issue #91 is highly related.

To solve both issues and make the process description clearer, I guess it could make sense to introduce a parameter that accepts an array of ranges. The indices of that array would be parallel to the returned array so that the process could simply return an array with the counts numbers.

jdries:
We explicitly choose to leave room for how bins are computed, as there are indeed a number of cases. The only possible parameter that we could support is a maximum number of bins.

m-mohr:
We should define it clearly. It's not useful for a user if it's not clear what is actually returned.

jdries:
For cases with a limited number of unique values, like the examples you give above, I would expect having a bucket per unique value.
For cases with number of unique values exceeding a fixed or requested number of buckets, the backend needs to determine the min and max value, and then somehow spread buckets in this range. The most common approach is perhaps to spread them evenly?

m-mohr:
@jdries Specifying a number of bins that are distributed evenly makes sense and would basically explain most of my questions. Question is whether the back-end should determin min/max or whether the user should be able to select it?

We should also add not so simply examples such as [-0.3667421020, 0.2468341124, 1.7006593618 -1.1597762280, 0.5253916115, 0.1204214514]...

jdries:
One of my main use cases for histogram is figuring out the distribution of the data. So I use it in cases where I don't have a clue about min/max values. For this case these 'smart' histograms are very convenient. So the question is, what is the 80% use case? Smart binning with optimized but 'undefined' result, or user defined bins with perfectly reproducable results.

m-mohr:
Not sure. We could also allow both...
If min and/or max are not specified, the back-end determines the missing min/max values.

So we would end up with something like `histogram(data, binCount, min = null, max = null)`? Or should binCount also be optional? How to describe the behavior of the process if binCount is undefined?

m-mohr:
@bossie @jdries Will you be able to update this PR until Jan, 22?

soxofaan:
> How to describe the behavior of the process if binCount is undefined?

numpy/matplotlib uses 10 bins by default. Which is a bit low IMHO however. We could pick a higher default (e.g. a power of two like 16 or 32).

soxofaan:
Another thing not covered here yet is that a user also might want to get the bin bounds themselves. Especially if the user does not specify min/max explicitly.

m-mohr:
@soxofaan I guess you could get the min/max values from the labels 🤔

m-mohr:
Please be aware of the recent changes to the processes, which makes the parameters an array, removes parameter_order and moved the parameter name into the corresponding object.

m-mohr:
Stale PR, please let me know whenever you have updated the process to 1.0.

Issue #98, opened 2019-12-06
coordinate reference system for load_collection


kempenep:
I think we already had several discussions on coordinate reference system (crs).
In my opinion, there should be a crs  parameter in the load_collection. Not only for selecting the region of interest as we do have now, but for defining the target reference system of the resulting cube. Once input images with different crs are selected and need to be combined, we need to be able to re-project them in a single cube. I think it is also best to define the crs as soon as possible in the work flow, which is in the load_collection process, I guess.

m-mohr:
This issue has a long history, see issues #4, #28, #29 and #89. 

Currently, load_collection already covers filter_bbox, filter_temporal, filter_bands and filter. In addition, it would then also cover resample_spatial with additional 4 parameters.
Reason is that it's not just the CRS that needs to match, but also the resolution (and maybe more things?). It is likely that we overload load_collection and it's getting to complicated.

So potential solutions are:
1. Add the 4 additional parameters from resample_spatial to load_collection
2. Expose collections with just a single CRS and/or expose collections matching all potential CRS for the collection.
3. Somehow get this information from the process graph (i.e. resample or save_result calls, but this might be ambiguous).
4. ...?

I must admit I don't like any of the options yet. Ideas? Thoughts?

kempenep:
I believe the spatial (and temporal) resolution need to be defined before actually loading a collection. Even a single collection such as Sentinel-2 can be stored in multiple projection (UTM) zones. Selecting a bounding box on the border of different UTM zones is ambiguous if not defining a target projection. I would be in favor of solution 1 (but I think we should also add the parameters for temporal resolution)

m-mohr:
We need to decide on the call on Thursday whether a process with 10+ parameters is desirable and what alternatives we have. I just covered spatial above, but temporal is indeed also an issue. Also, does this also apply to a process such as load_results or load_user_data (see #83)?

lforesta:
I am not in favor of a 10+ parameters process in general.

Resampling/reprojecting "early" may not always be the best option, since one may end up doing that operation for several hundreds or maybe thousands of timestamps (rasters). Instead the reprojection/downsampling can be potentially done at the end, it should be the user deciding when/if to do that. And I think the user now has all the capabilities to work with the projecton and spatial/temporal sampling that they want. If the user needs to have data in projection EPSG:xxxx, they simply use resample_spatial and/or resample_temporal just after load_collection and before applying any other process.
Are there cases which are not covered by this approach?

[...]

PR #102, opened 2019-12-11
CRS for load_collection

m-mohr:
Here's a proposal for #98 to base discussions upon. Please comment and/or up/down-vote.

### DIFF
--- a/load_collection.json
+++ b/load_collection.json
@@ -313,7 +313,7 @@
     ],
     "exceptions": {
         "TargetCrsMissing": {
-            "message": "The data requested from the collection has multiple CRS assigned and thus the parameter `crs` must be specified."
+            "message": "The data requested from the collection has multiple CRS assigned and thus a valid target CRS must be specified."
         }
     },
     "links": [
--- a/load_collection.json
+++ b/load_collection.json
@@ -1,7 +1,7 @@
 {
     "id": "load_collection",
     "summary": "Load a collection",
-    "description": "Loads a collection from the current back-end by its id and returns it as processable data cube. The data that is added to the data cube can be restricted with the additional `spatial_extent`, `temporal_extent`, `bands` and `properties`.\n\n**Remarks:**\n\n* The bands (and all dimensions that specify nominal dimension labels) are expected to be ordered as specified in the metadata if the `bands` parameter is set to `null`.\n* If no additional parameter is specified this would imply that the whole data set is expected to be loaded. Due to the large size of many data sets this is not recommended and may be optimized by back-ends to only load the data that is actually required after evaluating subsequent processes such as filters. This means that the pixel values should be processed only after the data has been limited to the required extents and as a consequence also to a manageable size.",
+    "description": "Loads a collection from the current back-end by its id and returns it as processable data cube. The data that is added to the data cube can be restricted with the additional `spatial_extent`, `temporal_extent`, `bands` and `properties`. If the data requested from a collection has multiple coordinate reference systems (CRS) assigned, a target CRS for the data cube must be specified with the parameter `crs`.\n\n**Remarks:**\n\n* The bands (and all dimensions that specify nominal dimension labels) are expected to be ordered as specified in the metadata if the `bands` parameter is set to `null`.\n* If no additional parameter is specified this would imply that the whole data set is expected to be loaded. Due to the large size of many data sets this is not recommended and may be optimized by back-ends to only load the data that is actually required after evaluating subsequent processes such as filters. This means that the pixel values should be processed only after the data has been limited to the required extents and as a consequence also to a manageable size.",
     "categories": [
         "cubes",
         "import"
@@ -11,7 +11,8 @@
         "spatial_extent",
         "temporal_extent",
         "bands",
-        "properties"
+        "properties",
+        "crs"
     ],
     "parameters": {
         "id": {
@@ -204,6 +205,40 @@
                 }
             ],
             "default": null
+        },
+        "crs": {
+            "description": "If the data requested from a collection has multiple coordinate reference systems (CRS) assigned, a target CRS for the data cube must be specified. Otherwise, the requested data must either be available in a single CRS or a `TargetCrsMissing` error is thrown.\n\nThe CRS must be specified as [EPSG code](http://www.epsg-registry.org/), [WKT2 (ISO 19162) string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html), [PROJJSON](https://proj.org/projjson.html) or [PROJ definition (deprecated)](https://proj.org/usage/quickstart.html). Defaults to `null` (no target CRS).",
+            "schema": [
+                {
+                    "title": "EPSG Code",
+                    "type": "integer",
+                    "subtype": "epsg-code",
+                    "examples": [
+                        7099
+                    ]
+                },
+                {
+                    "title": "WKT2",
+                    "type": "string",
+                    "subtype": "wkt2-definition"
+                },
+                {
+                    "title": "PROJ definition",
+                    "type": "string",
+                    "subtype": "proj-definition",
+                    "deprecated": true
+                },
+                {
+                    "title": "PROJJSON",
+                    "type": "object",
+                    "subtype": "projjson-definition"
+                },
+                {
+                    "title": "No target CRS",
+                    "type": "null"
+                }
+            ],
+            "default": null
         }
     },
     "returns": {
@@ -221,7 +256,7 @@
     },
     "examples": [
         {
-            "description": "Loading `Sentinel-2B` data from a `Sentinel-2` collection for 2018, but only with cloud cover between 0 and 10%.",
+            "description": "Loading `Sentinel-2B` data from a `Sentinel-2` collection for 2018, but only with cloud cover between 0 and 10%. Data cube will be projected to EPSG:3857.",
             "process_graph": {
                 "loadco1": {
                     "process_id": "load_collection",
@@ -268,13 +303,19 @@
                                     }
                                 }
                             }
-                        }
+                        },
+                        "crs": 3857
                     },
                     "result": true
                 }
             }
         }
     ],
+    "exceptions": {
+        "TargetCrsMissing": {
+            "message": "The data requested from the collection has multiple CRS assigned and thus the parameter `crs` must be specified."
+        }
+    },
     "links": [
         {
             "rel": "about",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `array_find`
     - `drop_dimension`
     - `rename_labels`
+- Parameter `crs` added to `load_collection`. [#98](https://github.com/Open-EO/openeo-processes/issues/98)
 - Added further examples
 
 ### Changed


jdries:
Not a big fan of this proposal, I prefer that users and process graphs do not enforce a CRS, unless the use case strictly demands it. 
There is also not really a concept of a 'multi-crs' collection, in the sense that it is not exposed in the metadata. So users might be tempted to start defining crs's everywhere 'just in case', at the expense of performance and cost.

m-mohr:
> There is also not really a concept of a 'multi-crs' collection, in the sense that it is not exposed in the metadata.

That's not correct, the API supports it since version 0.4.

lforesta:
I agree with @jdries that performance may be badly affected. Many users will be tempted to specify the CRS without a proper understanding of the implications (whereas if they explicitly use the resample_spatial process at any point in the pg, they are conscious about their choice).

If we go for it, the CRS should be null by default (as in the current PR), because we should also consider that EO includes lots of data still in sensor geometry (e.g. S1 and S3 L1 data), and simply specifying a CRS at load_collection will not suffice in this case (maybe relatively few users will need to work with this level of data, but many scientists do research on the L1 to L2 processing step).

m-mohr:
> I agree with @jdries that performance may be badly affected. Many users will be tempted to specify the CRS without a proper understanding of the implications (whereas if they explicitly use the resample_spatial process at any point in the pg, they are conscious about their choice).

We can simply add a note to the description of the parameter so a user is aware of the issue.

load_collection.json
```diff
@@ -268,13 +303,19 @@
...
+                        "crs": 3857
                     },
                     "result": true
                 }
             }
         }
     ],
+    "exceptions": {
+        "TargetCrsMissing": {
+            "message": "The data requested from the collection has multiple CRS assigned and thus the parameter `crs` must be specified."
```
lforesta: For Sentinel-2 collections (which on many back-ends have multiple CRS defined), one CRS MUST be specififed even if the default CRS can be a null? I'm confused by this sentence.

m-mohr: @lforesta You must specify a non-null target CRS for your Sentinel-2 example.
Maybe the message should be as follows?
```suggestion
            "message": "The data requested from the collection has multiple CRS assigned and thus a valid target CRS must be specified."
```

lforesta: So a CRS can be null only when a unique is present in the collection's metadata and will be used if no other is specified, right? Ok now it's clearer, but I am still not convinced about forcing a unique CRS for the output.

Suppose a group working on algorithms which are entirely scene based (i.e. processing rasters independently); with S2 they might be happy (or even prefer) that the output has the same CRS as the original raster, but with this approach, to achieve their goal they would need to submit one job for each CRS.

m-mohr: @lforesta There are no scenes (= granules) in the openEO world, but if a back-end gives users information on scenes and users would want to work on them individually, then I would expect that each scene is an individual data cube. That means crs doesn't need to be set in load_collection (1 scene = 1 crs) and they can work with the original crs until they want to merge and need to resample to a common crs. You can easily use multiple load_collection in a process graph and re-use algorithms with stored process graphs, so it's *not* necessarily one job for each crs. 

The back-end only throws the error if the data requested hast multiple crs assigned. If you load data from a single crs although the collection has multiple crs assigned, you don't need to throw an error. Only if you load data over a region that has multiple crs assigned, you need to make clear what the target crs is. It just needs to be unambiguous.

I think this makes sense, no?

lforesta: Ok never mind the scenes/granules. If my processing is pixel based (i.e. every pixel can be processed completely independently from others), it does not matter that my datacube has multiple CRS or one only. Clearly I'm thinking of a loose datacube, but I think there are people who work with applications of this type. For example, groups doing research on processors from L1 to L2, often they process data on a pixel level. Say I'm one such group and I want to create my S2-L2 product for an entire continent by using an openEO compliant back-end. I could do:
```
load_collection(S2, bbox=full-continent)
apply(my_UDF) # the UDF may be a set of standard processes
save_raster
```

If I'm forced to specify a unique CRS for this example, the back-end will have to reproject lots of data, although the group in this case is maybe happier to have the original UTM tile structure preserved (data like this is then used by other groups to do further processing, they will make sure it's all in the projection they want to use).

In this example, to achieve the same without reprojecting, either I create one job per CRS or within the same job I call load_collection multiple times with different bbox defined, right? This seems a lot of overhead for "keeping the original projection".

Anyway, maybe this is a corner case and in any case there are ways to create jobs where the original multiple CRSs are kept (paragraph above). I just wanted to point out example(s) where this PR may not be optimal.

m-mohr: I'm not sure how your example would work with save_result. How would a data cube of a whole continent be stored? When saving, the save_result process doesn't know any longer the original files and thus can't save it to the original tile structure. Maybe some back-ends can do that, but user can't expect that to happen on all back-ends. So this use case doesn't even work without having this PR merged, I think. I don't think your use case can currently be tackled very well in openEO. I guess we would need a process that can load individual tiles into individual data cubes for that, e.g. load_tiles() which returns an array of data cubes and then we'd need an array_each process so you can apply a "callback" on each data cube and save them individually.

m-mohr: Hey @lforesta, I think this is an interesting discussion. To not forget about it later, I moved it to #108. Please follow-up there regarding this tile-based use case.

edzer:
@jdries and I had a long discussion on this issue in the train on Wednesday, and may have come to a possible resolution. The problem is that `load_collection()` shall return a data cube, that a data cube with x and y dimensions shall have an unambiguous from index to coordinates, and that this is not the case when tiles in a collection span several UTM zones. Pragmatic solutions to this (picking the CRS at the center of the extent, picking the single CRS when the extent only spans a single CRS) may sometimes work but not always. Of course for downloading a GeoTIFF or viewing a WMTS a CRS needs to be set, but the use case where someone wants to compute (and save) NDVI (or something much more expensive) in the native pixel / tile space spanning several CRS's is valid, and should not be cut off by requiring setting a single CRS (and working on it) here.

The way out I propose here is to make CRS a dimension.

This way, x and y map into the same coordinates for different CRS, and the CRS slice you're in determines how they map to world coordinates. This way, you can do everything on spectral and time dimensions, and simply loop over x, y and CRS with no need for resampling.

Certain spatial things that only work on the local (relative) context can also be done, think of a low-pass filter. Other spatial things, like aggregate_spatial with a set of polygons, need an unfolding of the CRS dimension into a mosaic with a single CRS before they make sense, or be done in an apply over the CRS dimension (in which case polygons crossing UTM zones will not result in a single result). 

A similar idea of having CRS as a dimension would be having tile ID as a dimension: x and y indexes are the row/cols and then map into coordinates relative to the tile origin, with similar consequences. 

@mkadunc what do you think?

m-mohr:
Hm, interesting idea. We need to think about how that would work with our processes. What do we need to change in resample_spatial / resample_spatial_cube and how would for example a reduce on the crs dimension work?

edzer:
I'm planning to write a bit of a longer piece about this, looking into the topics you mention.

mkadunc:
Having CRS as a dimension makes a lot of sense, especially if it is treated as one of the spatial dimensions.

One concern I have is that adding CRS dimension might unnecessarily burden the user with this extra "singleton" dimension - just the fact that it's there is not a problem, but it can become one if you need to take it into account on almost each node of the process graph, even when the operations are not spatial. Maybe we should do some experiments with extra singleton dimensions and see how much they impact working with data cubes - if there's no impact we're OK, but if it's significant, we could find ways to implicitly take such extra dimensions into account without burdening the user.

---

*Side note: There is another (a bit more involved) option, where CRS would not necessarily be a dimension - if we extended the concept of axis labels into a more general concept of "cell-attributes", then we could just require that every spatial cell has a non-null `crs` attribute: the value of this could come from:*
- *the root of the cube (when the whole cube is a single CRS)*
- *the separate CRS axis (when there are different CRSs and not aligned with other axes)*
- *some other axis (when CRS can be derived from some other axis, as in the tileId-axis example)
In NetCDF, one would solve this with an extra `variable` (whose value is defined just once as a scalar or along a subset of axes), but since we decided to only hold a single value in a cube, we can't use the same mechanism.*

*The cell-attributes concept has additional uses, e.g.:*
- *`absoluteOrbitId` can be used as a dimension, and its labels can provide the well-known `time` attribute so it can function as a temporal axis*
- *`productId` as a dimension could, in addition to time and crs, provide per-scene info about the platform (`S2A` / `S2B`), processing version and other scene-related metadata (sun angles, viewing angles etc.)*

*This are maybe all things that could be handled with extra dimensions, but I think that it is more intuitive to think about dimensions, especially in the data cubes world, as "the minimal independent set of cell attributes that unambiguously address a single value" (in relational world these would be called ["candidate key"](https://en.wikipedia.org/wiki/Candidate_key)) — adding extra dimensions just to provide some metadata to the cells feels wrong.*

m-mohr:
We (@flahn and me) went a bit crazy discussing this proposal and have some more thoughts:

1. The proposal with the crs parameter in load_collection is the simplest from the user perspective. This would be very similar to what GEE offers. There are obviously quite a bunch of users that want this simplicity.
2. The proposal with crs as Data Cube is not easy to understand, I think. A GEE user will probably not want to think about how to reduce a crs dimension and just stick with the simpler GEE approach. Scientists who strictly want to work on the original data/crs will probably like this approach.

So could we combine these two approaches? Could we say that if crs is not required and only if the collection has multiple CRS and the crs parameter in load_collection is not set, then a user gets a crs dimension?

The question really is: Who is the target audience? Who do we cater for? Which audience would lead to most uptake? etc.

Another thought was to allow something like this (in JS-like pseude code):
```
epsgCodesArray = get_metadata(field = 'eo:epsg') // Get all epsg codes from metadata
dataCubesArray = epsgCodesArray.apply(epsgCode => {
  dataCube = load_collection(id = "Sentinel-2", temporal_extent = null, spatial_extent = null, properties = eq('eo:epsg', epsgCode)) // Load all data that is stored in this epsg code
  // further process data cubes
})
var mergedDataCube = dataCubesArray.reduce((cube1, cube2) => {
  return resample_spatial_cube(cube1, cube2);
}, binary = true)
save_result(mergedDataCube)
```
That's mostly possible already. It basically returns one data cube per crs defined for the collection, all of them just containing the data that is available for that specific crs. Afterwards process them individually and merge them using a binary reducer if wanted.

load_collection.json
```diff
@@ -268,13 +303,19 @@
                                     }
                                 }
                             }
-                        }
+                        },
+                        "crs": 3857
```
mkadunc: I wouldn't put the `crs` parameter in the example, as it would suggest that this is something that should always be used - users should rather be encouraged to leave the CRS choice to the backend.

load_collection.json
```diff
@@ -268,13 +303,19 @@
...
                     },
                     "result": true
                 }
             }
         }
     ],
+    "exceptions": {
+        "TargetCrsMissing": {
+            "message": "The data requested from the collection has multiple CRS assigned and thus a valid target CRS must be specified."
+        }
```
mkadunc: Even when the data has multiple CRSs assigned, the backend can probably do a better job of selecting the more appropriate option than the user. Forcing the backend to fail when CRS is not provided is IMO counter-productive.

m-mohr:
Closing, we are going the route described in #98.

Issue #105, opened 2019-12-13
import_*: Import data from non-API sources


m-mohr:
This originates from #83: There are probably several processes, which could be useful to import data from "non-API" sources, similarly to `load_uploaded_files`. Some ideas:
* `import_s3`: Load data from Amazon S3 (see [example in #83](https://github.com/Open-EO/openeo-processes/issues/83#issuecomment-540573097))
* `import_nfs`: Import from a network file storage system attached to the back-end. Example from VITO: [load_disk_data](http://openeo.vgt.vito.be/openeo/0.4.0/processes/load_disk_data)
* `import_stac`: Import from a STAC catalog/API
* `import_external_result`: Import a openEO job result from another back-end
* and more (GCS, Azure, ...)

Similarly, there could be export_* processes at some point.

lforesta:
I like the idea. Maybe we could write explicitly in their description that these processes may require a significant (and possibly slow) data transfer. This may not be clear to non-technical users.

m-mohr:
More or less encapsulates too many different new processes. There are some separate issues for such processes, e.g. #322

PR #106, opened 2019-12-13
load_uploaded_files

m-mohr:
Added load_uploaded files process.

Notes:
- Input file formats are specified in /file_formats
- Process to transform files into data cubes must be described in the /file_formats descriptions.
- Should `format` be required or be optional and if not specified the back-end can try to guess based on the file_extensions?
- For API changes see commit https://github.com/Open-EO/openeo-api/commit/c8eab70147addff9d46e8e15dd7bd3b90e024f6d


load_uploaded_files.json
```diff
@@ -0,0 +1,65 @@
...
+							"type": "string",
+							"subtype": "folder-path"
+						}
+					]
+				}
+			},
+            "required": true
+        },
+        "format": {
+            "description": "The file format to save to. It must be one of the values that the server reports as supported input file formats, which usually correspond to the short GDAL/OGR codes. This parameter is *case insensitive*.",
```
jdries: Shouldn't this be 'file format to read from', or am I missing something?

m-mohr: Oh, yes, thanks. I coped from save_result and forgot to change that part. Fixed.

load_uploaded_files.json
```diff
@@ -0,0 +1,65 @@
+{
+    "id": "load_uploaded_files",
+    "summary": "Load files from the user workspace.",
+    "description": "Loads user-uploaded files from the local user workspace / data store. The files must have been stored by the authenticated user on the back-end currently connected to.",
```
soxofaan: The process name, description, ...  talk about plural "files", while a lot of use cases will probably be about single files (e.g. a polygon masking). An even in the multiple file case I think it is worth mentioning in the description that all the data is  combined and returned as a single data cube.

load_uploaded_files.json
```diff
@@ -0,0 +1,65 @@
...
+        "import"
+	],
+    "parameter_order": [
+        "paths",
+        "format",
+        "options"
+    ],
+    "parameters": {
+        "paths": {
+            "description": "The file(s) and folder(s) to read.",
```
soxofaan: maybe just "The files or folders to read"?

soxofaan:
how is loading of a folder defined? loading all files in the folder? only files matching the format?

m-mohr:
@soxofaan I tried to incorporate your feedback. I tried to make it as useful as possible, but I'm not so sure about that part: 
> Implementations may skip files in folders that clearly can't be read using the specified format

soxofaan:
Another minor note: `file-path` and `folder-path` from Open-EO/openeo-api@c8eab70  explicitly state "relative path to a user-uploaded file". It might be useful to relax this a bit and allow absolute paths as well, with the caveat that this assumes backend-specific knowledge. This would make sense for some use cases at VITO. Alternatively this could also be covered by  suggestions from #105

m-mohr:
@soxofaan No, I don't think this is a good idea. This process is really only for loading files from the user-uploaded files, uploaded using the /files endpoints and there's no absolute path for them (one could argue the root of each users workspace is / though), but I guess it's simpler to restrict to relative?! If you want to load data from other sources, see #105.

soxofaan:
> but I'm not so sure about that part:   Implementations may skip files in folders that clearly can't be read using the specified format

Indeed, folder support adds a bunch of complications. Maybe we could start with just providing a `file-path` only version of this process and keeping the folder support feature for further discussion and follow-up PR? I also haven't seen enough use cases to estimate the value of folder support at the moment.

m-mohr:
My reason to add it was for example a (larger) set of GeoTIFF files I want to load. Could be hundreds to be combined in a single data cube. Listing them all individually is a pain, but the user experience could probably be somewhat improved by clients.

Adding the paragraph about skipping files, I'm thinking about skipping accompanying metadata files (e.g. if you upload a STAC catalog).

What other difficulties you are thinking of, @soxofaan ?

soxofaan:
As you hinted yourself in the process documentation, folder support introduces some aspects that complicate the API and backend implementation. Over time one might even want to expose these things as options to the user:

- recurse in subfolders or not? or skip particular ones?
- filter out some files (black listing)
- select files (based on format guessing, file extension, file globbing, ...)
- fail on unreadable files or skip them?

A file-only API is more explicit and avoids a lot of the above issues.
And indeed, clients are a good place to alleviate the pain of explicit file listing.


That being said, I was coming to this issue from a "single file" use case, so folder support seemed overkill. If you indeed are thinking of "lots of files" use cases, folder support makes sense. On the other hand, wouldn't there "lots of files" use cases be better served by the `import_` ideas of #105  ?

m-mohr:
@soxofaan Okay, I'm fine with a simpler specification, so I removed folder support from API and processes, but added that "Clients should assist to generate a list of files for folders." to the API.

> On the other hand, wouldn't there "lots of files" use cases be better served by the `import_` ideas of #105 ?

No, "import" and "load" refer to different data sources, not the amount of data loaded. "import" is for loading from non-API sources, "load" is for loading from API sources (e.g. from /collections and /files). For example, if we'd go for a "load from folder" process separately, I'd probably name it "load_uploaded_folder" or so.

soxofaan:
> No, "import" and "load" refer to different data sources, not the amount of data loaded.

I understand, but I meant it more like: if you want to use openEO on a lot of (possibly large) files of your own, maybe the openEO "upload"+"load" feature is not the best approach and you are better served with an approach that uses external storage (like S3) that you then "import".

m-mohr:
Sure, this could be a better alternative if supported by the back-end.

If the proposal if fine now, I'd appreciate an approval for the PR. :-)

Issue #62, opened 2019-06-05
Make aggregate_polygon more general


kempenep:
https://github.com/Open-EO/openeo-processes/blob/cee65f77dece8568e350e65a67c9761a254e1aca/aggregate_polygon.json#L1

I would like to suggest to rename the process aggregate_polygon to aggregate_spatial
Some motivation:
1) more in line with aggregate_temporal, where we aggregate in spatial rather than temporal domain
2) polygon parameter should not necessarily be of "polygon" type. Different type of features can be used. Even point features might be useful here. 

An extra parameter "buffer" could also be useful (positive: polygon gets extended, negative: polygon is reduced ). In case of point features, a positive buffer allows to calculate zonal statistics around the specified points

m-mohr:
This is made to align with filter_polygon, because we are also discussing to add filter_point etc. There could also be aggregate_point then. They may have different semantic and parameters.

The buffer parameter is another issue that should be in a separate issue.

kempenep:
My preference would still be the more generic "aggregate_spatial", as it does not restrict to a single feature type. For instance, we might want to aggregate SQLite vectors that combine different feature types (polygons, points, ...) in different layers.

m-mohr:
Could you please propose a concrete and complete process definition according to our /processes endpoint, incl parameters and schemas for it? 

If we change, it we should also tackle filter_spatial and #37.

kempenep:
I do not seem to have write permission to create new feature branch:
remote: Permission to Open-EO/openeo-processes.git denied to kempenep.

Attached the process definition candidate

Op vr 7 jun. 2019 om 15:14 schreef Matthias Mohr <notifications@github.com>:

m-mohr:
Attachment via mail didn't get through, unfortunately. You can get write permissions or fork an PR or use the GitHub Web interface to upload to this issue.

lforesta:
Telco conclusion
create a new process "aggregate_spatial" that accepts any type of feature, including point
deprecate current "aggregate_polygon"
@kempenep to create PR

m-mohr:
I'm curious to see in that PR how aggregations for points and line strings work. Without defining those cases clearly, it doesn't make sense to change the processes at all.

lforesta:
it basically would be an interpolation at that point or the points on the line

m-mohr:
Thanks. In that sense, we could also apply the more general rule also to mask, filter_polygon and load_collection.

kempenep:
[...]

Issue #94, opened 2019-11-26
Split binary and array based reduction?


m-mohr:
@mkadunc wrote in #89:
> I'm starting to think that we should move the 'binary' reduce into a new process (e.g. called 'fold' ) and describe the mechanics there (I know this is not the best place to start this proposal - will raise it in the meeting today)

It seems we forgot to discuss it today, but is a good topic to discuss next time. Still add your comments here in the meantime.

mkadunc:
Regarding the name: [fold](https://en.wikipedia.org/wiki/Fold_(higher-order_function)) is what Wikipedia uses, and is also the most frequent name for this function in languages that do not use "reduce". I don't know of any uses of "fold" in programming that would be inconsistent with the behaviour that what we're after.

Note: this issue may be viewed differently depending on what we decide about merging the multi-output-reduce and apply_dimension (see https://github.com/Open-EO/openeo-processes/issues/73 )

Reasons why I think splitting would make sense:

- we have a lot of difficulty with defining the behaviour and purpose of `reduce` even when considering just the variant where callback takes an array

- we have not yet discussed and considered in as much detail the binary variant, I suspect if/when we do, we will find the current description deficient in many ways

- imo `reduce` would be easier to understand for the users if there was clearly a single mode of operation, with corner cases and examples all dedicated to that same mode

- the mechanics of binary reduce (fold) are sufficiently different from the mechanics of array reduce that calling one or the other is not something that the user would be uncertain about

- the 'binary' parameter is so strongly coupled with the 'reducer' parameter that changing the operation mode from one to the other should only be done very deliberately — the kinds of changes this requires for the callback feel reasonable when changing the process, but more difficult to justify when flicking a single boolean parameter

- by packing a lot of fundamentally different functionality into a single process we're risking that the users and implementors will make mistakes and wrong assumptions — good APIs and SDKs should try to protect users from such risks

m-mohr:
@mkadunc I like the idea as it makes the processes more streamlined as you said. I'm not sure whether a completely new name makes sense as the binary variant still will be quite similar to the array reduce, so I think reduce_binary could also be an option. One important thing to keep in mind is that there are more processes with the binary option: `aggregate_polygon`, `aggregate_temporal`, `resample_cube_temporal` and `merge_cubes`. We need to think about what to do with them and if we split up those, too, it probably makes even more sense to just name them xxx_binary?!

m-mohr:
[...]

Issue #37, opened 2019-02-22
Filter by point process


GreatEmerald:
We discussed in the telco that it might make sense to have a process to filter by point, like we have a filter by polygon.

The use case we're thinking of is to quickly get all relevant data about a particular point in the world. What that entails is not entirely clear though and probably depends on what filtering by polygon does, so as to maintain consistency. Right now filtering by polygon effectively is defined as a `crop`/`clip` function, and the same could be done for filtering by point.

There are three more alternatives: to have it work as an `intersect`, so that dimensions of features or rasters are not changed but anything not touching (i.e. not relevant to) the point is discarded (that's how it works in GEE); to have it work as an `extract`, so that the space dimensions are collapsed (you get e.g. a time series); or to not have either of the filtering options altogether and just provide `clip`/`crop`, `intersect`, and `extract` as their own processes.

jwagemann:
For climatological applications, point retrieval is an important process. It is currently still too cumbersome to retrieve point information over long time-series data.

jdries:
What is a point exactly in a raster data cube context? A mathematical point, specified by geographic coordinates, can still be positoned on the point where two or more pixels (data cube cells) touch. In some use cases, points are also converted into square or circular bounding boxes, with a specific size. This tends to make the desired result more reproducable.

edzer:
Good points, we need to define all this somewhere. I think the usual (= GDAL) semantics about this is that one corner point and two sides belong to the pixel, the other sides / corners not. If y index increases downward (y cellsize negative) the top left corner and left and top sides belong to the pixel, so a point on this corner / these points intersect with the pixel; this way a grid is different from a set of polygons polygonizing the pixels.

As I mentioned in the telco, querying a raster data cube on a set of points is not an openEO filter operations, as all filter operations return a raster cube, and this returns a vector data cube (points x time). The closest in our typology is [aggregate_polygon](https://open-eo.github.io/openeo-api/v/0.4.0/processreference/#aggregate_polygon), which takes a raster data cube and a vector data cube, and returns a vector data cube. The aggregation function (reducer) would be the identity function. It's somewhat contrived, and we may want to pick a different name for this; `resample`, or even `sample` comes to mind.

GreatEmerald:
[...]

PR #107, opened 2019-12-13
aggregate_spatial

kempenep:
created new process aggregate_spatial as agreed during openEO process meeting on 20191212. The process aggregate_polygon will then be deprecated

Issue: #62


aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
+{
+    "id": "aggregate_spatial",
+    "summary": "Compute zonal statistics for a vector data cube",
```
m-mohr:
```suggestion
    "summary": "Compute zonal statistics for geometries",
```
GeoJSON is a potential argument, but not a data cube itself. So I would propose to make it more generic. Either "for vector data" or "for geometries".

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
...
+        "data": {
+            "description": "A raster data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            },
+            "required": true
+        },
+        "vector": {
+            "description": "A vector data cube containing features on which the aggregation will be based.",
```
m-mohr:
```suggestion
            "description": "Geometries on which the aggregation will be based. Either specified as GeoJSON or vector data cube.",
```

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
...
+    "summary": "Compute zonal statistics for a vector data cube",
+    "description": "Aggregates zonal statistics for features within a vector data cube (e.g., one or multiple points or polygons) over the spatial dimensions.\n\nIn case of polygon features, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
+    "categories": [
+        "cubes",
+        "aggregate & resample"
+    ],
+    "experimental": true,
+    "parameter_order": [
+        "data",
+        "polygons",
```
m-mohr:
```suggestion
        "geometries",
```
or `vector` to fix the build.

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
...
+    "parameters": {
+        "data": {
+            "description": "A raster data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            },
+            "required": true
+        },
+        "vector": {
```
m-mohr:
```suggestion
        "geometries": {
```

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
+{
+    "id": "aggregate_spatial",
+    "summary": "Compute zonal statistics for a vector data cube",
+    "description": "Aggregates zonal statistics for features within a vector data cube (e.g., one or multiple points or polygons) over the spatial dimensions.\n\nIn case of polygon features, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
m-mohr:
```suggestion
    "description": "Aggregates zonal statistics for one or more geometries (e.g. points or polygons) over the spatial dimensions.\n\nThe process considers all pixels for which the point at the pixel center intersects with the corresponding geometry (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
Trying to make it more consistent, also adding in a precise definition which pixels to consider for points and line strings.

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ],
+            "required": true
+        },
+        "reducer": {
+            "description": "A reducer to be applied on all values of each geometry. The reducer must be a callable process (or a set of processes as process graph) such as ``mean()`` that accepts by default array as input. The process can also work on two values by setting the parameter `binary` to `true`.",
```
m-mohr:
```suggestion
            "description": "A reducer to be applied on all values of each geometry. The reducer may be a single process such as ``mean()`` or consist of multiple sub-processes. By default, a reducer must accept an array as input. The process can also work on two values by setting the parameter `binary` to `true`.",
```
We are not using process graph when communicating with users. Also, this wording has been used throughout the processes in several places and if we want to improve the wording, we should change it everywhere, not just here.

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
...
+            "schema": {
+                "type": "boolean"
+            },
+            "default": false
+        }
+    },
+    "returns": {
+        "description": "A vector data cube or a GeoJSON object, depending on the input of the `vector` parameter.\n\nThe computed value is stored in a property (GeoJSON) or dimension (vector cube) with the name that was specified in the parameter `name`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) in each geometry. In GeoJSON these are stored as properties with the names `{name}_total_count` and `{name}_valid_count` (replace `{name}` with the value of the `name` parameter). In a vector data cube, these values are stored as attributes of the result value with the attribute names `total_count` and `valid_count`.\n\nIf the input was GeoJSON and the therefore the return value is also a GeoJSON object, the geometries (`Polygon` or `GeometryCollection`) get wrapped in a `Feature` or `FeatureCollection` respectively. The results of the computations are stored in the `properties` of each GeoJSON `Feature`.",
```
m-mohr:
```suggestion
        "description": "A vector data cube or a GeoJSON object, depending on the input of the `geometries` parameter.\n\nThe computed value is stored in a property (GeoJSON) or dimension (vector cube) with the name that was specified in the parameter `name`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) in each geometry. In GeoJSON these are stored as properties with the names `{name}_total_count` and `{name}_valid_count` (replace `{name}` with the value of the `name` parameter). In a vector data cube, these values are stored as attributes of the result value with the attribute names `total_count` and `valid_count`.\n\nIf the input was GeoJSON and the therefore the return value is also a GeoJSON object, the geometries (`Polygon` or `GeometryCollection`) get wrapped in a `Feature` or `FeatureCollection` respectively. The results of the computations are stored in the `properties` of each GeoJSON `Feature`.",
```

kempenep:
Thanks for the suggestions, agreed. The filter_polygon was already discussed during the meeting, where we decided to proceed as you suggested. This was going to be taken care of by ... (don't remember who)...

m-mohr:
@kempenep Do you want me to commit all the changes?

I think filter_spatial was assigned to @GreatEmerald.

I'm wondering whether filter/aggregate_geometries would be more intuitive to some users? Especially thinking about meteorology where I heard they would be confused about aggregate_spatial.

kempenep:
You can commit all changes. Don't know which of spatial/geometries is to be preferred. I like "spatial" as it is in line with the "temporal" we already have.

m-mohr:
@kempenep I don't have write permissions to your repository ;-) Can you batch commit them via the FIles tab here at GitHub, please?

kempenep:
@m-mohr 
https://github.com/kempenep/openeo-processes/invitations

m-mohr:
We still need to deprecate aggregate_polygon, other than that the PR looks good now.

jdries:
Is the behaviour for lines and points already defined?
Or is it this line:

> The process considers all pixels for which the point at the pixel center intersects with the corresponding geometry (as defined in the Simple Features standard by the OGC).

m-mohr:
> Is the behaviour for lines and points already defined?
> Or is it this line:
> 
> > The process considers all pixels for which the point at the pixel center intersects with the corresponding geometry (as defined in the Simple Features standard by the OGC).

@jdries Yes, it's that line.

jdries:
For points, and lines, I would consider an intersection of the pixel bounding box with the geometry, instead of considering only pixel center.
Alternative approach could be something like: in case of lines/points intersection geometry is buffered using half the size of a pixel (hoping that pixels are square), and than the aggregation reduces to aggregating over a polygon.
(That alternative could also be a solution for users in case the backends simply does not support lines and points.)

m-mohr:
And that's why I think separate processes are better suited for different geometry types and initially restricted it to polygons only... If you need to handle the geometry types differently, make different processes otherwise the process gets to difficult.

jdries:
I agree on that, and here are two more reasons:

- aggregation over point or line is way more exotic, so backends may not support it. Having separate processes would allow those backends to make that clear
- avoiding unneeded backwards incompatible changes in the process definitions

m-mohr:
Indeed, the wording regarding the pixel center works only well with polygons. And for points it should consider the pixel (the whole, not just the center) that intersersects.

But that's why I think separate processes are better suited for different geometry types and initially restricted it to polygons only... If you need to handle the geometry types differently, make different processes otherwise the process gets to difficult. We'll see how it works out also in filter_spatial.

lforesta:
So the decision now is between having this process or separate ones? Maybe we need to discuss this in the next telecon again..

m-mohr:
Yes, I think so, too.

GreatEmerald:
I don't see why we'd handle it any differently for different types of geometry. If the geomrety touches a pixel in any way, the pixel is included.

Alternatively, like in R, add an optional argument that says whether to include anything that touches, or anything that the geometry envelops, or any pixels whose centres the geometry envelops. That would make it more flexible. For line or point features, enveloping implies being at the exact same location (so if it's right on the pixel centre and the user asks for enveloping, it's OK, otherwise the result is empty)

kempenep:
I agree with Dainius here. I am also in favor of the optional argument, where I would adopt the GDAL terminology (as in gdal_rasterize):
ALL_TOUCHED:  all pixels touched by lines or polygons will be updated, not just those on the line render path, or whose center point is within the polygon.  Defaults to disabled for normal rendering rules.

m-mohr:
> I don't see why we'd handle it any differently for different types of geometry. If the geomrety touches a pixel in any way, the pixel is included.

That sounds reasonable and would work consistently across all geometry types.

> Alternatively, like in R, add an optional argument that says whether to include anything that touches, or anything that the geometry envelops, or any pixels whose centres the geometry envelops. That would make it more flexible. 

But also more complex to implement. Maybe we can start with the simpler case above and then see at which point we need the additional parameter.

aggregate_spatial.json
```diff
@@ -0,0 +1,130 @@
+{
+    "id": "aggregate_spatial",
+    "summary": "Compute zonal statistics for geometries",
+    "description": "Aggregates zonal statistics for one or more geometries (e.g. points or polygons) over the spatial dimensions.\n\nThe process considers all pixels for which the point at the pixel center intersects with the corresponding geometry (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
m-mohr:
```suggestion
    "description": "Aggregates zonal statistics for one or more geometries (e.g. points or polygons) over the spatial dimensions.\n\nThe process considers all pixels which touch the corresponding geometry (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```

This would mean that in some cases pixels are considered multiple times in aggregations, e.g. counting all pixels in Germany and The Netherlands would in most cases count all pixels on the border for both Germany and The Netherlands.

soxofaan: FYI: At VITO we have a use case where the possibility of pixels being in multiple aggregation buckets simultaneously result in important performance/implementation  issues.

mkadunc: See https://open-eo.github.io/openeo-api/glossary/#aggregate-reducing-resolution - it says the assignment to groups is **unique**.

As a side note - I suggest we stop [treating pixels as little squares](https://open-eo.github.io/openeo-api/glossary/#aggregate-reducing-resolution) and just go with pixel centres whenever we need to think about the pixel in continuous space.

m-mohr: Right. So how would that work with lines/points? It seems we need to define separate behavior for polygons (pixel centers) and lines/points (pixel centers in a buffer around the feature? Are assignment in this case always unique?). So I'm questioning the original proposal to merge all feature types into a single process...

mkadunc: With lines and points, it is quite difficult to enforce unique assignment unless we somehow prevent "occupied" pixels from being used. If we use center-inside rule for polygons, a nearest-pixel rule could be seen as analogous, e.g. for point set `A`:
- find the pixel centre `Pxy` nearest to point `Ai` (this is equivalent to finding the square cell that `Ai` falls into - squares are the [voronoi cells](https://en.wikipedia.org/wiki/Voronoi_diagram) of the centre-point grid)
- check if `Ai` is the nearest point from set `A` to pixel centre `Pxy` (to ensure uniqueness)
- if `Ai` is not the nearest, then `Ai` gets no pixels

Another option would be to include a parameter `radius` (with default value `sqrt(2) * gsd` where `gsd` is raster ground sampling distance) and include, for each point, all pixels `Pxy` within `radius` from point `Ai`, for which `Ai` is the nearest point from set `A`. (this second check is not necessary if the set `A` does not contain pairs of points that are less than `2 * radius` apart).

For linestrings, we can do similar - either use all pixels whose cells intersect the line `Ti` and for which the line `Ti` is the nearest of all the lines in set `T` (i.e. select all pixels whose centres are nearest to at least one point on the linestring); or use a radius (default could again be `sqrt(2) * gsd`) with the same uniqueness check (which is not necessary if any two lines from `T` are at least `2 * radius` apart).

While these process would fit our description for the *aggregate* process, I'm not sure whether aggregation is really what the user wants when calling this with a non-polygon geometry. If I understand the raster as a discrete sampling of some underlying smooth function, then:
- for points, the more common operation one would expect would be that of `resample` (i.e. I would like to get a sampling of that underlying function at locations that are different from the ones provided in the raster); the value will usually be better if created from raster interpolation rather than just selection of a single raster value;
- for lines, I imagine the need of the user is to get summary statistics of the values that the *underlying smooth function* assumes along the provided linear feature, which might be again better served by using an interpolation of the raster (and then some aggregation for statistics) rather than all-or-nothing aggregation of some neighbouring pixels

I would support removing the point and linestring features from the `aggregate_spatial` process and creating a more targeted process for the point-sampling and summary-along-a-linestring use-cases.

m-mohr: I agree. I think polygon and line/point are too different and should be separate processes.

GreatEmerald:
It's true that the "aggregate" name is a bit misleading in the point and line case, but from the usability point of view I think it's still better to have one process that gives summary statistics over geometry, than to have separate processes. You don't want to tell users "use this process to extract information over a polygon, and this process to extract information over lines and points", for the users it's all geometry anyway. You can make it handle different geometry in different ways if that makes the most sense (so it would be a process that behind the scenes calls one of three processes based on geometry type), as long as that makes it do the right thing for most users.

Using a buffer around points and lines is also fine as far as implementation goes. (Though we need to think about what happens if you have a pixel on the corner between 4; I'd say it should return all 4 and reduce it with the given reducer.)

m-mohr:
Telco: We remove the "uniqueness" constraint for aggregate from the glossary.
We allow all geometries in aggregate_spatial. If I remember correctly, we'll still use the pixel centers, but for lines and points it will consider the closest pixel centers. This means pixels may be used for multiple geometries. Correct? @mkadunc @jdries @lforesta @aljacob
Any other things I should mention in the process description? Did I forget anything?

m-mohr:
Closing for now, I'll work on a joint PR for the issues #37, #62, also re-based on #94.

m-mohr:
There's a new proposal in PR #131

Issue #101, opened 2019-12-10
Implicit bounding box filtering by aggregate_polygon


jdries:
In the geopyspark backend, we currently use the bounding box of the geometry/feature collection specified in 'aggregate_polygon' as an alternative to 'filter_bbox' or spatial_extent in load_collection.

This is very convenient for the user, as he does not have to extract the bounds himself, which can be annoying, especially when working with large files, or when trying to write process graphs that take any file as input.

Is this something that we should clarify in the process definition, or can we leave it implicit? (Compatibility across backends would be affected if other backends choose not to support this.).

This unit test in our backend tries to show the type of process graphs that we want to support:
https://github.com/Open-EO/openeo-python-driver/blob/5c32f52363de96977745e6c9d43214e8086b4e70/tests/test_views_execute.py#L499

m-mohr:
I guess the best way to find out would be to just make a PR and propose this change on the call on Thursday.

lforesta:
> This is very convenient for the user, as he does not have to extract the bounds himself, which can be annoying, especially when working with large files, or when trying to write process graphs that take any file as input.

For this the user can use filter_polygon instead of filter_bbox, no?

Anyway, I also would implicitly filter data when using aggregate_polygon (we have not yet implemented this process)

jdries:
Telco conclusion: agreed to improve documentation.
Some notes:
- Explain the mechanism and conditions when a spatial filter can be 'pushed down'
- Explain that backends might still throw an exception if the push-down behavior is not supported
- Point user to 'filter_polygon' which can make his process graph more robust across different backends

m-mohr:
What is "push-down behavior"? Seems like an internal optimization?! I don't think this is something a user would really understand or care about. It sounds reasonable to just say that aggregate_* also implicitly applies filter_*.

m-mohr:
See PR #101 for a proposal.

lforesta:
PR #109 ?

m-mohr:
Indeed, 101 is this issue.

m-mohr:
Merged.

PR #109, opened 2019-12-17
aggregate_polygon improvements

m-mohr:
Two changes for `aggregate_polygon` (applies similarly to #107 if accepted):
- Doesn't allow returning a GeoJSON any longer. We can't really work with the returned GeoJSON anyway. To export GeoJSON use save_result on a vector data cube.
- This is the change to debate, based on issue #101: The data cube implicitly gets restricted to the bounds of the polygons as if `filter_polygon` would have been used afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Default values are now specified on the parameter-level, not in the JSON schemas.
 - Processes supporting multiple data types in parameters or return values with `anyOf` are now listing the data types directly as array. `anyOf` is discouraged.
 - `add_dimension`: Parameter `value` renamed to `label`.
-- `aggregate_polygon`: The data cube implicitly gets restricted to the bounds of the polygons as if `filter_polygon` would have been used afterwards. [#101](https://github.com/Open-EO/openeo-processes/issues/101)
+- `aggregate_polygon`: The data cube implicitly gets restricted to the bounds of the polygons as if `filter_polygon` would have been used beforehand. [#101](https://github.com/Open-EO/openeo-processes/issues/101)
 - `clip`: Works on a single value instead on arrays (replaced parameter `data` with `x`). [#75](https://github.com/Open-EO/openeo-processes/issues/75)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Default values are now specified on the parameter-level, not in the JSON schemas.
 - Processes supporting multiple data types in parameters or return values with `anyOf` are now listing the data types directly as array. `anyOf` is discouraged.
 - `add_dimension`: Parameter `value` renamed to `label`.
+- `aggregate_polygon`: The data cube implicitly gets restricted to the bounds of the polygons as if `filter_polygon` would have been used afterwards. [#101](https://github.com/Open-EO/openeo-processes/issues/101)
```
mkadunc:
```suggestion
- `aggregate_polygon`: The data cube implicitly gets restricted to the bounds of the polygons as if `filter_polygon` would have been used beforehand. [#101](https://github.com/Open-EO/openeo-processes/issues/101)
```

aggregate_polygon.json
```diff
@@ -88,17 +88,11 @@
...
-            {
-                "type": "object",
-                "subtype": "geojson"
-            },
-            {
-                "type": "object",
-                "subtype": "vector-cube"
-            }
-        ]
+        "description": "A vector data cube with the computed results. The vector data cube implicitly gets restricted to the bounds of the polygons as if ``filter_polygon()`` would have been used with the same values for the corresponding parameters after this process.\n\nThe computed value is stored in dimension with the name that was specified in the parameter `name`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) in each geometry. These values are stored as attributes of the result value with the attribute names `total_count` and `valid_count`.",
```
mkadunc: `after` -> `before`

```suggestion
        "description": "A vector data cube with the computed results. The vector data cube implicitly gets restricted to the bounds of the polygons as if ``filter_polygon()`` would have been used with the same values for the corresponding parameters immediately before this process.\n\nThe computed value is stored in dimension with the name that was specified in the parameter `name`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) in each geometry. These values are stored as attributes of the result value with the attribute names `total_count` and `valid_count`.",
```

Issue #103, opened 2019-12-13
Invert mask


soxofaan:
The mask process allows polygons to be passed as mask, in which case the pixels covered by the polygons are masked out (replaced with replacement value):

> All pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC) are replaced.

(from https://openeo.org/documentation/0.4/processes.html#mask)

Wouldn't it be useful as well to have an option to flip this behaviour: preserve the pixels that are covered by any of the polygons and replace the "outside" of the polygons. (This would actually be my intuitive expectation of how this process would handle polygons by default)

m-mohr:
Sounds useful. I guess the parameter should also apply to the other options of `mask`.

soxofaan:
indeed (and actually, my intuition would expect that flipped behavior to be the default for raster masks as well, but maybe it's an EO convention I'm not used to yet)

lforesta:
We could also use the process "not" to invert the mask, but it's a bit clearer if this behavior can be defined directly when calling "mask".

soxofaan:
> We could also use the process "not" to invert the mask, 

But that wouldn't work for a polygon (GeoJSON) based mask, right?

m-mohr:
It would, but with more steps: Create mask from GeoJSON, apply `not` on all pixels in mask, apply mask on original data.
That's not very intuitive and I agree that there should be a more user-friendly option for it.

m-mohr:
There's now PR #110 to implement this proposal. @soxofaan

m-mohr:
Has been merged. Is now available in mask_polygon.

Issue #55, opened 2019-04-16
mask: optional parameter 'mask-dimensions'


lforesta:
While writing the process graph for use case 1 with @claxn, we had to mask two cubes with dimension (x,y,t_1,s_1) and (x,y,t_2,s_2)

The mask needed is defined in dimensions (x,y), but following the 'mask' process definition, it has to have the same dimensions as the input raster cube to be masked. So in the process graph, we create the 'same' mask twice.

What about adding an optional argument to the process definition <mask-dimensions:(dim_1, dim_2, ...)>
If this option is present, the back-end knows in which dimensions the mask is defined and will apply it to the input cube for all elements of the other dimensions (t and s in the example above). And if it isn't specified, it would work as currently defined and the back-end would expect the mask to have identical dimensions and cardinality as the raster cube to be masked.

What do you think?

note: the name 'mask-dimensions' can be interpreted in two opposing ways, so if people like the idea, we may come up with a better name for this option.

m-mohr:
Add a parameter dimensions that name which dimensions the mask applies to.

jdries:
Isn't this something that can be inferred by the backend? At least applying a 2D mask to a 3D datacube seems straightforward. 
We could for instance say that the dimension names of the mask should match the ones in the datacube, and missing dimensions simply means that the mask is to be applied for each label of the missing dimension.
Seems to me that this would be easier to use.

lforesta:
this works when the mask is a datacube with its dimensions and I'm happy with this suggestion.

What about when a user want to use a user-uploaded raster as mask? We need to create a cube with dimensions out of it before applying the mask?

m-mohr:
> What about when a user want to use a user-uploaded raster as mask? We need to create a cube with dimensions out of it before applying the mask?

Yes, accepted types for the mask are: geojson, raster-cube (, vector-cube). There's no simple "raster" data type in openEO?!

mkadunc:
I agree with Jeroen - if mask axes are a subset of data axes, we can automatically apply it. If mask has extra dimensions, we should probably fail. Same if mask has missing labels on any of the common axes.

m-mohr:
Anyone up for a PR? I'm a bit lost what the conclusion is...

Is it the following? @mkadunc @jdries @lforesta 
1. Don't add a new parameter?
2. The data cube and the mask have to be compatible, but dimensions can be missing in the mask.
[...]

PR #110, opened 2019-12-17
Invert mask

m-mohr:
~New parameters `invert` for process `mask` as proposed in issue #103.~

Split vector- and raster-based masking, ~added expression parameter to mask~ and added mask_polygon.


mask.json
```diff
@@ -1,14 +1,15 @@
...
+    "description": "Applies a mask to a raster data cube. By default, a mask can either be specified as:\n\n* **Raster data cube** for which parallel pixels among `data` and `mask` are compared and those pixels in `data` are replaced, which are non-zero (for numbers) or `true` (for boolean values) in `mask`.\n* **GeoJSON or vector data cube** containing one or more polygons. All pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC) are replaced.\n\nThis behavior can be inverted with the parameter `invert`. The pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data). No data values will be left untouched by the masking operation.",
```
soxofaan: The "by default" looks a bit strange there imho.

"parallel pixels" also sounds bit strange to me, isn't "corresponding pixels" a more conventional way of explaining this?

> No data values will be left untouched by the masking operation.

Could be interpreted as "all values will be touched" :)  Also not clear if it is about `mask` or `data` values. I guess it could be rewritten as "Pixels corresponding with no-data `mask` pixels will be left untouched by the masking operation". It also only applies to the raster data cube case, so could be moved there as well?

soxofaan: > intersects with the corresponding polygon

intersects with any polygon

mask.json
```diff
@@ -48,6 +49,13 @@
                 ]
             },
             "default": null
+        },
+        "invert": {
+            "description": "If set to `true`, inverts the mask so that the behavior changes as follows:\n\n* **Raster data cube**: Parallel pixels among `data` and `mask` are compared and those pixels in `data` are replaced, which are **zero** (for numbers) or **`false`** (for boolean values) in `mask`.\n* **GeoJSON or vector data cube**: All pixels for which the point at the pixel center does **not** intersect with the corresponding polygon (as defined in the Simple Features standard by the OGC) are replaced.",
```
soxofaan: > does **not** intersect with the corresponding polygon

does **not** intersect with any polygon

m-mohr: Thanks for spotting this, was copy/pasted from aggregate_polygon obviously. Will be fixed. ;-)

mask.json
m-mohr: > The "by default" looks a bit strange there imho.

Removed.

> "parallel pixels" also sounds bit strange to me, isn't "corresponding pixels" a more conventional way of explaining this?

Not sure, I think I took this wording from another library originally.

> Also not clear if it is about `mask` or `data` values.

Hmm, I think the reason was, because it applies to both, but I'm not sure about this any longer. What makes sense from the users perspective?

> It also only applies to the raster data cube case, so could be moved there as well?

Only if it only applies to the mask, but I think it was meant to also apply on the data itself (see above).

soxofaan: > What makes sense from the users perspective?

I would expect that only `mask` defines the masking:

- mask == true : replace
- mask == false : preserve `data
- mask == null : preserve `data`

and in invert mode:

- mask == true: preserve
- mask == false: replace
- mask == null : preserve `data`

m-mohr: What happens with no data values in data? I think I would expect no data to stay as no data, but not sure this is what usually is expected. Or should we add "ignore_nodata" as in other processes?

soxofaan: I would expect that no-data is replaced with the replacement value if the mask dictates so. The null-ness of the `data` value is irrelevant I would think.

m-mohr: Okay, thanks. More opinions on this?

mask.json
```diff
@@ -1,14 +1,15 @@
...
+    "description": "Applies a mask to a raster data cube. A mask can either be specified as:\n\n* **Raster data cube** for which parallel pixels among `data` and `mask` are compared and those pixels in `data` are replaced, which are non-zero (for numbers) or `true` (for boolean values) in `mask`.\n* **GeoJSON or vector data cube** containing one or more polygons. All pixels for which the point at the pixel center intersects with any polygon (as defined in the Simple Features standard by the OGC) are replaced.\n\nThis behavior can be inverted with the parameter `invert`.\n\nThe pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data). No data values in `data` or `mask` will be left untouched by the masking operation.",
     "categories": [
         "masks"
     ],
     "parameter_order": [
         "data",
         "mask",
-        "replacement"
+        "replacement",
+        "invert"
```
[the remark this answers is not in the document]
soxofaan: Good point indeed, "invert" does not clearly communicate what the inverted and non-inverted behaviour are.
The nice thing about `active_mask_value` is that it even could be extended to integer valued mask (not only binary masks)

On the other hand, the alternative parameter names you provide only work well for raster based masks. For polygon based masks they make less sense, while "invert" is a lot more intuitive.

Not sure what the best approach is here.
We could even think about splitting the process in a raster based process and a polygon based process. It would make the process descriptions and documentation also simpler (no need for "if raster mask then ..., if polygon then ...." repeatedly).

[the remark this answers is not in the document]
m-mohr: I'll try to split the processes, I think. It would really make the process(es) simpler and we can better describe specifics.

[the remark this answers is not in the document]
m-mohr: Thinking about the raster masking, what would be a good default for `active_mask_value`? Or should it be required? Current behavior is "non-zero (for numbers) or `true` (for boolean values)" is getting replaced. That behavior doesn't work with a parameter with a specific value any longer. I'm leaning towards a default value `true` as it's easy to create a boolean mask with apply(data, gt(x, 0)) for example.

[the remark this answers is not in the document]
soxofaan: I think it's not so uncommon to automatically translate bool `true` to "non-zero" in int land. So if you have an int-valued mask and the default (or explicitly set) value of param `active_mask_value` is bool `true`, you mask out the pixels with non-zero mask values.

Another issue when extending to int-masks with an active_mask_value `M`: do you mask out pixels `where mask is M`  or `where is not M`?
Both use cases are useful, I guess. Take for example a coverage class based mask with classes like "vegetated", "water", "urban", "clouded". In some cases you probably want to keep vegetated pixels and replace all the rest, while in other cases you want to replace clouded pixels and keep all the rest. To do this with an int-mask, you probably still need an "invert"-like parameter to toggle between these two behaviors.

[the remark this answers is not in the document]
m-mohr: > I think it's not so uncommon to automatically translate bool true to "non-zero" in int land. 

Yes, although this is some kind of implicit magic and what happens if a user really just want to replace true... 

> Another issue when extending to int-masks with an active_mask_value M: do you mask out pixels where mask is M or where is not M?

@soxofaan Yes, that's exactly what I'm thinking about at the moment and what keeps me from pushing the changes for review. Another option would be to define a expression like callback where you could use things like eq(), neq(), between(), not(between()) etc. Then we are not restricted to a single value and don't need the invert param.

mask.json
```diff
@@ -48,6 +49,13 @@
                 ]
             },
             "default": null
+        },
+        "invert": {
+            "description": "If set to `true`, inverts the mask so that the behavior changes as follows:\n\n* **Raster data cube**: Parallel pixels among `data` and `mask` are compared and those pixels in `data` are replaced, which are **zero** (for numbers) or **`false`** (for boolean values) in `mask`.\n* **GeoJSON or vector data cube**: All pixels for which the point at the pixel center does **not** intersect with any polygon (as defined in the Simple Features standard by the OGC) are replaced.",
```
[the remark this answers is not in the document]
m-mohr: Probably obsolete with the recent changes.

mask.json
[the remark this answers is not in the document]
m-mohr: @soxofaan @mkadunc Please re-review the recent changes. There's now mask with an expression and mask_polygon.

mask.json
m-mohr: Now that there's a proposal for the parameter "expression" in mask, should we leave it up to the user to decide whether to care for null in the mask or not?

mask_polygon.json
```diff
@@ -0,0 +1,71 @@
+{
+    "id": "mask",
```
soxofaan: must be mask_polygon, I guess

m-mohr: Sure, will update.

mask.json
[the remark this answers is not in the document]
soxofaan: Quick note: from the standpoint of the python client, the approach with an expression feels like overkill because we already have  very easy, pythonic "band math" to build boolean masks  (e.g. `(coverage != 4) & (coverage != 5)` . I'm not sure if it's good for API orthogonality and simplicity.

Maybe, in the `mask` process,  we should just focus on boolean masks and provide basic support for integer masks? More advanced integer mask use cases are still possible anyway by first computing  a boolean mask and  using that in the mask process.

mask_polygon.json
```diff
@@ -0,0 +1,71 @@
+{
+    "id": "mask_polygon",
+    "summary": "Apply a polygon mask",
+    "description": "Applies a polygon mask to a raster data cube. All pixels for which the point at the pixel center **does not** intersect with any polygon (as defined in the Simple Features standard by the OGC) are replaced. This behaviour can be inverted by setting the parameter `inside` to `true`.\n\nThe pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data). No data values in `data` will be left untouched by the masking operation.",
```
soxofaan: > No data values in `data` will be left untouched by the masking operation.

In my opinion,  no-data values in `data` should still be masked/maskable. It won't make a big difference in practice as the replacement value is null typically. Also, to me, the sentence itself reads as "No values will be untouched ...", which is bit strange. I would just remove the sentence.

mask.json
```diff
@@ -1,14 +1,15 @@
...
+    "summary": "Apply a raster mask",
+    "description": "Applies a mask to a raster data cube. A mask here is a raster data cube for which parallel pixels among `data` and `mask` are compared and those pixels in `data` are replaced for which a condition in `mask` is fulfilled (see parameter `expression`).\n\nThe pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data). No data values in `data` or `mask` will be left untouched by the masking operation.",
```
soxofaan: I'm more a fan of using "corresponding pixels" instead of "parallel pixels"

Apart from that: there should probably be a note about how this parallellism/correspondance is handled in terms of cube dimensions and shape. Simple approach: dimensions (and shape) should be the same. But in practice one probably wants also be able to mask (x, y, t, band) data cubes with (x,y) masks?

mask.json
soxofaan: > should we leave it up to the user to decide whether to care for null in the mask or not?

the output of the expression could still contain null I guess, so the backend still has to handle that case in a consistent way, no?

mask.json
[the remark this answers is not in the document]
m-mohr: @soxofaan Other clients can't do that though. How do you translatethe Python example into a process graph?

I don't understand how your comments https://github.com/Open-EO/openeo-processes/pull/110#discussion_r365274430 and https://github.com/Open-EO/openeo-processes/pull/110#discussion_r365324900 - they seem to contradict each other?

So at the moment I'm either in favor of a boolean only approach and thus people need to prepare a mask with true values (so they need to create the expression in apply anyway, so what's the benefit?) or having the expressions in mask directly. It seems the other options have too many drawbacks.

mask.json
m-mohr: Indeed, I forgot about that. I'll add documentation for that case.

mask.json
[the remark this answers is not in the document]
soxofaan: > How do you translate the Python example into a process graph?

see https://gist.github.com/soxofaan/1121ca95b45fc43f697fc82b7e5ed7b7 for example

> your comments seem to contradict each other

just some samples from a thought process :)  . 
I guess I'm converging to the conclusion to keep things simple and just support a boolean mask (the API has all the functionality already to build these, no need to replicate that inside a process parameter) .
Also note that the original request from #103 was about an invert option for polygon based masking, not raster masks.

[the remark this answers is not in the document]
m-mohr: For me it seems that having the expression is simpler for a user though. In the end it's just an implicit apply call that a user would need to specify explicitly for a mask.

mask.json
m-mohr: Indeed, suggestions how to phrase the latter?

mask.json
m-mohr: Although, it is implicitly clear:

> A pixel is replaced in `data` if the expression returns `true`

That means for false and null nothing changes...

mask_polygon.json
m-mohr: Removed the sentence.

mask.json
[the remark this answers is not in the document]
GreatEmerald: In R for raster masking, we have two values: what value in the mask we care about, and what value we replace those pixels with in the raster we're masking. So we just have two integer (arrays). The defaults can be 1 and NA respectively, because normally users want to put NAs into areas that they want to mask.
For vector masking, same idea but you can have a parameter that says `replace_inside=TRUE`, that way values inside of the geometry will get replaced, otherwise values outside the geometry will be replaced.

mask.json
m-mohr: Telco: Will be tackled in a separate PR.

m-mohr: Oh, seems there's already a related issue: #55

mask.json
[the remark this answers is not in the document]
m-mohr: Telco: Remove expressions for now.

mask_polygon.json
m-mohr: There's now a test catching this...

Issue #87, opened 2019-11-14
merge_cubes: Clarify merging procedure


jdries:
In our use case, we want to run two separate operations on a given datacube, and then combine those two results.
To do this, we either need a reducer that can reduce values coming from different cubes, or we need a process that 'joins' the cubes first to a datacube with two bands. This is probably the most generic solution and needed anyway.

m-mohr:
There's merge_cubes, which has been defined at the VITO sprint. If that's not working properly, we should probably improve that process instead of adding a new one.

jdries:
Indeed, I found that one as well. Perhaps we should then clarify how it handles bands?
Or is there a reducer that basically leaves the bands untouched? Like
merge([first_cube_band],[second_cube_band]) = [first_cube_band,second_cube_band]

m-mohr:
My understanding of the process is that if you pass a two datacube with dimensions x,y,t,b that one has bands a,b and the other has bands c,d the result will be a datacube x,y,t,b, with bands a,b,c,d. Same applies for t etc.

@mkadunc You were pushing that definition at the VITO sprint. Do you agree?

jdries:
That would make sense, adding that to the documentation would solve the issue.

mkadunc:
Yes, that was my understanding.

A caveat for Jeroen's scenario - if the two input cubes do not yet have band dimension, merge should fail (`merge(A(x,y,t), B(x,y,t))` won't work whenever any cell is present in both cubes) - to be able to use the `merge` process on such cubes, one should extend the cubes first:
```
    merge(
        add_dimension(A(x,y,t), "bands", "a"),
        add_dimension(B(x,y,t), "bands", "c")
    ) // returns cube with dimensions (x, y, t, bands)
```

m-mohr:
@mkadunc Shouldn't `merge(A(x,y,t), B(x,y,t))` still work? I would expect that the temporal dimension could similarly be merged. For example if A has values for 2017 and B for 2018, why should that fail? Seems to be a reasonable use case to me?

mkadunc:
> Shouldn't merge(A(x,y,t), B(x,y,t)) still work?
Yes, of course - in your example there are no overlapping cells, because the two cubes are separated in time.

I was commenting on Jeroen's scenario, in which the two cubes occupy the same space and time, and the introduction of an extra dimension makes them disjoint again.

m-mohr:
Thanks, got it.

mkadunc:
[...]

PR #112, opened 2019-12-18
Clarified merge behavior of merge_cubes #87

m-mohr:
I tried to clarify how merge_cubes works.

For easier reading, here's the new process description:

> The data cubes have to be compatible. A merge is the inverse of a split if there is no overlap. The process doesn't add dimensions.
>
> This means that the data cubes must have the same dimensions. Each dimension must be available in both data cubes have the same name, type, reference system and resolution. One of the dimensions can have different labels, for all other dimensions the labels must be equal. If data overlaps, the parameter `overlap_resolver` must be specified to resolve the overlap.
>
> **Examples for merging two data cubes:**
> 
> 1. Data cubes with the dimensions `x`, `y`, `t` and `bands` have the same dimension labels in `x`,`y` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the first cube and `B3` and `B4`. An overlap resolver is *not needed*. The merged data cube has the dimensions `x`, `y`, `t` and `bands` and the dimension `bands` has four dimension labels: `B1`, `B2`, `B3`, `B4`.
> 2. Data cubes with the dimensions `x`, `y`, `t` and `bands` have the same dimension labels in `x`,`y` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the first cube and `B2` and `B3`. An overlap resolver is *required* to resolve overlap in band `B2`. The merged data cube has the dimensions `x`, `y`, `t` and `bands` and the dimension `bands` has three dimension labels: `B1`, `B2`, `B3`.
> 3. Data cubes with the dimensions `x`, `y` and `t` have the same dimension labels in `x`,`y` and `t`. There are two options:
>    1. Keep the overlapping values separately in the merged data cube: An overlap resolver is *not needed*, but for each data cube you need to add a new dimension using ``add_dimension()``. The new dimensions must be equal, except that the labels for the new dimensions must differ by name. The merged data cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.
>    2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.
> 4. ~Data cubes with the dimensions `x`, `y` and `t` have the same dimension labels in `x`,`y` and `t`. An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.~

@jdries What do you think?


merge_cubes.json
```diff
@@ -1,7 +1,7 @@
...
...s with the dimensions `x`, `y` and `t` have the same dimension labels in `x`,`y` and `t`. There are two options:\n   1. Keep the overlapping values separately in the merged data cube: An overlap resolver is *not needed*, but for each data cube you need to add a new dimension using ``add_dimension()``. The new dimensions must be equal, except that the labels for the new dimensions must differ by name. The merged data cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.\n   2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.\n4. Data cubes with the dimensions `x`, `y` and `t` have the same dimension labels in `x`,`y` and `t`. An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.",
```
soxofaan: > Each dimension must be available in both data cubes have the same name, type,

I guess there is an "and" missing between "cubes" and "have"

soxofaan:
what is the difference between 3ii and 4?

m-mohr:
> what is the difference between 3ii and 4?

Hah, good point. I added the split into 3i/ii later and forgot to remove 4. Fixed.

merge_cubes.json
m-mohr: Fixed.

soxofaan:
while at it, maybe also improve this part
```
        "OverlapResolverMissing": {
            "message": "Two data cubes with overlap but without an overlap resolver have been specified."
        }
```

e.g "Overlapping data cubes, but no overlap resolver has been specified"

merge_cubes.json
```diff
@@ -1,7 +1,7 @@
...
...` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the first cube and `B2` and `B3`. An overlap resolver is *required* to resolve overlap in band `B2`. The merged data cube has the dimensions `x`, `y`, `t` and `bands` and the dimension `bands` has three dimension labels: `B1`, `B2`, `B3`.\n3. Data cubes with the dimensions `x`, `y` and `t` have the same dimension labels in `x`,`y` and `t`. There are two options:\n   1. Keep the overlapping values separately in the merged data cube: An overlap resolver is *not needed*, but for each data cube you need to add a new dimension using ``add_dimension()``. The new dimensions must be equal, except that the labels for the new dimensions must differ by name. The merged data cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.\n   2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.",
```
mkadunc: > A merge is the inverse of a split if there is no overlap.

This sounds out-of-place and kind of cryptic... How about this:  
*The data cubes have to be compatible. A merge operation without overlap should be reversible with (a chain of) `filter` operations for each of the two cubes. The process doesn't add dimensions.*

mkadunc: In example 2.
> `bands` are `B1` and `B2` for the first cube and `B2` and `B3`.

Should be *... and `B2` and `B3` for the second.`

m-mohr: Sounds good, changed.

m-mohr: Fixed.

m-mohr:
> e.g "Overlapping data cubes, but no overlap resolver has been specified"

Sounds good, fixed.

m-mohr:
Seems there are no more objections => merging.

PR #115, opened 2020-01-13
Improvements for comparison processes

m-mohr:
Changes:
- Comparison processes `eq`, `gt`, `gte`, `lt`, `lte`, `neq` and `between` accept all data types as input for the operands.
- `text_begins`, `text_contains`, `text_ends`: `null` values are supported and get passed through.
- Fix: `between` may return a `null` value.
- Several clarifications

I think all changes are quite straightforward. The thing to debate is probably whether arrays and objects should be comparable or always return false?

Main reason for the changes was the absence of some data types from some of the comparison, especially booleans.


m-mohr:
Telco: Only work on scalars.

Issue #69, opened 2019-08-20
apply_kernel: clarify documentation


jdries:
Some feedback after implementing apply_kernel:
* There actually seem to be multiple ways in which a kernel can be applied, but I believe we are referring to a 2D convolution. This can be mentioned more explicitly. Probably may make sense to reference more elaborate explanations like: http://www.songho.ca/dsp/convolution/convolution2d_example.html
* The spec says that the kernel should have the same dimensions as the datacube. The most logical case however is to have a spatial 2D kernel, and to apply that kernel on the 2 spatial dimensions. We can then also allow backends to support xD kernels, if they want to do that. This reduces complexity for the user and backend implementors, as they don't have to worry about higher dimensional kernels.
* Applying a kernel is only properly defined if the kernel has uneven dimension. Currently my implementation throws an exception when trying for instance to use a 2x2 kernel. The docs can mention this, or we should come up with a convention on how to handle kernels with even dimensions.

m-mohr:
Solved in PR #118

PR #118, opened 2020-01-14
apply_kernel: 2D convolutions

m-mohr:
Implementing suggestions made in issue #69

@jdries


apply_kernel.json
```diff
@@ -21,18 +21,22 @@
...
+                    "items": {
+                        "type": "number"
+                    }
                 }
             },
             "required": true
         },
         "factor": {
-            "description": "A factor that is multiplied to each value computed by the focal operation.\n\nThis is basically a shortcut for explicitly multiplying each value by a factor afterwards, which is often required for some kernel-based algorithms such as the Gaussian blur.",
+            "description": "A factor that is multiplied to each value after the kernel has been applied.\n\nThis is basically a shortcut for explicitly multiplying each value by a factor afterwards, which is often required for some kernel-based algorithms such as the Gaussian blur.",
```
soxofaan: >  after the kernel has been applied

does this have to be specified explicitly here? Isn't that just an implementation detail? e.g. a backend may want to pre-multiply the kernel before applying the convolution instead of post-multiplying the output.

m-mohr: This is a description for the user. I have a hard time describing this in a implementation agnostic way. Does it really matter how implementations do it internally? The result should be the same anyway... Writing something like "A factor that is multiplied either to each computed value afterwards or to each kernel element beforehand." is confusing for the user, I think. Suggestions?

soxofaan: I would just avoid explicit "afterwards/beforehand". Suggestion:

> An additional multiplication factor to multiply the result with

m-mohr: Okay, I'll remove that.

m-mohr: But it's still mentioned in the process description and I really find it difficult to explain how the process works without saying whether the factor is applied to each value of the kernel or the result of the focal operation.

m-mohr:
Merging for now as I have to make some bigger structural changes to processes now. Please open a new PR for the open issue: https://github.com/Open-EO/openeo-processes/pull/118#discussion_r368426866

PR #119, opened 2020-01-14
mask: Clarify behavior for masks with missing dimensions

m-mohr:
Clarify behavior for masks with missing dimensions. Issue #55

See also https://github.com/Open-EO/openeo-processes/issues/55#issuecomment-567393535 and https://github.com/Open-EO/openeo-processes/pull/110#discussion_r365336523


mask.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Applies a mask to a raster data cube. To apply a vector mask use ``mask_polygon()``.\n\nA mask is a raster data cube for which corresponding pixels among `data` and `mask` are compared and those pixels in `data` are replaced which are non-zero (for numbers) or `true` (for boolean values). The pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data).\n\nThe data cubes have to be compatible so that each dimension in mask must also be available in the raster data cube with the same name, type, reference system, resolution and labels. Dimensions can be missing in the mask with the result that the mask has to be applied for each label of the missing dimension in the data cube. The process fails if there's a incompatibility found between the raster data cube and the mask.",
```
soxofaan: > The process fails if ...

do we have a particular error code for dimension mismatches?

m-mohr: I thought there's potentially too many different things to consider and inform a user about so that a single error code / definition would be to general and not very helpful.

soxofaan: ok fair eneough

mkadunc: - make it more clear that non-zero or `true` refers to values in `mask`.
- "mask **is** applied" instead of "mask **has to be** applied" - to make it clear that the system does applies it automatically; it's not the user's job to apply the mask for each label.
- "a incompatibility" -> "an incompatibility"

```suggestion
    "description": "Applies a mask to a raster data cube. To apply a vector mask use ``mask_polygon()``.\n\nA mask is a raster data cube for which corresponding pixels among `data` and `mask` are compared and those pixels in `data` are replaced whose pixels in `mask` are non-zero (for numbers) or `true` (for boolean values). The pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data).\n\nThe data cubes have to be compatible so that each dimension in mask must also be available in the raster data cube with the same name, type, reference system, resolution and labels. Dimensions can be missing in the mask with the result that the mask is applied for each label of the missing dimension in the data cube. The process fails if there's an incompatibility found between the raster data cube and the mask.",
```

m-mohr: Thanks

Issue #114, opened 2019-12-20
Better array handling


m-mohr:
We have some cases where we may want to iterate over an array so I'm thinking about either defining a function array_each or allow arrays as input data for apply.

array_each (the definition is a bit outdated):
```json
{
    "id": "array_each",
    "summary": "Applies a unary process to each array element",
    "description": "Applies a **unary** process which takes a single value such as `abs` or `sqrt` to each value in the array.",
    "categories": [
        "arrays"
    ],
    "parameter_order": ["data", "process"],
    "parameters": {
        "data": {
            "description": "An array.",
            "schema": {
                "type": "array",
                "items": {
                    "description": "Any data type is allowed."
                }
            },
            "required": true
        },
        "process": {
            "description": "A process (callback) to be applied on each value. The specified process must be unary meaning that it must work on a single value.",
            "schema": {
                "type": "object",
                "format": "callback",
                "parameters": {
                    "x": {
                        "description": "A value of any data type could be passed."
                    }
                }
            },
            "required": true
        }
    },
    "returns": {
        "description": "An array with the newly computed values. The number of elements are the same as for the original array.",
        "schema": {
			"type": "array",
			"items": {
				"description": "Any data type is allowed."
			}
        }
    }
}
```

m-mohr:
It's probably also useful to define array_filter and array_reduce?!

m-mohr:
Solved with PR #121.

PR #121, opened 2020-01-15
Add array_apply and array_filter, generally improve array handling

m-mohr:
Implements #114


array_contains.json
```diff
@@ -1,14 +1,14 @@
 {
     "id": "array_contains",
-    "summary": "List contains an element",
-    "description": "Checks whether the array specified for `data` contains the value specified in `value`. Returns `true` if there's a match, otherwise `false`.\n\n**Remarks:**\n\n* To get the index of the element found, use ``array_find()``.\n* All definitions for the process ``eq()`` regarding the comparison of values apply here as well. This includes that data types MUST be checked strictly, for example a string with the content *1* is not equal to the number *1*.\n* An integer *1* is equal to a floating point number *1.0* as `integer` is a sub-type of `number`. Still, this process may return unexpectedly `false` when comparing floating point numbers due to floating point inaccuracy in machine-based computation.\n* Temporal strings are treated as normal strings and MUST NOT be interpreted.",
+    "summary": "Check whether the List contains a value",
```
soxofaan: maybe better:
> Check whether an array contains a given value

array_filter.json
```diff
@@ -0,0 +1,50 @@
...
+            "schema": {
+                "type": "array",
+                "items": {
+                    "description": "Any data type is allowed."
+                }
+            },
+            "required": true
+        },
+        "expression": {
+            "description": "An expression that is evaluated against each value in the array. An array element is dropped if the expression returns `false`.",
```
soxofaan: what about null values returned by expression?
maybe explain it positively: 

> Only the array elements where the expression returns `true` are preserved

array_filter.json
```diff
@@ -0,0 +1,50 @@
...
+    "id": "array_filter",
+    "summary": "Filter an array based on a logical expression",
+    "description": "Filters the array elements based on a logical expression so that afterwards an array is returned that only contains the values conforming to the expression.",
+    "categories": [
+        "arrays",
+        "filter"
+    ],
+    "parameter_order": [
+        "data",
+        "expression"
```
soxofaan: in #128 , the usage of "expression" was replaced by "condition" (for clarity reasons I guess). Maybe also do that here?

array_filter.json
m-mohr: Thanks, fixed.

array_contains.json
m-mohr: Thanks, fixed.

array_filter.json
m-mohr: Good catch, fixed.

PR #123, opened 2020-01-16
Split binary and array based reduction

m-mohr:
As discussed in issue #94...


aggregate_polygon.json
```diff
@@ -1,18 +1,16 @@
...
+    "description": "Aggregates zonal statistics for one or multiple polygons over the spatial dimensions. This process passes a list of values to the reducer. In contrast, ``aggregate_polygon_binary()`` passes two values, which may be better suited for UDFs.\n\nThe process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
mkadunc: `two raster dimensions` -> `two spatial dimensions`
```suggestion
    "description": "Aggregates zonal statistics for one or multiple polygons over the spatial dimensions. This process passes a list of values to the reducer. In contrast, ``aggregate_polygon_binary()`` passes two values, which may be better suited for UDFs.\n\nThe process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n\nThe data cube must have been reduced to only contain two spatial dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```

mkadunc: BTW: Why are we limiting this to one remaining dimension - e.g. why not get statistics over space for each (time,band) combination?

aggregate_polygon_binary.json
```diff
@@ -0,0 +1,92 @@
+{
+    "id": "aggregate_polygon",
```
mkadunc:
```suggestion
    "id": "aggregate_polygon_binary",
```

aggregate_polygon_binary.json
```diff
@@ -0,0 +1,92 @@
...
+    "summary": "Compute zonal statistics for polygons",
```
mkadunc:
```suggestion
    "summary": "Compute zonal statistics for polygons using a binary reducer",
```

aggregate_polygon.json
m-mohr: @mkadunc I can't remember why we chose this at the VITO sprint. Maybe for simplicity? I guess it's a good idea to open a new issue for this?!

m-mohr:
Thanks, @mkadunc. I integrated your suggestions.

reduce_dimension_binary.json
```diff
@@ -0,0 +1,67 @@
+{
+    "id": "reduce_dimension_binary",
+    "summary": "Reduce dimensions",
+    "description": "Applies a binary reducer to a data cube dimension by collapsing all the pixel values along the specified dimension into an output value computed by the reducer. This process passes a list of values to the reducer. In contrast, ``reduce_dimension()`` passes a list of values.\n\nThe dimension is dropped. To avoid this, use ``apply_dimension()`` instead.",
```
mkadunc:
```suggestion
    "description": "Applies a binary reducer to a data cube dimension by collapsing all the pixel values along the specified dimension into an output value computed by the reducer. This process repeatedly passes a pair of values to the reducer. In contrast, ``reduce_dimension()`` passes a list of values.\n\nThe dimension is dropped. To avoid this, use ``apply_dimension()`` instead.",
```

aggregate_polygon.json
m-mohr: Opened issue #126

reduce_dimension_binary.json
```diff
@@ -0,0 +1,67 @@
...
+        "data": {
+            "description": "A data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            },
+            "required": true
+        },
+        "reducer": {
+            "description": "A reduction operator to be applied consecutively on tuples of values. It must be both associative and commutative as the execution may be executed in parallel and therefore the order of execution is arbitrary. The reduction operator may be a single process such as ``multiply()`` or consist of multiple sub-processes.",
```
mkadunc:
```suggestion
            "description": "A reduction operator to be applied consecutively on pairs of values. It must be both associative and commutative as the execution may be executed in parallel and therefore the order of execution is arbitrary. The reduction operator may be a single process such as ``multiply()`` or consist of multiple sub-processes.",
```

Issue #84, opened 2019-10-09
Fix/improve data filtering


sinergise-anze:
We are trying to implement support for for Sentinel-1 GRD data in [load_collection process](https://open-eo.github.io/openeo-api/processreference/#load_collection). The data in this collection has a property "orbitDirection" ([sar:pass_direction](https://github.com/radiantearth/stac-spec/tree/master/extensions/sar)), with possible values `"ascending"`, `"descending"` or `null` (any). Our backend supports filtering by this property.

1) In [load_collection process](https://open-eo.github.io/openeo-api/processreference/#load_collection), I assume `properties` is meant for filtering the data that is to be loaded. However, it is unclear to me how the process graph would look like. Are there any examples available, or could someone please create a short example for this?

2) The issue with `properties` (as defined, that is: with callbacks) is that it would imho be quite difficult to convert the properties filters to query parameters for our backend, which means that `load_collection` would need to fetch all metadata, perform filtering on its end (for example just items with `orbitDirection` == `ascending`), and then request only data for remaining items. Not impossible, just more difficult to implement and a bit less optimal performance-wise. Ideally we would want to pass the appropriate parameters when fetching metadata, so that backend would already take care of filtering.

    That said, since `properties` are not finalized yet, one alternative suggestion would be to use `filters` instead of `properties`, and use notation of [STAC Query API](https://github.com/radiantearth/stac-spec/blob/master/api-spec/extensions/query/README.md). For example: 
    ```
    {
      "process_graph": {
        "loadco1": {
          "process_id": "load_collection",
          "arguments": {
            "id": "S1GRD",
            "spatial_extent": ...,
            "temporal_extent": ["2019-08-16", "2019-08-18"],
            "filters": {
              "sar:pass_direction": {
                "eq": "ascending"
              }
            }
          }
        },
        ...
      }
    }
    ```
    This would of course open an issue of how to describe which kind of queries the backend supports (`startsWith` might be unsupported) - or, alternatively, `load_collection` could simulate filtering if backend lacks such support.

I would appreciate some thoughts on this.

m-mohr:
> 1. In [load_collection process](https://open-eo.github.io/openeo-api/processreference/#load_collection), I assume `properties` is meant for filtering the data that is to be loaded.

Correct.

> However, it is unclear to me how the process graph would look like. Are there any examples available, or could someone please create a short example for this?

I don't have a process graph at hand, but it should be something like:

[...]

PR #128, opened 2020-01-17
Improvements for data filtering and other minor changes

m-mohr:
See issue #84.


if.json
```diff
@@ -24,55 +24,55 @@
...
-            "description": "A value that is returned if the boolean expression is `true`. Defaults to `true`.",
+            "description": "A value that is returned if the boolean value is `true`. Defaults to `true`.",
             "schema": {
                 "description": "Any data type is allowed."
             },
             "default": true
         },
         "reject": {
-            "description": "A value that is returned if the boolean expression is `false`. Defaults to `false`.",
+            "description": "A value that is returned if the boolean value is `false`. Defaults to `false`.",
```
soxofaan: I'm wondering if there would be more useful options for the default values of accept and reject. Now with true/false as defaults, it would make the `if` process redundant in a way (`if(x)` would be just `x` basically).

Another option would be to use `null` as default for reject, but I'm not sure what a good default for accept would be then (int `1` maybe?)

m-mohr: @soxofaan Indeed, I think my original intention was to allow expressions/conditions (e.g. eq(foo, bar)) in if, but the way it is now it's really just a not so useful boolean to boolean mapper by default. Do we even need a default for accept? Maybe remove the default for accept and then set the default for reject to null?

soxofaan: > Maybe remove the default for accept and then set the default for reject to null?

that sounds very sensible indeed

m-mohr: Done.

PR #131, opened 2020-01-20
Make aggregate_polygon and filter_polygon more general

m-mohr:
Make aggregate_polygon and filter_polygon more general (broaden it for points and line strings). (Is there a use case for mask_polygon, too?) 

Tackles issues #62 and #37. Based on PR #107 from @kempenep.


filter_spatial.json
```diff
@@ -1,7 +1,7 @@
 {
-    "id": "filter_polygon",
-    "summary": "Spatial filter using polygons",
-    "description": "Limits the data cube over the spatial dimensions to the specified polygons.\n\nThe filter retains a pixel in the data cube if the point at the pixel center intersects with at least one of the polygons (as defined in the Simple Features standard by the OGC).",
+    "id": "filter_spatial",
+    "summary": "Spatial filter using geometries",
+    "description": "Limits the data cube over the spatial dimensions to the specified geometries.\n\n- For **polygons**, the filter retains a pixel in the data cube if the point at the pixel center intersects with at least one of the polygons (as defined in the Simple Features standard by the OGC).\n- For **points** and **lines** (line strings), the filter retains all the pixel centers closest to any of the geometries.",
```
soxofaan: What happens to pixels outside of the geometries but still inside the corresponding bbox? Are they set to null?
If so: what is the difference with mask_polygon?
If not: what is the difference with filter_bbox?

m-mohr: They get removed whenever possible and all the other get set to null. It's basically like applying filter_bbox with the bbox of the geometry and all the other values not covered by the geometry in the bbox get set to null. Mask polygon doesn't restrict the area and you can choose replacement value and invert the area.

soxofaan: maybe too subtle to justify defining separate processes? For example, we could add an option like "optimize_bbox" to the mask process to implement what filter_spatial is doing.
To me, "mask_polygon" sounds more self-explanatory than "filter_spatial" anyway.

m-mohr: I think both have their value. There's another difference: mask_polygon doesn't support points and lines, which I think is somewhat reasonable.

aggregate_spatial.json
```diff
@@ -63,15 +57,15 @@
...
+        "description": "A vector data cube with the computed results. The vector data cube implicitly gets restricted to the bounds of the geometries as if ``filter_spatial()`` would have been used with the same values for the corresponding parameters immediately before this process.\n\nThe computed value is stored in dimension with the name that was specified in the parameter `name`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) in each geometry. These values are stored as attributes of the result value with the attribute names `total_count` and `valid_count`.",
```
mkadunc: If the `total_count` and `valid_count` are stored for each geometry, shouldn't they be values inside the vector cube? Or are these attributes meant to be arrays with the same cardinality as the spatial dimension of the vector cube?

m-mohr: @mkadunc Oh, yes, that's a leftover from the time where we returned GeoJSON only. See the most recent commit for the change.

m-mohr:
Hey @GreatEmerald! Could you please check this PR asap so that it is sure it suits your use-case? I'll need a response this week.

GreatEmerald:
OK, I'll try to look into this tomorrow.

aggregate_spatial.json
```diff
@@ -0,0 +1,80 @@
+{
+    "id": "aggregate_spatial",
+    "summary": "Zonal statistics for geometries",
+    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions. This process passes a list of values to the reducer. In contrast, aggregate_spatial_binary() passes two values, which may be better suited for UDFs.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points** and **lines** (line strings), the process considers all the closest pixel centers.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.\n\nThe data cube must have been reduced to only contain two spatial dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
GreatEmerald: I'm still unsure about the specification for "closest pixel centres" for a line. Closest to what? A line is made of infinite points, and pixel centres will be closer or further away from the line depending on the angle. For me, only a buffer of half the pixel size away from the line makes sense. Of course, if the backend developers know what "closest pixels to a line" means, then that's also fine for me as it is.

aggregate_spatial.json
```diff
@@ -0,0 +1,80 @@
...
+                            "items": {
+                                "description": "Any data type."
+                            }
+                        }
+                    }
+                ]
+            }
+        },
+        {
+            "name": "target_dimension",
```
GreatEmerald: For a vector, I'd call this `target_attribute` (since this is a column in a table), but it doesn't matter much.

aggregate_spatial.json
```diff
@@ -0,0 +1,80 @@
...
+            "description": "The new dimension name to be used for storing the results. Defaults to `result`.",
+            "schema": {
+                "type": "string"
+            },
+            "default": "result",
+            "optional": true
+        }
+    ],
+    "returns": {
+        "description": "A vector data cube with the computed results. The vector data cube implicitly gets restricted to the bounds of the geometries as if ``filter_spatial()`` would have been used with the same values for the corresponding parameters immediately before this process.\n\nThe computed value is stored in dimension with the name that was specified in the parameter `target_dimension`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) for each geometry. These values are stored as new dimension with a dimension name derived from `target_dimension` by adding the suffix `_meta`. The new dimension has the dimension labels `total_count` and `valid_count`.",
```
GreatEmerald: Why do we need to mention anything about how "The vector data cube implicitly gets restricted to the bounds of the geometries"? The vector data cube doesn't change at all, aside from gaining an extra attribute/dimension that holds the aggregation result.

aggregate_spatial.json
```diff
@@ -0,0 +1,80 @@
...
+            "name": "data",
+            "description": "A raster data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "geometries",
+            "description": "Geometries as GeoJSON on which the aggregation will be based.",
```
GreatEmerald: Why a GeoJSON and not a vector datacube? Or both?

aggregate_spatial.json
```diff
@@ -0,0 +1,80 @@
...
+    "returns": {
+        "description": "A vector data cube with the computed results. The vector data cube implicitly gets restricted to the bounds of the geometries as if ``filter_spatial()`` would have been used with the same values for the corresponding parameters immediately before this process.\n\nThe computed value is stored in dimension with the name that was specified in the parameter `target_dimension`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) for each geometry. These values are stored as new dimension with a dimension name derived from `target_dimension` by adding the suffix `_meta`. The new dimension has the dimension labels `total_count` and `valid_count`.",
+        "schema": {
+            "type": "object",
+            "subtype": "vector-cube"
+        }
+    },
+    "exceptions": {
+        "TooManyDimensions": {
+            "message": "The number of dimensions must be reduced to three for 'aggregate_spatial'."
```
GreatEmerald: Would be better to be more specific that this refers to the raster datacube, not the vector one.

aggregate_spatial_binary.json
```diff
@@ -65,15 +59,15 @@
...
+        "description": "A vector data cube with the computed results. The vector data cube implicitly gets restricted to the bounds of the geometries as if ``filter_spatial()`` would have been used with the same values for the corresponding parameters immediately before this process.\n\nThe computed value is stored in dimension with the name that was specified in the parameter `target_dimension`.\n\nThe computation also stores information about the total count of pixels (valid + invalid pixels) and the number of valid pixels (see ``is_valid()``) for each geometry. These values are stored as new dimension with a dimension name derived from `target_dimension` by adding the suffix `_meta`. The new dimension has the dimension labels `total_count` and `valid_count`.",
         "schema": {
             "type": "object",
             "subtype": "vector-cube"
         }
     },
     "exceptions": {
         "TooManyDimensions": {
-            "message": "The number of dimensions must be reduced to three for 'aggregate_polygon'."
+            "message": "The number of dimensions must be reduced to three for 'aggregate_spatial'."
```
GreatEmerald: This should be `aggregate_spatial_binary`?

aggregate_spatial_binary.json
```diff
@@ -1,7 +1,7 @@
...
+    "id": "aggregate_spatial_binary",
+    "summary": "Zonal statistics for geometries by binary aggregation",
+    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions. This process passes two values to the reducer. In contrast, ``aggregate_spatial()`` passes a list of values.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points** and **lines** (line strings), the process considers all the closest pixel centers.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise this process fails with the `TooManyDimensions` error.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
GreatEmerald: How is this different from `aggregate_spatial`? And how is it possible to pass only two values altogether? Is it passing two values for every pair in the list in `aggregate_spatial`?

aggregate_temporal.json
```diff
@@ -209,7 +209,7 @@
     ],
     "exceptions": {
         "TooManyDimensions": {
-            "message": "The number of dimensions must be reduced to three for 'aggregate_polygon'."
+            "message": "The number of dimensions must be reduced to three for 'aggregate_spatial'."
```
GreatEmerald: Should be `aggregate_temporal`?

filter_spatial.json
```diff
@@ -0,0 +1,40 @@
+{
+    "id": "filter_spatial",
+    "summary": "Spatial filter using geometries",
+    "description": "Limits the data cube over the spatial dimensions to the specified geometries.\n\n- For **polygons**, the filter retains a pixel in the data cube if the point at the pixel center intersects with at least one of the polygons (as defined in the Simple Features standard by the OGC).\n- For **points** and **lines** (line strings), the filter retains all the pixel centers closest to any of the geometries.\n\nMore specifically, pixels outside of the bounding box of the given geometry will not be available after filtering. All pixels inside the bounding box that are not retained will be set to `null` (no data).",
```
GreatEmerald: My idea was to only have it work as a crop, so the masking part could be optional (we have a separate process for that, indeed). That might bring significant speed improvements because masking is often times expensive.

filter_spatial.json
```diff
@@ -0,0 +1,40 @@
...
+            "name": "geometries",
+            "description": "One or more geometries used for filtering, specified as GeoJSON.",
+            "schema": {
+                "type": "object",
+                "subtype": "geojson"
+            }
+        }
+    ],
+    "returns": {
+        "description": "A data cube restricted to the specified geometries. Therefore, the cardinality is potentially lower, but the resolution and the number of dimensions are the same as for the original data cube.",
```
GreatEmerald: This should be a list of data cubes. If we have several polygons, then the result should be a data cube per polygon. Otherwise, you would need to change the above to "a single geometry" and allow for some kind of iterator, though that would be quite complex on the process graph...

aggregate_spatial.json
m-mohr: I was also confused by that, but was proposed as it is in one of the calls. Maybe someone can clarify? Maybe @jdries or @mkadunc ?

aggregate_spatial.json
m-mohr: We haven't defined vector cubes at all in openEO and came to the conclusion that we only use them as "black box" for the return value of this process and in no other place until we found a proper definition for us. That's a bit "hacky", but is the best we could come up with due to time constraints.

aggregate_spatial_binary.json
m-mohr: Indeed, fixed.

aggregate_temporal.json
m-mohr: Indeed, fixed.

filter_spatial.json
soxofaan: also see discussion at https://github.com/Open-EO/openeo-processes/pull/131#discussion_r368863643

if it is only about cropping: what is the added value of filter_spatial compared to filter_bbox?

m-mohr: Cropping only works for a rectangles, but there might be values outside a non-rectangular geometry, which need to be set to null.

filter_spatial.json
m-mohr: Why can't it just be a single data cube with all things outside of the geometries set to null?

aggregate_spatial.json
m-mohr: There's only a single data cube this could refer to, the one passed in data. I guess it is logical that this doesn't refer to the cube that is returned as that's nothing you could influence as a user.

aggregate_spatial.json
m-mohr: This was requested by some partners in #101.

aggregate_spatial.json
m-mohr: We don't have the concept of tables or attributes in openEO. So it's probably a dimension in our case.

aggregate_spatial_binary.json
m-mohr: This has a technical background. If lists get to large, especially for UDFs, you'd need to use binary reduction. Yes, it's basically passing the first two values in the list, then the result and the third, then the new result and the fourth and so on...

aggregate_spatial.json
GreatEmerald: Huh, but that one is talking about the raster data cube, not the vector one...

aggregate_spatial.json
GreatEmerald: Hm, but then you have a problem in that the output of aggregate_spatial cannot be used to chain to another aggregate_spatial, because the output types do not match... This can be quite useful if you want to extract values of multiple raster (cubes) and store them all in a single vector (cube).

aggregate_spatial.json
GreatEmerald: But you do also pass the geometry, and that's something the user can influence. But I guess most of the times vectors wouldn't have more than three dimensions, but then again it depends on the vector data cube definition...

aggregate_spatial_binary.json
GreatEmerald: Ah, OK. Would be nice if this could be clarified a bit in the description because it's confusing for me as a user...

filter_spatial.json
GreatEmerald: @soxofaan there is no special added value; my idea is that `filter_spatial` is simply a generalisation of `filter_bbox` that can take more than just a bbox as an input.
@m-mohr Values outside the polygon are just not interesting for the user, but it doesn't meant hat they have to be null. The user just doesn't care about them.

filter_spatial.json
GreatEmerald: Then it will be very sparse. If you use two points, one in Africa and one in North America, you don't want a raster of almost entirely NAs, because you are not interested in those.

aggregate_spatial.json
mkadunc: Closest to at least one out of the infinite points on the line — if the raster grid is square, or if "raster cell" is the voronoi cell of the raster points, then:
- "process considers all the pixels whose centers are closest to at least one point on the line"

 is equivalent to saying
- "process considers all the pixels whose cells intersect the geometry"

...but without mixing up pixel center-points and pixel cells in the same paragraph and going into the whole raster-is-area vs. raster-is-point philosophical debate.

aggregate_spatial.json
m-mohr: Yes, that won't be supported for now or is that something we need for the openEO use cases?

aggregate_spatial.json
m-mohr: Your input is a raster data cube, which the vector data cube will be built upon. Thus the vector cube inherits that restriction.

aggregate_spatial.json
m-mohr: We dont support 4+ dimensional GeoJSON thus it won't be a problem, I hope. We may revisit this when reintroducing vector data cubes to that parameter.

filter_spatial.json
m-mohr: Indeed, so two separate calls to the process would be better. In most cases I hope users will have two spatially restricted data cubes anyway to avoid loading huge datacubes at startup.

aggregate_spatial_binary.json
m-mohr: I'll check how I can do that.

aggregate_spatial.json
GreatEmerald: No, it's not something we would need for the use case, but then for sure this will need to be changed in the future to make sure that the types match. Well, for that matter, the vector data cubes need to be defined for this to work altogether, because that's the return type of the process, unless we also set the return type to GeoJSON...

aggregate_spatial.json
GreatEmerald: The second point is clear to me. It's fine if all backends understand it that way, then.

aggregate_spatial.json
GreatEmerald: What do you mean? The input is a vector geometry, and the process simply adds a dimension to the vector geometry. The raster input is just there as the source of the dimension values, it is read-only.

#101 was actually talking about how the raster data cube gets cropped to the polygons as an optimisation, so that all the rest of the code doesn't need to run on areas that will not be used in the end. But that's already implied (backends should not do work that is not needed) and backend-specific. So I don't see any reason to mention it in the description.

aggregate_spatial.json
GreatEmerald: That's fine, but I don't see any downsides to making the message more specific and read `The number of raster data cube dimensions must be reduced to three for 'aggregate_spatial'.` right now rather than doing it later.

filter_spatial.json
GreatEmerald: Yes, that's fine (if there is a way to iterate over multiple geometries; I'm not completely sure if there is?), but then is there any reason to support multiple geometries in this process? We can just say that it's only for a single geometry. If they do want a sparse raster, they can make a union of the geometry.

aggregate_spatial.json
m-mohr: Yes, there's work needed in the future, probably after the project and/or once the first use cases arrive. We handle the return type of aggregate_spatial as a special case at the moment with save_result being the only process to be able to work on it. It's basically a black box the back-end need to fill for now. This was decided to be done this was on the last calls.

aggregate_spatial.json
m-mohr: @mkadunc Thanks, I updated the process definitions accordingly:

- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).
- For **points**, the process considers the closest pixel center.
- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.

aggregate_spatial_binary.json
m-mohr: I have tried to improve the description.

aggregate_spatial.json
m-mohr: I moved things around a bit in commit https://github.com/Open-EO/openeo-processes/pull/131/commits/cde82763da87d9b7779aadeef79fe56a18a9d6a7, which indeed seems to make a bit more sense. Do you agree, @jdries?

filter_spatial.json
m-mohr: Well, I guess there are valid use cases, e.g. passing the geometries of country borders in Europe and getting zonal statistics for them for a data cube with data of Europe. It really depends on your use case. If you want a pixel in New York and a Pixel in Dubai you are probably better using two separate data cubes, which you could iterate over with array_apply in theory. Or just use user-defined processes and issue two jobs (that's probably what back-ends support first).

You can't really iterate over the output (= vector cube) of the process at the moment. As I said in another comment, we only allow to save the results at the moment. In theory, it would be a call to apply_dimension in the future.

filter_spatial.json
m-mohr: Indeed, but we are defining interoperable processes here and thus need to produce comparable results, which means we need to define such things. It's not desirable that some back-ends may set the parts to null and others leave them in.

m-mohr: @soxofaan filter_bbox is an easy way to specify a bounding box without needing to mess around with GeoJSON geometries. I just reaalized that the major difference between filter_bbox and filter_spatial is that filter_spatial only works on WGS84 (due to GeoJSON) and filter_bbox allows different CRS. This difference will be weakened once vector cubes are defined though.

PR #133, opened 2020-01-21
Added support for labeled arrays

m-mohr:
See issue https://github.com/Open-EO/openeo-api/issues/245. 

Related PR in API: https://github.com/Open-EO/openeo-api/pull/255.

Also renames processes trim to trim_cube and labels to dimension_labels for clarity.

~Based on PR #132, which needs to be merged first.~


array_apply.json
```diff
@@ -34,11 +34,16 @@
                     },
                     {
                         "name": "index",
-                        "description": "The zero-based index of the current element being processed.",
+                        "description": "The zero-based index or the label of the current element being processed. Labels take preference over indices.",
```
mkadunc: This is weird. Either `array_apply` should work only on plain arrays and forward index, or it should work only on labeled arrays and forward label (and maybe index). In any case it should not be one or another at the backend's discretion.

array_filter.json
```diff
@@ -35,11 +35,16 @@
                     },
                     {
                         "name": "index",
-                        "description": "The zero-based index of the current element being processed.",
+                        "description": "The zero-based index or the label of the current element being processed. Labels take preference over indices.",
```
mkadunc: See above comment against `array_apply` - we should not let the backend decide whether to pass label or index - this should be unambiguously and explicitly defined and clear to the user. I suggest adding a `label` parameter to the callback, which will be populated with the label in case the input array is labeled.

array_find.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "array_find",
-    "summary": "Get the index for a value in an array",
-    "description": "Checks whether the array specified for `data` contains the value specified in `value` and returns the zero-based index for the first match. If there's no match, `-1` is returned.\n\n**Remarks:**\n\n* To get a boolean value returned use ``array_contains()``.\n* All definitions for the process ``eq()`` regarding the comparison of values apply here as well. This includes that data types MUST be checked strictly, for example a string with the content *1* is not equal to the number *1*.\n* An integer *1* is equal to a floating point number *1.0* as `integer` is a sub-type of `number`. Still, this process may return unexpectedly `false` when comparing floating point numbers due to floating point inaccuracy in machine-based computation.\n* Temporal strings are treated as normal strings and MUST NOT be interpreted.",
+    "summary": "Get the index/label for a value in an array",
```
mkadunc: I think changing the meaning of the returned value like this, depending on the input, is bound to cause issues down the line.

We could remove the ambiguity by always referring to labels (and have them default to 0,1,2,.... The only place where you would then need to mention unlabelled arrays is maybe in constructors.

Another option is to remove both index and label as return of the `array_find`.

Another option is to have two processes - `array_find_label` and `array_find_index`.

Another option is to always return index, and let user fetch the label - e.g. `array_label(array_find(arr, 'someValue'))`...

array_apply.json
m-mohr: It's not at the backend's discretion: "Labels take preference over indices."
So whenever a label is available (i.e. an array of subtype `labeled-array` has been passed) the label has to be returned instead of the array.

array_filter.json
m-mohr: See above.

array_find.json
m-mohr: > I think changing the meaning of the returned value like this, depending on the input, is bound to cause issues down the line.

Why's that?

> We could remove the ambiguity by always referring to labels (and have them default to 0,1,2,.... 

I don't understand how that solves the problem. It's just a different description of the behavior, which I understood above is causing issues down the line?!

> Another option is to remove both index and label as return of the `array_find`.

What is returned instead?

> Another option is to have two processes - `array_find_label` and `array_find_index`.

That seems like a bit too much.

> Another option is to always return index, and let user fetch the label - e.g. `array_label(array_find(arr, 'someValue'))`...

We don't have array_label yet and `array_label(arr, array_find(arr, 'someValue'))` seems rather cumbersome, it's basically another form of array_element again.

array_apply.json
m-mohr: I added an optional label parameter to the callbacks in array_apply and array_filter.

array_find.json
mkadunc: > @mkadunc Why's that?

Well, since labeled-array extends array, this method violates the [Liskov Substitution Principle (LSP)](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

An example of the consequence of LSP violation: Imagine that I define a custom process, say `getCurvatureAtMaximum(array)`, without worrying (or knowing) about labelled-arrays (i shouldn't have to worry about them because they're an extension of plain arrays) as:

```
const h = 10; //meters

function getCurvatureAtMaximum(array) {
    const maxVal =max(array);
    const idxCenter = array_find(maxVal);
    if (idxCenter == 0 || idxCenter == array.length-1) {
        return null;   
    }
    const leftVal = array[idxCenter - 1];
    const rightVal = array[idxCenter + 1];
    return leftVal - 2*maxVal + rightVal;
}
```

If I call this with `[0.3, 0.9, 0.5, 0.2, 0.8]`, I get `0.3 - 2*0.9 + 0.5 = -1`.

If I call this with same values, but labelled — `[1 : 0.3,  3 : 0.9,  5 : 0.5, 7 : 0.2, 9 : 0.8]` — I rightfully expect the same result (the function promises curvature at maximum for arrays, and labelled-array is an array, so nothing wrong with my expectation). The result I get is `|array[3-1] - 2 * 0.9 + array[3+1]| == 0.5 - 2*0.9 + 0.8 == -0.5`.

If I call this with some array that has String labels or one that has different numeric labels, the function either fails because string and integer cannot be subtracted or added, or because of a null pointer when retrieving `leftVal` or `rightVal`.

>> We could remove the ambiguity by always referring to labels (and have them default to 0,1,2,....
>
> I don't understand how that solves the problem. It's just a different description of the behavior, which I understood above is causing issues down the line?!

It solves the problem by changing the signature of the function from `array_find(data::array, value::any) returns index` to `array_find(data::labeled-array, value::and) returns label` — by 'specialising' this function to work only on labeled arrays, and always returning only labels, we remove the possibility of a labeled-array being used inadvertently in some process that expects this function to return the ordinal index.

If we stay with the current implementation, this function is practically unusable for any use that expects *index* to be returned (such as my example above) - in order to force it to return index, one would always need to strip any labels from the array (even in case the array is declared to be `array`) - my above code would look like this:

```
const h = 10; //meters

function getCurvatureAtMaximum(array) {
    const strippedArray = (array instanceof labeled-array) ? array.stripLabels() : array;
    const maxVal =max(strippedArray);
    const idxCenter = array_find(maxVal);
    if (idxCenter == 0 || idxCenter == strippedArray.length-1) {
        return null;   
    }
    const leftVal = strippedArray[idxCenter - 1];
    const rightVal = strippedArray[idxCenter + 1];
    return leftVal - 2*maxVal + rightVal;
}
```

This code is smelly, because it couples functionality that is conceptually simple (only deals with `arrays`), with the extended concept of labeled arrays. This, in addition to being more error-prone (because one needs to remember to do such things), causes more complexity in the code organization and implementation than necessary.

m-mohr: Thanks for taking the time to write this, @mkadunc .

I agree with you on the problems regarding the processes array_find and so on, but don't agree regarding the getCurvatureAtMaximum example and the LSP violation. This is probably due to me not documenting labeled arrays properly. While the problem of the example is the flawed array_find, labeled arrays as such would work in the example. My intention was to add labels as addition to normal arrays thus the labeled array still has the indices and left and right values are still known. Therefore, I don't think LSP is broken here. A dumb class structure would be:

```java
class ArrayObj {
  protected Array[] data;
}
class LabeledArray extends ArrayObj {
  protected Array[] labels;
}
```

I'm not 100% sure which solution I like most. Maybe adding a boolean parameter get_label with default false to the processes with that problem. Or the array_label process although I'm thinking what more user friendly would be. Maybe both?! What do you think?

mkadunc: I agree - the LabeledArray class is OK, and the problem is in the `array_find` (whether or not this is LSP violation or some other smell is philosophical, I only mentioned it because it looks similar and IMO has similar consequences).

---

Re `get_label` parameter: I'm not a fan of flags that fundamentally change the behaviour / output of functions ([a lot of arguments for that can be found in this stackexchange post](https://softwareengineering.stackexchange.com/questions/147977)).

In this case, however, it could work if we can produce a reasonable output object that can represent both index and a label, e.g.: `{index: 1, label: 3}` for labeled-array inputs, and `{index: 1}` for non-labeled array inputs (this would mean we have covariance of the return type).

We could prevent some uncaught problems by returning less-coercable types (`number` for index and `{label: 3}` for label) and hope that runtime exceptions will alert the developer to issues...

The preferred option for me would be to always return the index in `array_find` and then, if one wants the label, either support `array_label(labeled-array, index)` or `array_labels(labeled_array)[index]`.

---

As for `array_element`, the object-for-label vs. number-for-index approach works better, i.e. we can support both:
- index-based array access: `labeled-array[1]` (or, with process instead of `[]` syntax: `array_element(labeled-array, 1)`)
- label-based array access `labeled-array[{label: 3}]`, ( `array_element(labeled-array, {label: 3})`)

m-mohr: I see the point, but the other alternatives are also not optimal IMHO.

The object thing may work for array_element, but it doesn't really work for array_find. We don't have processes to deal with objects at all and thus a user can't further handle an object returned by array_find. For array_element it puts the burden to the clients and they need to deal with it, which may result in a boolean flag again. 

Having more processes will work and is probably the cleanest solution although we're getting slammed with processes. It's getting more and more, but maybe that's the way it is?!

Having array_label and array_labels seems likely to confuse users as it's just an additional s in the process name.

Hmm...

Issue #137, opened 2020-01-27
Create process graphs for "alias" processes


m-mohr:
Based on https://github.com/Open-EO/openeo-api/issues/161 / https://github.com/Open-EO/openeo-api/pull/256: Create process graphs for "alias" processes (neq, ndvi, ...).

m-mohr:
- [ ] absolute
- [ ] add
- [ ] and
- [ ] between
- [ ] clip
- [ ] exp
- [ ] gte
- [ ] linear_scale_range
- [ ] ln
- [ ] lte
- [ ] multiply
- [ ] neq
- [ ] normalized_difference
- [ ] or
- [ ] round
- [ ] sd
- [ ] sgn
- [ ] sort
- [ ] sqrt
- [ ] absolute
- [ ] variance
- [ ] xor

- [ ] mask?
- [ ] sin/cos/...?

PR #140, opened 2020-01-29
Process graphs for pre-defined processes

m-mohr:
Added process graphs that can be described with other processes.

See issue #137


absolute.json
```diff
@@ -59,5 +59,26 @@
...
+            "arguments": {
+                "base": {
+                    "from_node": "power2"
+                },
+                "p": 0.5
+            },
+            "result": true
+        }
+    }
 }
```
mkadunc: As a default implementation, I would prefer `|x| = x>= 0 ? x : -1 * x`, mostly for performance reasons, but also because it is closer to everyday understanding of `abs`...

array_contains.json
```diff
@@ -138,5 +138,35 @@
...
+            "process_id": "not",
+            "arguments": {
+                "x": {
+                    "from_node": "is_nodata"
+                }
+            },
+            "result": true
+        }
+    }
 }
```
mkadunc: This implementation is inconsistent with the example `array_contains([1, 2, null], null) => true`

variance.json
```diff
@@ -90,5 +90,28 @@
...
+            "process_id": "power",
+            "arguments": {
+                "base": {
+                    "from_node": "sd"
+                },
+                "p": 2
+            },
+            "result": true
+        }
+    }
```
mkadunc: Is it OK to have cyclic aliases (`variance` references `sd`)? Maybe we should write variance as
```
variance(data) = {
    avg = mean(data);
    return sum(apply(data, val => pow(val - avg, 2)));
}
```

m-mohr: Yes, that's okay, although it's always better to have it more low-level.

absolute.json
m-mohr: Makes sense, changed that.

array_contains.json
m-mohr: There's a more general problem, too. We refer to eq, but eq on null always returns null.

variance.json
m-mohr: Problem is: I can't access avg in the apply callback.

array_contains.json
m-mohr: The question is how to best fix it:
1. array_find/contains can't search for null values
2. Rephrase `eq` to say "If exactly one of the operands is `null`, the return value is `null`." instead of "If any operand is `null`, the return value is `null`.". Would mean `eq(null, null)` => `true`, but `eq(null, 1)` => `null`. This would make array_find/contains work as expected.

mkadunc: I'd go with 1) . `eq(null, null)` should return `null` (e.g. for booleans, eq is the as not(xor(null, null))) 

IMO `array_find` should not try and search for `null`s, but just return `null` if the searched value is `null`. If one really wanted to find the (first) index of a missing value, we could provide a more general process (`array_findfirst` or `array_select`) that takes a predicate (callback) and returns the first index where the predicate returns `true` - by using isValid in the predicate you can get the first `null`...

m-mohr: On the other hand, you can also use eq in the property filters in load_collection for example and there null is not necessarily no-data and returning null could be somewhat unexpected.

m-mohr: I think array_find and array_contains for null values can be emulated as follows:

array_find: `array_find(array_apply(data, x => is_nodata(x)), true)`
array_contains: `array_contains(array_apply(data, x => is_nodata(x)), true)`

variance.json
m-mohr: For 1 there's a PR here: #142

array_contains.json
m-mohr: I'll leave this open for discussion until the end of the day. If there's no additional motion, I'll implement 1.

variance.json
m-mohr: I'll leave this open for discussion until the end of the day. If there's no additional motion, I'll implement option 1.

array_contains.json
jdries: Fine with #1

variance.json
jdries: Not sure here, I'm fine with option 1 as long as it's backwards compatible. (So everything that currently does not need context simply continues to work.)

m-mohr: Process graphs and most processes are not really backward-compatible anyway, but options 1 and 2 are both "backward-compatible" in a sense that they don't break things compared to the state before merging this PR. Worst issue I see with 2 is that it get's harder to resolve dependencies in the graph and it need's an adoption of the parser, whereas 1 doesn't need a change in process graph parsing or the API at all.

array_contains.json
m-mohr: For option 1 there's a PR here: #143, including examples for finding null values.
Do you agree, @mkadunc ?

sgn.json
```diff
@@ -58,5 +58,44 @@
             "href": "http://mathworld.wolfram.com/Sign.html",
             "title": "Sign explained by Wolfram MathWorld"
         }
-    ]
+    ],
+    "process_graph": {
+        "gt0": {
```
soxofaan: `process_id` and `arguments` are missing here? for `lt0` too?

sgn.json
```diff
@@ -58,5 +58,44 @@
...
+            "y": 0
+        },
+        "if_gt0": {
+            "process_id": "if",
+            "arguments": {
+                "value": {
+                    "from_node": "gt0"
+                },
+                "accept": 1,
+                "reject": {
```
soxofaan: why not just `"reject" : 0`  ? With `"from_parameter": "x"` it looks bit confusing

variance.json
soxofaan: > 2. access of parent nodes in callbacks?

what do you mean with this? In the variance example, the `avg` node would not be a parent, but more some kind of sibling node of the apply callback node, right?

sgn.json
m-mohr: It can be either 0 or null and that's what is in "from_parameter": "x" after the lt/gt checks.

sgn.json
m-mohr: Whoops, fixed.

m-mohr:
> what do you mean with this? In the variance example, the avg node would not be a parent, but more some kind of sibling node of the apply callback node, right?

@soxofaan 
Looking at Mihas example:
```
    avg = mean(data);
    return sum(apply(data, val => pow(val - avg, 2)));
```
the following is a child process graph:
`pow(val - avg, 2)`

The process graph specification only allows access to nodes in the same process graph (i.e. no access to children or parents). avg was computed in the "root" scope, but is accessed in the "child" scope and this is undefined in the "child" scope. Therefore, you either need to broaden the scope for from_node or introduce an explicit was to pass computed data, e.g. via a context parameter.

variance.json
m-mohr: > what do you mean with this? In the variance example, the avg node would not be a parent, but more some kind of sibling node of the apply callback node, right?

@soxofaan 
Looking at Mihas example:
```
    avg = mean(data);
    return sum(apply(data, val => pow(val - avg, 2)));
```
the following part is a child process graph:
`pow(val - avg, 2)`

The process graph specification only allows access to nodes in the same process graph (i.e. no access to children or parents). avg was computed in the "root" scope, but is accessed in the "child" scope and this is undefined in the "child" scope. Therefore, you either need to broaden the scope for from_node (option 2) or introduce an explicit was to pass computed data, e.g. via a context parameter (option 1). Is this better to understand?

soxofaan: ok thanks, I understand, I just misread a part apparently :)

array_contains.json
mkadunc: I agree with proposed changes

variance.json
m-mohr: Hey @mkadunc,

my statistics classes were long time ago, but the sum in your example must be replaced with mean, I think?!

The process graph would then be:
```json
{
        "mean": {
            "process_id": "mean",
            "arguments": {
                "data": {
                    "from_parameter": "data"
                }
            }
        },
        "apply": {
            "process_id": "apply",
            "arguments": {
                "data": {
                    "from_parameter": "data"
                },
                "process": {
                    "process-graph": {
                        "subtract": {
                            "process_id": "subtract",
                            "arguments": {
                                "x": {
                                    "from_parameter": "x"
                                },
                                "y": {
                                    "from_parameter": "context"
                                }
                            }
                        },
                        "power": {
                            "process_id": "power",
                            "arguments": {
                                "base": {
                                    "from_node": "subtract"
                                },
                                "p": 2
                            },
                            "result": true
                        }
                    }
                },
                "context": {
                    "from_node": "mean"
                }
            }
        },
        "mean2": {
            "process_id": "mean",
            "arguments": {
                "data": {
                    "from_node": "apply"
                },
                "ignore_nodata": {
                    "from_parameter": "ignore_nodata"
                }
            },
            "result": true
        }
    }
```

mkadunc:
I didn’t even have statistics classes ;) but mean sounds about right

PR #145, opened 2020-03-11
Processes for climatology use cases

m-mohr:
Processes for climatology use cases as defined with @jwagemann this week.

Reviews are appreciated.

cc: @claxn


m-mohr:
This PR is now open for final review. @jwagemann

aggregate_temporal_frequency.json
```diff
@@ -0,0 +1,112 @@
...
+            "description": "Additional data to be passed to the reducer.",
+            "schema": {
+                "description": "Any data type."
+            },
+            "optional": true,
+            "default": null
+        }
+    ],
+    "returns": {
+        "description": "A data cube with potentially lower resolution and cardinality, but the same number of dimensions as the original data cube. The specified temporal dimension has the following dimension labels (`YYYY` = four-digit year, `MM` = two-digit month, `DD` two-digit day of month):\n\n* `hourly`: `YYYY-MM-DD-0` - `YYYY-MM-DD-23`\n* `daily`: `YYYY-1` - `YYYY-365`\n* `weekly`: `YYYY-1` - `YYYY-52`\n* `monthly`: `YYYY-1` - `YYYY-12`\n* `yearly`: `YYYY`\n* `seasons`: `YYYY-djf` (December - February), `YYYY-mam` (March - May), `YYYY-jja` (June - August), `YYYY-son` (September - November).\n* `tropical_seasons`: `YYYY-ndjfma` (November - April), `YYYY-mjjaso` (May - October).",
```
soxofaan: maybe more consistent to always same number of digits for hours (`00`-`23`), days (`001`-`365`), ...

climatological_normal.json
```diff
@@ -0,0 +1,68 @@
...
+            },
+            "default": [
+                "1981",
+                "2010"
+            ],
+            "optional": true
+        }
+    ],
+    "returns": {
+        "description": "A data cube with potentially cardinality, but the same resolution and number of dimensions as the original data cube. The temporal dimension has the following dimension labels:\n\n* `daily`: `1` - `365`\n* `monthly`: `1` - `12`\n* `climatology_period`: `climatology_period`\n* `seasons`: `djf` (December - February), `mam` (March - May), `jja` (June - August), `son` (September - November)\n* `tropical_seasons`: `ndjfma` (November - April), `mjjaso` (May - October)",
```
soxofaan: > A data cube with potentially cardinality,

lower cardinality?

soxofaan: for daily/monthly: are the labels strings or itegers?

anomaly.json
```diff
@@ -0,0 +1,53 @@
+{
+    "id": "anomaly",
```
soxofaan: seems like very generic name for a specific use case. Shouldn't this be namespaced a bit, like e.g. " climatological_normal_anomaly"?

aggregate_temporal_frequency.json
```diff
@@ -0,0 +1,112 @@
+{
+    "id": "aggregate_temporal_frequency",
```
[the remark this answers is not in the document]
edzer: I'd suggest `aggregate_temporal_period`.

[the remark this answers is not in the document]
m-mohr: This originates from `climatological_normal`. First we had two parameter there `period` and `climatology_period`, but it felt confusing as there's an "unnamed" period together with climatology_period so we tried to avoid that and the best solution we found was frequency. As these processes are often used together, we decided to use frequency in all to avoid confusion. I guess we could revert to period if that feels better overall. Maybe the two period parameters are not as confusing?

aggregate_temporal_frequency.json
m-mohr: That comes from aligning with climatological_normal, where we store integer labels and thus 001 would be simply 1...

climatological_normal.json
m-mohr: We implemented it for integers, but in general I didn't specify that to support cases where back-ends may only support strings.

anomaly.json
m-mohr: We talked to TU Vienna and they said the process is broad enough to also support EO use cases so it's not only meant for a specific use case.

climatological_normal.json
m-mohr: cardinality in the processes refers to the number of dimension labels, but it seems to confuse users. Maybe I should generally replace it just with "number of dimension labels" in all processes?!

claxn: Yes, I agree with you @m-mohr , from my point of view cardinality always refers to the dimensions, not the number if labels along a dimension.

anomaly.json
```diff
@@ -0,0 +1,53 @@
...
+                    "hourly",
+                    "daily",
+                    "weekly",
+                    "monthly",
+                    "yearly",
+                    "single_period",
+                    "climatology_period",
+                    "seasons",
+                    "tropical_seasons"
+                ]
```
claxn: How about adding `decadal` to the list? This could be somehow attractive for climate/weather analysis.

aggregate_temporal_frequency.json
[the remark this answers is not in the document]
claxn: Daily is still a period from 0-24h, but is for sure less often used in EO than e.g. monthly or seasonal for higher level analysis, I think. I am fine with frequency, because I am often using Pandas (e.g., [Pandas Date Range](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.date_range.html)) and there frequency is also a common term.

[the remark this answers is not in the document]
m-mohr: Daily was an example, could also be yearly. What I meant is, it not just a single exactly specified period like 2010-2011, but here it is describing how something is split into multiple exact periods depending on the input data: 2010-2011, 2011-2012, 2012-2013, ...

anomaly.json
m-mohr: I asked @jwagemann about decades, but we couldn't really come up with a good use case, but sure, we can add it if you have a use case in mind? :-)

climatological_normal.json
m-mohr: I'll solve this in a separate PR. Here's the related issue: #149

aggregate_temporal_frequency.json
soxofaan: I meant these, where offsets are converted to string representation:

> `hourly`: `YYYY-MM-DD-0` - `YYYY-MM-DD-23`
> `daily`: `YYYY-1` - `YYYY-365`
> `weekly`: `YYYY-1` - `YYYY-52`
> `monthly`: `YYYY-1` - `YYYY-12`

mismatch between alphabetical and logical sorting will  cause headaches somewhere down the line

m-mohr: Yes, I understood that. Still, for consistentcy we'd should then pad the numbers in the other processes, too.

soxofaan: > we'd should then pad the numbers in the other processes, too

certainly, I added this comment to this line, but it's also about all other places where this pattern occurs of concatenating ints in string.

Also, I assume the `YYYY`, `MM` and `DD` are also zero-padded, so I would just do it for all components of these strings (where possible)

anomaly.json
soxofaan: FYI: last weeks I worked on a use case (not openeo based so far)  involving decad based vegetation monitoring. The goal was basically to produce some sort of temporally smoothed NDVI tile each decad based on a larger time window (several months)

aggregate_temporal_frequency.json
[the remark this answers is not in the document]
m-mohr: Renamed frequency to period throughout the PR. Also, it's now day instead of daily, season instead of seasons etc.

aggregate_temporal_frequency.json
m-mohr: Changed this in the last commit, please check. I'll check the implementation for potential issues.

climatological_normal.json
m-mohr: With the recent commits, all labels are strings now.

anomaly.json
m-mohr: Added "decade" as frequency to the processes (except climatological_normal), please check.

soxofaan: Sorry, in my use case I referring to 10 **day** periods  (three per month: [days 1-10], [days 11-20] and [days 21-31], so 36 per year), and assumed @claxn  was referring to that as well. Compared to the well known "decade" (10 years) it's a more obscure period (I just learned about it couple of weeks ago)  using confusion-prone terminology "dekad" and "dekadal". I have no idea how popular it is and whether it is worth including in the enum.

m-mohr: Oh, I thought decades (10 years) were meant. What was your intention @claxn ?

claxn: My intention was to use it for 10 years for climate science, because there "decadal" is a well-known term. However, we have also had some "decadal" (10 days, exactly like @soxofaan described) use cases in remote sensing, so it seems this is more likely to be used. However, if we use the term "decadal" it could confuse a user. I have googled a bit and found this: 
[10 days](https://english.stackexchange.com/questions/363469/is-there-any-word-for-10-days-in-english) 
How about "tri-monthly"/"trice-monthly"? I think this is very clear then. Should we also include "decadal" (10 years) or drop it? I don't have a use case for it right now...

m-mohr: For tri-monthly I'd need to revert the changes to period/frequency and month/monthly.

Are both 10 years and 10 days common use cases we should cover?

I'm thinking whether it would make sense to integrate another parameter that allows to specify an amount (defaulting to 1), e.g.
period = day, amount = 10
period = year, amount = 10
But maybe that makes implementation too complex? Maybe we should just use "10-days" and "decade" or "10-years" as option?

soxofaan: FYI: note that with the "10 day" dekads, the last dekad of the month does not always have exactly 10 days: it can range from 8 to 11.

So if you call it a "10-day" period, the user might expect exact 10 day periods, instead of the "monthly aligned, roughly 10 day" periods. But again, this concept is new to me as well, I have no idea how well know/settled all this is.

claxn: If @jwagemann does not think decadal is urgently needed for climate and meteorological use cases, then I would go for 10 days at the moment. We can add it later if need be. Our use case @ TUW was just to have a 10 day composite of satellite imagery to close some data gaps. 

I prefer tri-month(ly) over 10-day, because for a longer time series a 10 day step is breaking desired temporal boundaries, e.g. end of a month, end of a year etc.

m-mohr: True. Best would be if there's a well-known name. tri-monthly also doesn't really tell how things are split. So it depends on good documentation anyway.

m-mohr: True, "10-day" is misleading. Best would be if there's a well-known name. tri-monthly also doesn't really tell how things are split. Could even mean "every three months". So it needs good documentation anyway. What I just read is "[decadaily](https://english.stackexchange.com/questions/64086/how-do-you-say-three-times-a-month-in-one-word)". The thread also discusses tri-monthly as confusing word due to every three months vs. month split in 3 parts.

m-mohr: I opened issue #150 to track the decadal issue, we may add it in a separate PR. Please keep the discussion going there. Merging without decades for now to get things started.

PR #157, opened 2020-06-18
Add `constant` process

soxofaan:
While working on support of user-defined processes it felt  that a `constant` process could be handy to avoid too much hardcoding of predefined values. With the `constant` process, you introduce a constant value node in the process graph, that you can reuse with `from_node`, instead of having to copy-paste that value all over the place. For standard openEO usage that is probably not necessary as the client probably hides that away, but for handcrafted user-defined processes it might be valuable.

Basic usage example:
```
                "margin1": {
                    "process_id": "constant",
                    "arguments": {
                        "x": 0.1
                    }
                },
                "west1": {
                    "process_id": "subtract",
                    "arguments": {
                        "x": 3.5,
                        "y": {
                            "from_node": "margin"
                        }
                    }
                },
                "east1": {
                    "process_id": "add",
                    "arguments": {
                        "x": 3.8,
                        "y": {
                            "from_node": "margin"
                        }
                    }
                }

```

### DIFF
--- a/constant.json
+++ b/constant.json
@@ -1,7 +1,7 @@
 {
     "id": "constant",
     "summary": "Define a constant value",
-    "description": "Define a constant value that can be reused in multiple places of the process graph",
+    "description": "Defines a constant value that can be reused in multiple places of a process.",
     "categories": [
         "math > constants"
     ],
@@ -10,46 +10,14 @@
             "name": "x",
             "description": "The value of the constant.",
             "schema": {
-                "type": "number"
+                "description": "Any data type."
             }
         }
     ],
     "returns": {
         "description": "The value of the constant.",
         "schema": {
-            "type": "number"
+            "description": "Any data type."
         }
-    },
-    "examples": [
-        {
-            "title": "Reuse of a constant value",
-            "process_graph": {
-                "margin1": {
-                    "process_id": "constant",
-                    "arguments": {
-                        "x": 0.1
-                    }
-                },
-                "west1": {
-                    "process_id": "subtract",
-                    "arguments": {
-                        "x": 3.5,
-                        "y": {
-                            "from_node": "margin"
-                        }
-                    }
-                },
-                "east1": {
-                    "process_id": "add",
-                    "arguments": {
-                        "x": 3.8,
-                        "y": {
-                            "from_node": "margin"
-                        }
-                    }
-                }
-            }
-        }
-    ],
-    "links": []
+    }
 }
--- a/constant.json
+++ b/constant.json
@@ -1,0 +1,55 @@
+{
+    "id": "constant",
+    "summary": "Define a constant value",
+    "description": "Define a constant value that can be reused in multiple places of the process graph",
+    "categories": [
+        "math > constants"
+    ],
+    "parameters": [
+        {
+            "name": "x",
+            "description": "The value of the constant.",
+            "schema": {
+                "type": "number"
+            }
+        }
+    ],
+    "returns": {
+        "description": "The value of the constant.",
+        "schema": {
+            "type": "number"
+        }
+    },
+    "examples": [
+        {
+            "title": "Reuse of a constant value",
+            "process_graph": {
+                "margin1": {
+                    "process_id": "constant",
+                    "arguments": {
+                        "x": 0.1
+                    }
+                },
+                "west1": {
+                    "process_id": "subtract",
+                    "arguments": {
+                        "x": 3.5,
+                        "y": {
+                            "from_node": "margin"
+                        }
+                    }
+                },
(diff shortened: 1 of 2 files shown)


m-mohr:
Why not use a process parameter with a default value?

soxofaan:
Note: I'm not sure if the example in the json file is properly structured, but I could not find another process that has an example with multiple process graph nodes

soxofaan:
> Why not use a process parameter with a default value?

because sometimes you really want a constant value that can not be changed from the outside. For example we currently have `e` and `pi`, but some domains might have their own constants that should not be a "variable"

m-mohr:
At least it should work:

```json
{
  "parameters": [
	{
	  "name": "margin",
	  "default": 0.1
	}
  ],
  "process_graph": {
    "3": {
      "process_id": "add",
      "arguments": {
        "y": {
          "from_parameter": "margin"
        },
        "x": 3.8
      },
      "result": true
    },
    "4": {
      "process_id": "subtract",
      "arguments": {
        "y": {
          "from_parameter": "margin"
        },
        "x": 3.5
      }
    }
  }
}
```

soxofaan:
> Why not use a process parameter with a default value?

Or another reason: maybe the constant is calculated from other input. E.g.: you calculate the width of an input bbox and want to reuse that width in several places

m-mohr:
I see that this could be useful and I'm not against adding this, just want to show that this somehow already works.
In most cases the computations are done in child processes and then the parameters can't be changed if defined like this, but I see that it can be useful to have it really unchangeable and also it is not the most obvious solution for a user.

> Or another reason: maybe the constant is calculated from other input. E.g.: you calculate the width of an input bbox and want to reuse that width in several places

I don't get this point though. How would a constant help with this? Why not access the calculation directly?

soxofaan:
>  maybe the constant is calculated from other input. 

yeah sorry, I just realized this is already possible. Scratch that use case :)

constant.json
```diff
@@ -0,0 +1,55 @@
...
+                        "x": 3.8,
+                        "y": {
+                            "from_node": "margin"
+                        }
+                    }
+                }
+            }
+        }
+    ],
+    "links": []
```
m-mohr, line 54: Can be removed, it not a required field.

m-mohr, line 54: Can be removed, it's not a required field.

constant.json
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+    ],
+    "returns": {
+        "description": "The value of the constant.",
+        "schema": {
+            "type": "number"
+        }
+    },
+    "examples": [
```
m-mohr, line 23: Support for process graph examples has been removed. Such examples use links now.

constant.json
```diff
@@ -0,0 +1,55 @@
...
+    "description": "Define a constant value that can be reused in multiple places of the process graph",
+    "categories": [
+        "math > constants"
+    ],
+    "parameters": [
+        {
+            "name": "x",
+            "description": "The value of the constant.",
+            "schema": {
+                "type": "number"
```
m-mohr, line 13: I think it makes sense to allow any data type here, right? Defining a larger array or string once makes sense, too.

constant.json
```diff
@@ -0,0 +1,55 @@
+{
+    "id": "constant",
+    "summary": "Define a constant value",
+    "description": "Define a constant value that can be reused in multiple places of the process graph",
```
m-mohr, line 4:
```suggestion
    "description": "Defines a constant value that can be reused in multiple places of a process.",
```

We don't use the process graph phrase in user communication any longer.

constant.json
soxofaan: I couldn't come up with a good use case for strings, but arrays make sense indeed

m-mohr: Maybe something like a re-occurring band or dimension name? Not sure, but also don't see a reason to restrict this to any data type (although boolean and null seems really weird).

constant.json
soxofaan: I also should find a better example than the one with "margin" because parameters feel more suited for it than a constant indeed

constant.json
m-mohr: I think the schema should simply be:

```json

            "schema": {
                "description": "Any data type."
            },
```

I could also find useful use cases for objects, for example.

Also, the return value also needs to be changed.

constant.json
m-mohr: A dot is still missing at the end of the line ;-)

soxofaan:
squashed and force-pushed some additional tweaks

m-mohr:
Thanks, has been merged and will be part of 1.0.

Issue #154, opened 2020-06-09
Support applying a process to a neighbourhood of pixels


jdries:
Use case: run a machine learning model that works on 128x128 'patches' or 'tiles' of EO data, and retain only the 64x64  inner pixels as a result. So the resulting datacube is the combination of these 64x64 results, that can be computed in parallel.

Possible solution:
This sounds like an apply_dimension, but:
- x and y are separate dimensions, while we want to work on both of them
- it is not necessary to work on the 'full' space dimension, this could be very large, we simply want a small window

So:
- allow specifying more than one dimension in apply_dimension?
- or create a separate function to work on a 'window' of data. Where the extent and overlap of windows needs to be specified for each dimension. 

So my window would be:

1. all bands
2. 64 pixels along x and y, padded with 32 pixels on each side
3. 1 date

Similar function in xarray:
First 'chunk' the dataset:
http://xarray.pydata.org/en/stable/generated/xarray.Dataset.chunk.html
Then apply function to each chunk:
http://xarray.pydata.org/en/stable/generated/xarray.map_blocks.html

But also rolling:
xarray.DataArray.rolling

In fact I believe I need a combination of rolling and chunking in my use cases. Rolling provides the ability to 'pad' data in space and time, chunking allows me to for instance specify that I want to transform 64x64 tiles.

Earth Engine has this creative solution:
ee.Image.neighborhoodToBands
Which would help in some cases: for a time window, we could add the required dates as bands, maybe even in a udf. Then a reduction of the bands dimension would also have temporal info.

m-mohr:
Isn't that exactly a use case for apply_kernel? Although I acknowledge that the `kernel` array is getting pretty large.

jdries:
It's certainly related to apply_kernel, but that process is restricted to 2D convolutions, while I want to do a more generic transformation. So I would maybe not overload apply_kernel with this behavior.

m-mohr:
I think it was supporting your usecase in 0.4, but we restricted it for 1.0 (can't remember why).

I'm still not 100% sure what is missing from apply_kernel so maybe you could propose a draft process in a PR? We could discuss that in the next dev meeting or so. I feel like it should be a separate process to also not overload apply_dimension.

jdries:
I added a pull request for a 'chunk' process. Advantage is that we do not need to modify/overload existing processes.
[...]

PR #160, opened 2020-06-24
#154 specify a chunk process

jdries:
Proposal for a 'chunk' process.


edzer:
This is more like setting a parameter for a follow-up process, right? I don't think of "chunking" as a process (what would be the outcome of it?), but as a strategy.

jdries:
Indeed, it would not have a visible effect. Although a backend that uses chunking internally could do things, but it would not be visible.
You could say that it makes 'chunks' a user visible property of our data cubes. Whereas until now it was probably there, but hidden. 
Maybe a 'positive' side effect is that this allows backends to specify their support for chunks through the processes API. If we simply add an optional chunking parameter to existing processes, backends would maybe not support it, but it would not be visible for the user...
(I'm myself not entirely decided on the best approach, just worked out one of the options to get some discussion going.)

m-mohr:
Ah, now I understand #154 better. But now I'm wondering: Isn't that what we always said to abstract away from a user? How would a user know when to use this and what to set for the parameters? Especially this "physical measure" mapping seems unclear.

edzer:
Yes, ideally, the back-end would be smart enough to sort this out, but I believe that in practice this is usually rather black magic (= trial and error).

jdries:
The designer of the algorithm would know that he needs some padding for his data. Or for instance, for a machine learning model, he may have designed it to run on 128x128 patches spatially, and a 2 month period in time. (Those are the use cases that I'm running into.)

How to specify the chunks/padding indeed needs improvement. For a time dimension, people often use a 'physical' representation like x days, while for spatial it is more common to use pixels. (As people mostly design for a given sensor with a known resolution.)

m-mohr:
Oh, I see. In the description it sounded more like a hardware than a algorithm issue though?!

Are there other similar options that we may need? I'm thinking whether this could be something to be set in load_collection or save_result. Having it as part of save_result would be pretty easy to integrate as we have format parameters etc. already integrated and so each back-end could handle that individually without any API/process change.

jdries:
Problem is you might have multiple operations with different chunking/padding options. So having it in load_collection would work for the first operation, but not for the next. In save_result makes less sense, as it would give the impression that it's a property of the output format.
If adding to specific processes, the list would be fairly limited:
- apply
- apply-dimension
- reduce_dimension
Maybe we could define an object that specifies chunking options, and include that as a parameter. (That way we have only have to document the details in one place, but can reuse it in multiple processes.)

m-mohr:
I see.

> * apply
> * apply-dimension
> * reduce_dimension
>   Maybe we could define an object that specifies chunking options, and include that as a parameter. (That way we have only have to document the details in one place, but can reuse it in multiple processes.)

They all have the context parameter, maybe back-ends can re-use that?

jdries:
Or maybe just go for a new apply_chunk process, with a definition similar to apply dimension?
Instead of working on a single dimension, it would work on a multidimensional subcube.

m-mohr:
From all mentioned alternatives, that seems to be the most reasonable approach.

jdries:
Reworked it, I actually went for apply_neighbourhood, maybe better than 'chunk'? (Although I'm fine either way.)
reduce_neighbourhood could indeed also work, although it would be fairly similar to the apply variant.

In general I do have some (backend specific?) issues with what we allow in terms of changing dimensions:

- spatial dimensions can not really be messed with, I don't really support reducing them or changing dimension labels (that would be resampling)
- For the bands dimension: these have discrete labels, a neighbourhood operation would always work on all bands, are perhaps we could do something with bands indices.

m-mohr:
@jdries, two organizational issues:
- This PR was made from master, not draft. I tried to change the branch to merge to from master to draft, but gives issues. Could you please either merge draft into the PR branch or redo the PR for draft?
- The build is failing. That also needs to be fixed, but I can also do that for you. Likely some spacing issues (no spaces before/after \n for example) and summary too long.


Regarding naming I'll check what would be most consistent with the other processes, but the parameter names are pretty long and relatively hard to type correctly, e.g. `neighbourhood_padding`. Would it be okay to shorten, e.g. to "size" and "padding"? Also, wouldn't "overlap" be more intuitive? I had to read the description to understand what's meant with padding, but likely would have directly known from "overlap". The schemas for size and padding are still to vague. It's not clear what the objects should be.

PR #163, opened 2020-06-29
#154 add apply_neighbourhood

jdries:
New pull request for apply_neighbourhood


m-mohr:
Old PR with some early discussions: #160

m-mohr:
8a8ae430905e148a003eaa4c1d70eccb0abfd5c6 fixes the build.

jdries:
I'm trying to define a schema for size & overlap per dimension, I was thinking of something like:
```
size:{
  'x': '256px',
  'y': '256px',
  't': '5d',
 'bands': 'all'
}
overlap:{
 'x':'32px',
 'y':'32px'
}
```
But not sure if that is possibe in json-schema? For temporal dimensions, we would again need some kind of 'duration' description.

This is an alternative that might work better with json-schema:
sizes:[
{ 'dimension_name':'x','value':256,'unit':'m'},
{ 'dimension_name':'y','value':256,'unit':'m'}
]

jdries:
> [8a8ae43](https://github.com/Open-EO/openeo-processes/commit/8a8ae430905e148a003eaa4c1d70eccb0abfd5c6) fixes the build.
> 
> We should make clear in the descriptions what the difference of apply_kernel and apply_neighbourhood is.

Maybe the definition of apply_kernel can refer to apply_neighbourhood? Like:
For cases requiring more generic focal operations, see apply_neighbourhood.

And in apply_neighbourhood:
For the special case of 2D convolution, it is recommended to use apply_kernel.

m-mohr:
Regarding schema for size & overlap: What I find most challenging is to find a way to properly communicate what kind of values and units the user can specify per dimension. That could be pretty different between back-ends and there can be even "other" dimensions (e.g. crs).

In `cube:dimensions` we have a step property, which for temporal uses ISO8601 durations (e.g. P1D). So that seems the most obvious choice.
From your example, spatial could be (k)m and px. No idea for bands and other. In general, we could say it must be in the unit of the reference system. That would avoid specifying a unit, but then there's likely no "px". For nominal values like bands, it seems most reasonable to use arrays...

https://github.com/Open-EO/openeo-processes/pull/163#issuecomment-652314405 is a good suggestion.

apply_neighbourhood.json
```diff
@@ -0,0 +1,74 @@
...
+            "schema": {
+                "type": "object",
+                "subtype": "process-graph",
+                "parameters": [
+                    {
+                        "name": "data",
+                        "description": "A labeled array with elements of any type.",
+                        "schema": {
+                            "type": "array",
+                            "subtype": "labeled-array",
```
m-mohr, line 29: subtype should be `raster-cube`

apply_neighbourhood.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+            "name": "size",
+            "description": "Neighbourhood sizes along each dimension. Default is to provide only one value.\n This object maps dimension names to either a physical measure, (e.g. 100 m, 10 days), or pixels (e.g. 32 pixels).\nTo include all values along a dimension, use `all`. This may result in an exception if the dimension is too large to be fully included.",
+            "schema": {
+                "type": "object"
+            }
+        },
+        {
+            "name": "overlap",
```
m-mohr, line 55: add: `optional: true` and a default value.

apply_neighbourhood.json
```diff
@@ -0,0 +1,74 @@
...
+            "description": "Overlap of chunks along each dimension. This allows neighbourhoods to overlap. For instance a temporal dimension can add 1 month before and after a neighbourhood. In the spatial dimensions, this is often a number of pixels. Modifying overlapping data in subsequent operations should have no effect.",
+            "schema": {
+                "type": "object"
+            }
+        }
+    ],
+    "returns": {
+        "description": "A data cube with the newly computed values. The resolution and the number of dimensions are the same as for the original data cube.",
```
soxofaan, line 63: > the resolution .. are the same as for the original data cube

Will this be the case in general? If you use a large `size` (and small `overlap`), the result will have a "downsampled" resolution if I understand correctly.

m-mohr: Maybe I misunderstood the operation, but I thought the resolution will not change.

m-mohr:
@jdries I made a first draft/proposal for a schema for the size and overlap parameters. I also resolved some of the other issues discussed yesterday, added a Changelog entry and added the references to apply_kernel etc.

apply_neighbourhood.json
jdries: No spatial resolution should remain unchanged. For bands and time dimensions, our backend could be more flexible, but not sure if that is a general thing. (People can always set all values to nodata to 'remove' something as well).
So it's more like apply_dimension, which can change timestamps in the temporal dimension if I'm correct.

jdries:
Thanks! I may also want to add an example, but that'll be for monday.
Based on this, I can also start implementing the process to see if that would work.

apply_neighbourhood.json
m-mohr: I guess than the description in the process parameter:
> The specified process needs to accept an array as parameter and must return an array with least one element. A process may consist of multiple sub-processes.

has to be removed

soxofaan: Ok I misunderstood the goal of this process apparently. 
So in terms of resolution, input/output cardinality it's a generalization of a convolution (`apply_kernel`)?

m-mohr: As far as I understood, yes.

soxofaan: but then I don't understand why you have both `size` and `overlap`

m-mohr: I understood it like that:
The overlap is just additional information you may use for computation, but once you "replace" the data in the original data with the newly computed data, the overlap is skipped.

soxofaan: As far as I understand focal operations (and convolutions), the window size and window overlap are tied to each other 1-on-1. 
For example in 1 dimension: if you process the value at index `x`, you take input values in the window `[x-n,x+n]` (so the window size is `2n+1`). For the next pixel at index `x+1`, the shifted window has an overlap of `2n` pixels  with the previous window, so overlap size = window size - 1 pixel.

If overlap is about how much margin should be taken when doing tile based processing, then still the overlap can be derived from the window size I would think (`n` in the 1D example).

m-mohr: @jdries will add a better description to the process, including the ML example.

m-mohr:
@jdries I fixed the example in 7cde70beb257a063022caaae4adf6053dac021fe. I also clarified the default values/behavior a bit more.

apply_neighbourhood.json
```diff
@@ -0,0 +1,217 @@
...
+                        },
+                        "optional": true,
+                        "default": null
+                    }
+                ]
+            }
+        },
+        {
+            "name": "size",
+            "description": "Neighbourhood sizes along each dimension.\nThis object maps dimension names to either a physical measure, (e.g. 100 m, 10 days), or pixels (e.g. 32 pixels).\nFor dimensions not specified, the default is to provide all values. This may result in an exception if the dimension is too large to be fully included.",
```
m-mohr, line 46: Do you think we need to say the following?

>  This may result in an exception if the dimension is too large to be fully included.

This is also true for other processes, but we never specify this explicitly. So I think we can remove it (or if it should stay, specify a concrete exception).

apply_neighbourhood.json
soxofaan: for completeness, to conclude this subthread: as discussed in telco this morning: `size` is not only (part of) the size of the input window, but also the size of the block of output pixels you write for a given position of the sliding window (e.g. with convolutions this is always 1 pixel, but `apply_neighbourhood` allows larger areas at once). The sliding window moves with multiples of `size`. And `overlap` is an additional extension of the sliding window to take more input pixels (output "block" size and step size are  still `size`).

apply_neighbourhood.json
```diff
@@ -43,7 +43,7 @@
         },
         {
             "name": "size",
-            "description": "Neighbourhood sizes along each dimension.\nThis object maps dimension names to either a physical measure, (e.g. 100 m, 10 days), or pixels (e.g. 32 pixels).\nFor dimensions not specified, the default is to provide all values. This may result in an exception if the dimension is too large to be fully included.",
+            "description": "Neighbourhood sizes along each dimension.\nThis object maps dimension names to either a physical measure, (e.g. 100 m, 10 days), or pixels (e.g. 32 pixels).\nFor dimensions not specified, the default is to provide all values. This may result in an exception if the dimension is too large to be fully included.\n\n The moving window moves with multiples of 'size' along each dimension. All values in the window specified by size can be modified by the process.\n\n The neighbourhood size should be kept small enough, to avoid running beyond computational resources, but a too small size will result in a larger number of process invocations, which may slow down processing.",
```
m-mohr, line 46: I don't feel this really helps a user:

> The neighbourhood size should be kept small enough, to avoid running beyond computational resources, but a too small size will result in a larger number of process invocations, which may slow down processing.

There are no number to refer to so it's really not so meaningful what actually is too large or too small. Thus, I'm not sure it should be included here.

apply_neighbourhood.json
```diff
@@ -90,7 +90,7 @@
...
-            "description": "Overlap of chunks along each dimension. This allows neighbourhoods to overlap. For instance a temporal dimension can add 1 month before and after a neighbourhood. In the spatial dimensions, this is often a number of pixels. The overlap specified is added before and after, so an overlap of 8 pixels will add 8 pixels on both sides of the window, so 16 in total.\nModifying overlapping data in subsequent operations should have no effect.\nFor dimensions not specified, the default is to provide no overlap.",
+            "description": "Overlap of neighbourhoods along each dimension. For instance a temporal dimension can add 1 month before and after a neighbourhood. In the spatial dimensions, this is often a number of pixels. The overlap specified is added before and after, so an overlap of 8 pixels will add 8 pixels on both sides of the window, so 16 in total.\nModifying overlapping data in subsequent operations should have no effect.\n\nOverlaps are necessary to avoid border effects, but large overlaps also increase the need for computational resources. This trade-off should be kept in mind when designing your process graph.",
```
m-mohr, line 93: We don't use process graph in user communication any longer. Should just be process.

apply_neighbourhood.json
m-mohr: Was adding something similar to this conclusion also part of your commit https://github.com/Open-EO/openeo-processes/pull/163/commits/9a2cc6cf2b1758ebcedcfb90df1048cbd644971a, @jdries ?

apply_neighbourhood.json
jdries: Indeed, there are no numbers, maybe I can add some for better guidance. In my experience, users do appreciate this kind of advice, as otherwise they might use entirely wrong settings. For backends, it also gives an idea of what is common.
Proposal:
Window sizes for spatial dimensions typically are in the range of 64px to 512px, while overlaps of 8px to 32px are common.

m-mohr: Sounds good.

apply_neighbourhood.json
jdries: I would prefer having some kind of recommendation to limit dimensions that don't have a clearly finite number of values, like the bands dimension. But not sure how to word it right.

m-mohr: @jdries Maybe something like: 

> Be aware that including all values from very large dimensions may not be processed at once.

Is a bit inspired from the *_binary processes.

apply_neighbourhood.json
m-mohr: @soxofaan I used some of your wording to improve the process description.

m-mohr:
I made several changes to the process and parameter descriptions in the last commit. They should solve all comments made above. @jdries

m-mohr:
Rendered version of the current state:

image

m-mohr:
Rendered version of the current state (pretty long):

image

m-mohr:
Two more questions: https://github.com/Open-EO/openeo-processes/issues/154#issuecomment-654710816

m-mohr:
@jdries Could you please check https://github.com/Open-EO/openeo-processes/pull/163/commits/95dd63bf07d2e38633811ab958aa3169d15ba087?

jdries:
Reviewed latest changes, looking good!

PR #166, opened 2020-07-01
Fixes #165: clarify kernel orientation

### LINKED ISSUE #165: kernel orientation in apply_kernel
https://github.com/Open-EO/openeo-processes/blob/9ee7b1ef45e3c968fed7d2f74bb6d7b1440d072e/apply_kernel.json#L20

I think the spec should specify explicitly how the levels of the nested array correspond with the x and y dimension.

I think in most languages (python/numpy, matlab, ...), the inner level  corresponds with the x dimension and outer level corresponds with the y dimension, because that is the most intuitive way to render it:

    array([[1, 2],
           [3, 4]])

But I don't think we should let this open for interpretation

soxofaan:


m-mohr:
To be reviewed on Monday :-)

apply_kernel.json
```diff
@@ -17,7 +17,7 @@
         },
         {
             "name": "kernel",
-            "description": "A two-dimensional weighted kernel. Each dimension of the kernel must have an uneven number of elements, otherwise the process throws a `KernelDimensionsUneven` error.",
+            "description": "Kernel as a two-dimensional array of weights. The inner level of the nested array corresponds to the `x` dimension and the outer level corresponds to the `y` dimension. Each dimension of the kernel must have an uneven number of elements, otherwise the process throws a `KernelDimensionsUneven` error.",
```
m-mohr, line 20: Referring to the x and y dimension feels a bit unclear. In the metadata it's called x and y axis. If we way dimension, it sounds a bit like the dimension needs to be names as such, but that's not the case although that's the suggested default.

```suggestion
            "description": "Kernel as a two-dimensional array of weights. The inner level of the nested array corresponds to the `x` axis and the outer level corresponds to the `y` axis. Each dimension of the kernel must have an uneven number of elements, otherwise the process throws a `KernelDimensionsUneven` error.",
```

soxofaan: what about 

> horizontal spatial dimension (commonly called `x`) and vertical spatial dimension (commonly called `y`)

?

m-mohr: Hm, y is not really vertical? x and y are both horiztonal. Throughout the API z is the vertical axis.

soxofaan: hmm indeed, I'm too stuck in my 2d world :)

is it common to speak of *first* and *second* spatial dimension? (so without using "horizontal" because that could introduce confusion)

m-mohr: Not sure. I guess the long form would be:

> The inner level of the nested array corresponds to the dimension with the axis set to `x` and the outer level corresponds to the dimension with the axis set to `y`.

That is at least very close to your metadata, where the horizontal dimensions have a property "axis" set to "x" or "y".

soxofaan: To not make it too confusing I went with your first suggestion 

> The inner level of the nested array aligns with the `x` axis ...

(force pushed amended commit)

soxofaan:
(also added changelog item)

Issue #168, opened 2020-07-02
mod operation


m-mohr:
Some things I'd like to improve in the mod process:

- Is it defined on integers only? If yes, change the schema. If no, add a non-integer example.
- Can a DivisionByZero error occur?

Related:
- https://github.com/Open-EO/openeo-processes-python/blob/master/src/openeo_processes/math.py#L1761 @claxn

PR #169, opened 2020-07-02
#168: Clarifications for mod (modulo)

m-mohr:
See #168


mod.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "mod",
     "summary": "Modulo",
-    "description": "Remainder after division of `x` by `y`.\n\nThe result of a modulo operation has the sign of the divisor. The handling regarding the sign of the result [differs between programming languages](https://en.wikipedia.org/wiki/Modulo_operation) and needs careful consideration while implementing this process.\n\nThe no-data value `null` is passed through and therefore gets propagated if any of the arguments is `null`.",
+    "description": "Remainder after division of `x` by `y` for both integers and floating-point numbers.\n\nThe result of a modulo operation has the sign of the divisor. The handling regarding the sign of the result [differs between programming languages](https://en.wikipedia.org/wiki/Modulo_operation#In_programming_languages) and needs careful consideration to avoid unexpected results. To remove the sign, use the process ``absolute()``.\n\nThe no-data value `null` is passed through and therefore gets propagated if any of the arguments is `null`. A modulo by zero results in ±infinity if the processing environment supports it. Otherwise a `DivisionByZero` error must the thrown.",
```
soxofaan, line 4: > To remove the sign, use the process ``absolute()``

apply `absolute()` on `y` or the result of `mod()`? I read the doc as applying it to the result, which is usually not the correct approach.

m-mohr: Well, I guess I just remove that part and leave it up to the user... ;-)

Issue #170, opened 2020-07-03
apply_kernel - what to do with the borders pixels?


clausmichele:
In the API it's specified that the returned array/datacube has the same x and y dimension, but it doesn't specify what to do with pixels on the border. Possibilities:

1. Keep border pixels untouched
2. Pad the data with zeros before applying convolution
3. Replicate border pixels (padding with same values) before applying convolution

There may be other solutions, let me know what do you think.

m-mohr:
What is common ground in other software or research papers?

clausmichele:
Have a look here
https://docs.opencv.org/4.3.0/d2/de8/group__core__array.html#ga209f2f4869e304c82d07739337eae7c5
or here
http://image.dask.org/en/latest/dask_image.ndfilters.html

m-mohr:
It seems like an additional parameter would make sense?!

aljacob:
I agree in an ideal case that should be user defined behavior, with maybe a documented default handling in the back-end implementation e.g. repeating last known value or fill with zeros

m-mohr:
@claxn @jdries @soxofaan What's the default behavior you've implemented?

soxofaan:
I'm not that familiar with current implementation (so for @jdries to confirm)  but as far as I understand:  we pad with NaN or 0 (which are practically equivalent here) before calculating the convolution.

m-mohr:
Make PR with either 0 or reflect as default or set values that don't have a full window to null.

Also: Clarify how nulls are handled. Users may need to pre-process, but we should mention that in the docs.

Discuss again in Dev Telco.

claxn:
@m-mohr : I haven't implemented this process yet, but when applying a kernel I would use convolution, where the following boundary cases are common (`fill`, `wrap`, and `symm`): [https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve2d.html](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve2d.html)

m-mohr:
I'm now comparing the tools mentioned above (scypi, dask, opencv + R EBImage filter2; astropy seems to use scypi?) regarding possible border types and their defaults:

1. user-defined constant value `iiiiii\|abcdefgh\|iiiiiii` (scypi: fill 0, dask: constant, opencv: constant, r: *pass number*) -> openEO pass numeric value
[...]

Issue #173, opened 2020-07-07
interpolation process


m-mohr:
Now that I'm working on apply_kernel and the null/NA behavior, I'm thinking what to suggest to users to do before. They can certainly do a apply_neighboorhood before or replace nulls with apply/if or mask. On the other hand, it seems not unusual to use an interpolation function and so I'm wondering whether it could be useful to add an interpolate process. It could also allow to take the time dimension into account (idea from @claxn).

m-mohr:
PR proposal for linear interpolation for arrays: https://github.com/Open-EO/openeo-processes/pull/232

PR #174, opened 2020-07-07
#170: apply_kernel - border parameter and nodata handling

m-mohr:
see #170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - `apply_kernel`:
     - Clarify orientation of the 2D kernel array- [#165](https://github.com/Open-EO/openeo-processes/issues/165)
-    - Clarify no-datta handling. [#170](https://github.com/Open-EO/openeo-processes/issues/170)
+    - Clarify no-data handling. [#170](https://github.com/Open-EO/openeo-processes/issues/170)
 - `load_collection`: Removed outdated `require` property from `value` callback parameter in process parameter `properties`.
 - `filter_bbox`, `load_collection`, `resample_spatial`: Fixed invalid EPSG code examples.
 - `aggregate_temporal`: Fixed outdated message for exception `TooManyDimensions`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Process graph examples from `arccos`, `arcsin`, `arctan`, `arsinh`, `artanh`, `e`, `ln` and `pi`. [API#285](https://github.com/Open-EO/openeo-api/issues/285)
 
 ### Fixed
-- `apply_kernel`: Clarify orientation of the 2D kernel array [#165](https://github.com/Open-EO/openeo-processes/issues/165)
+- `apply_kernel`:
+    - Clarify orientation of the 2D kernel array- [#165](https://github.com/Open-EO/openeo-processes/issues/165)
+    - Clarify no-datta handling. [#170](https://github.com/Open-EO/openeo-processes/issues/170)
 - `load_collection`: Removed outdated `require` property from `value` callback parameter in process parameter `properties`.
 - `filter_bbox`, `load_collection`, `resample_spatial`: Fixed invalid EPSG code examples.
 - `aggregate_temporal`: Fixed outdated message for exception `TooManyDimensions`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Process graphs added to:
     - `mean`
     - `median`
+- `apply_kernel`: Added parameter `border` [#170](https://github.com/Open-EO/openeo-processes/issues/170)
 - Folder with examples (`examples/`). [#136](https://github.com/Open-EO/openeo-processes/issues/136)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Process graphs added to:
     - `mean`
     - `median`
-- `apply_kernel`: Added parameter `border` [#170](https://github.com/Open-EO/openeo-processes/issues/170)
+- `apply_kernel`: Added parameters `border` and `replace_nodata` [#170](https://github.com/Open-EO/openeo-processes/issues/170)
 - Folder with examples (`examples/`). [#136](https://github.com/Open-EO/openeo-processes/issues/136)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Process graphs added to:
     - `mean`
     - `median`
-- `apply_kernel`: Added parameters `border` and `replace_nodata` [#170](https://github.com/Open-EO/openeo-processes/issues/170)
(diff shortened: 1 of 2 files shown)


apply_kernel.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "apply_kernel",
     "summary": "Apply a spatial convolution with a kernel",
-    "description": "Applies a 2D convolution (i.e. a focal operation with a weighted kernel) on the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nEach value in the kernel is multiplied with the corresponding pixel value and all products are summed up afterwards. The sum is then multiplied with the factor.",
+    "description": "Applies a 2D convolution (i.e. a focal operation with a weighted kernel) on the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nEach value in the kernel is multiplied with the corresponding pixel value and all products are summed up afterwards. The sum is then multiplied with the factor. The process can't handle no-data values (`null`). The data cube needs to be pre-processed with ``mask()`` or other processes to replace such values as otherwise a `NoDataException` is thrown.",
```
soxofaan, line 4: Maybe put the part of no-data handling in its own paragraph?

Also: the pre-processing to replace the nans is not that trivial I think (you need at least 3 additional processes: `apply` with `isnan` as subprocess and `mask`). And if you specify that `NoDataException` can thrown, any user will probably want to add this preprocessing to be sure. Wouldn't it be easier (and more efficient on backend side) to specify that `apply_kernel` will automatically replace nan with zero and if the user wants different behaviour, they have to do their own pre-processing explicitly?

Also(2) : what about no-data in the kernel? Here it would be fine to just throw `NoDataException` I think.

apply_kernel.json
```diff
@@ -38,6 +38,26 @@
             },
             "default": 1,
             "optional": true
+        },
+        {
+            "name": "border",
+            "description": "Determines how the data is extended when the filter overlaps the borders. Defaults to fill the border with zeroes.\n\nThe following options are available:\n\n* *numeric value* - fill with a user-defined constant number `n`: `nnnnnn|abcdefgh|nnnnnn` (default, with `n` = 0)\n* `replicate` - repeat the value from the pixel at the border: `aaaaaa|abcdefgh|hhhhhh`\n* `reflect` - mirror/reflect from the border: `fedcba|abcdefgh|hgfedc`\n* `reflect_pixel` - mirror/reflect from the center of the pixel at the border: `gfedcb|abcdefgh|gfedcb`\n* `wrap` - repeat/wrap the image: `cdefgh|abcdefgh|abcdef`",
```
soxofaan, line 44: > when the filter overlaps the borders

instead of "filter", use "kernel" because that is used in other places?

Maybe also add an option `auto` (which could even be the default): let back-end pick the approach that is most efficient  (e.g. zero-filling for direct calculation, wrap for FFT based calculations)

apply_kernel.json
m-mohr: > Maybe put the part of no-data handling in its own paragraph?

I thought about that, too, and was undecided. So I'll put a line-break in-between.

> Also: the pre-processing to replace the nans is not that trivial I think (you need at least 3 additional processes: `apply` with `isnan` as subprocess and `mask`).

I thought about that, too. ;-) @claxn argued that there's likely a mask anyway so I just put that there. But we can also expand (or remove) it. For now, I just added a larger list of potentially useful processes. Do you have better suggestions?

> And if you specify that `NoDataException` can thrown, any user will probably want to add this preprocessing to be sure. Wouldn't it be easier (and more efficient on backend side) to specify that `apply_kernel` will automatically replace nan with zero and if the user wants different behaviour, they have to do their own pre-processing explicitly?

With @claxn, we thought to add another parameter that allows to set a certain numeric value (e.g. 0 by default) that replaces NA and additionally thought about adding interpolate (see #173). Conclusion was that we don't want to overdo it with this process as then it does quite a lot at a time. Also, we thought it might in many cases not be so useful to just replace with a number and you need a more advanced pre-process step (like interpolate) anyway. What do you think?

> Also(2) : what about no-data in the kernel? Here it would be fine to just throw `NoDataException` I think.

Yes, added that.

apply_kernel.json
m-mohr: > instead of "filter", use "kernel" because that is used in other places?

Yes, changed!

> Maybe also add an option `auto` (which could even be the default): let back-end pick the approach that is most efficient (e.g. zero-filling for direct calculation, wrap for FFT based calculations)

If really necessary, back-ends can still change the default value in the process description although that decreases interoperability. So `auto` is not really needed, but both options are not recommended anyway.

apply_kernel.json
soxofaan: as a user I would prefer automatic replacing NaN with 0 by `apply_kernel` itself instead of having to think about preparing my data everytime I want to run `apply_kernel`. But maybe other user viewpoints prefer exceptions above automatic/blind replacements.

another solution is introducing a new process like `replace_nan` to replace NaNs with a desired value. 
That would simplify the overhead users have to take care of before running `apply_kernel`

m-mohr: I just realized the kernel can't contain a nodata value. Thus, I removed the exception from the code again.

m-mohr: And added a replace_nodata parameter now. Can you check it, please?

m-mohr:
@clausmichele @claxn @aljacob @lforesta Latest draft is available. What do you all think?

CHANGELOG.md
```diff
@@ -28,7 +29,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Process graph examples from `arccos`, `arcsin`, `arctan`, `arsinh`, `artanh`, `e`, `ln` and `pi`. [API#285](https://github.com/Open-EO/openeo-api/issues/285)
 
 ### Fixed
-- `apply_kernel`: Clarify orientation of the 2D kernel array [#165](https://github.com/Open-EO/openeo-processes/issues/165)
+- `apply_kernel`:
+    - Clarify orientation of the 2D kernel array- [#165](https://github.com/Open-EO/openeo-processes/issues/165)
+    - Clarify no-datta handling. [#170](https://github.com/Open-EO/openeo-processes/issues/170)
```
lforesta, line 34: typo: no-datta

apply_kernel.json
```diff
@@ -38,6 +38,35 @@
             },
             "default": 1,
             "optional": true
+        },
+        {
+            "name": "border",
+            "description": "Determines how the data is extended when the kernel overlaps the borders. Defaults to fill the border with zeroes.\n\nThe following options are available:\n\n* *numeric value* - fill with a user-defined constant number `n`: `nnnnnn|abcdefgh|nnnnnn` (default, with `n` = 0)\n* `replicate` - repeat the value from the pixel at the border: `aaaaaa|abcdefgh|hhhhhh`\n* `reflect` - mirror/reflect from the border: `fedcba|abcdefgh|hgfedc`\n* `reflect_pixel` - mirror/reflect from the center of the pixel at the border: `gfedcb|abcdefgh|gfedcb`\n* `wrap` - repeat/wrap the image: `cdefgh|abcdefgh|abcdef`",
```
lforesta, line 44: "overlaps with the borders."

CHANGELOG.md
m-mohr:
```suggestion
    - Clarify no-data handling. [#170](https://github.com/Open-EO/openeo-processes/issues/170)
```

apply_kernel.json
m-mohr:
```suggestion
            "description": "Determines how the data is extended when the kernel overlaps with the borders. Defaults to fill the border with zeroes.\n\nThe following options are available:\n\n* *numeric value* - fill with a user-defined constant number `n`: `nnnnnn|abcdefgh|nnnnnn` (default, with `n` = 0)\n* `replicate` - repeat the value from the pixel at the border: `aaaaaa|abcdefgh|hhhhhh`\n* `reflect` - mirror/reflect from the border: `fedcba|abcdefgh|hgfedc`\n* `reflect_pixel` - mirror/reflect from the center of the pixel at the border: `gfedcb|abcdefgh|gfedcb`\n* `wrap` - repeat/wrap the image: `cdefgh|abcdefgh|abcdef`",
```

Issue #177, opened 2020-07-13
format errors for save_result


soxofaan:
from https://github.com/Open-EO/openeo-api/blob/draft/CHANGELOG.md:

> Removed all error messages with tag Processes (CRSInvalid, CoordinateOutOfBounds) or related to storing file formats (FormatUnsupported, FormatArgumentUnsupported, FormatArgumentInvalid, FormatUnsuitable) as they are usually defined directly in the process specification as exceptions.

should these format errors be listed in e.g. `save_result` ?

m-mohr:
That would probably be good, although they also need to be added to load_uploaded files [i.e. all parameters with subtype output-format(-options) or input-format(-options)]

PR #178, opened 2020-07-15
Issue #177: add exceptions to save_result and load_uploaded_files

soxofaan:
see #177 

Based a bit on the original (removed) errors from openeo-api, including placeholders, I'm however not sure this is allowed as there seem to be no other exception messages with placeholders in openeo-processes

### DIFF
--- a/load_uploaded_files.json
+++ b/load_uploaded_files.json
@@ -44,5 +44,19 @@
             "type": "object",
             "subtype": "raster-cube"
         }
+    },
+    "exceptions": {
+        "FormatUnsupported": {
+            "message": "Input format is not supported."
+        },
+        "FormatOptionRequired": {
+            "message": "Format option '{option}' is required."
+        },
+        "FormatOptionInvalid": {
+            "message": "The value passed for '{option}' is invalid: {reason}."
+        },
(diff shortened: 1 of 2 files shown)


m-mohr:
Thanks for doing this. Didn't had the time myself yet.

Indeed, placeholders are not specified in the process spec and thus it's not ideal to use them.
errors.json in the API itself doesn't follow any spec and was originally just copy&pasted from the GEE driver, that's the reason it is as it is.

soxofaan:
> maybe it's better to use the ProcessParameterInvalid error

that would avoid having to specify too much different exceptions indeed. Possible problem is that the format options are sub-items in a single `options` parameter, which makes it less straightforward to point to the desired option

m-mohr:
Yes, indeed. But can't think of a better (and easy) solution yet...

soxofaan:
hmm strange, I pushed an update commit to my `soxofaan:issue-177-save-result-exceptions` branch, but github doesn't seem to pick it up here

soxofaan:
github seems to be having degraded performance issues at the moment

m-mohr:
Thanks, closing in favor of #180

Issue #184, opened 2020-07-31
Clarify x and y in merge_cubes


m-mohr:
I would like to understand if x and y of the overlap resolver refer to cube1 and cube2 (x -> cube1, y -> cube2), since it is not explicitly stated in the API. This doubt comes from [some inconsistencies](https://github.com/Open-EO/openeo-earthengine-driver/issues/47) I got from testing with the GEE backend, which I'm not sure are due to the implementation or the definition of the process itself.

_Originally posted by @clausmichele in https://github.com/Open-EO/openeo-processes/issues/87#issuecomment-667095522_

m-mohr:
Yes, x refers to a value from cube1 and y refers to a value from cube2. We should clarify that in the processes.

claxn:
@m-mohr, @clausmichele : Thanks for the update. I have fixed it in the Python parser with the rule: 

- If the number of sub-parameters ('x', 'y', ...) is the same number as the input nodes in the parent process, then assume that, e.g.,  'x' refers to the first input node and 'y' to the second input node (`overlap_resolver` example: `{'process_id': 'subtract', 'arguments': {'x': {'from_node': '3_2'}, 'y': {'from_node': '4_4'}, 'result': True}`
- If the number of sub-parameters ('x', 'y', ...) is not the same number as the input nodes in the parent process, then assume that each input node is split up into 'x', 'y', ... by the back-end (`overlap_resolver` example: `{'process_id': 'subtract', 'arguments': {'x': [{'from_node': '3_2'}, {'from_node': '4_4'}], 'y': [{'from_node': '3_2'}, {'from_node': '4_4'}]}, 'result': True}`).

Would this be the intended behaviour? Or how can one figure out the correct `from_parameter` relation in this case?

m-mohr:
I don't understand this explanation @claxn. You just need from_parameter for x and y?!

m-mohr:
The corresponding PR is #186.

claxn:
[...]

PR #186, opened 2020-08-04
Clarify overlap resolver #184

m-mohr:
Clarified overlap resolver as discussed in #184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased / Draft
 
+### Fixed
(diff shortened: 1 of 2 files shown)


clausmichele:
So, if we want to merge two datacubes and perform some operation on their data they must have the same band name or label to be considered overlapping, right? The code should first check if there are corresponding bands/labels and act accordingly.

Did I understand it correctly?

m-mohr:
> So, if we want to merge two datacubes and perform some operation on their data they must have the same band name or label to be considered overlapping, right?

If "their data" is "their overlapping data", then yes.


> The code should first check if there are corresponding bands/labels and act accordingly.
If "code" means "back-end code", then yes.

clausmichele:
Alright, now it's clearer.

Issue #189, opened 2020-09-15
is_nodata([null, null])?


soxofaan:
What is the behavior of `is_nodata` given an array of only nan/null values?

    is_nodata([null, null])

From current docs I guess the return value should be false, but it's not that clear I think

m-mohr:
Yes, false.

It's indeed not very clear, but there's a reference to equals and `equals([null, null], null)` would also return false. I will make a PR for it.

soxofaan:
thanks

> but there's a reference to equals

I didn't read the "equals" as a reference to the `eq` process (which indeed has in-depth rules), just as a "handwavy" verb.
Also: `eq` does not specify rules for NaN

m-mohr:
It has a rule for it (null = nodata; NaN is not covered by is_nodata, see is_nan): 

> If any operand is null, the return value is null. Therefore, eq(null, null) returns null instead of true.

Indeed, I implied here that null will be casted to false.

soxofaan:
I'm not sure If I should piggy back this issue or create a new one: but what about

    is_nan(null)

description of  `is_nan` says 

> All non-numeric data types MUST also return true.

as `null` is non-numeric, `is_nan(null)` should return True?

soxofaan:
An while we're at it, let's also look at `is_valid`

Attempt to make a truth table of my current understanding of the docs

x | is_nan(x) | is_nodata(x) | is_valid(x)
----|----|----|----
null | True?  | True  | False
NaN | True | True | False
1 (finite int/float)  | False | False | True
"a string" | True | False | True
[1, 2]   | True | False | True
[null, null]) | True | False | True ?
[NaN, NaN] | True | False | True ?

is that correct?

m-mohr:
You had it mostly correct:

| x                    | is_nan(x) | is_nodata(x) | is_valid(x) |
| -------------------- | --------- | ------------ | ----------- |
| null                 | True      | True         | False       |
| NaN                  | True      | False        | False       |
| 1 (finite int/float) | False     | False*       | True        |
| infinity             | False     | False        | False       |
| "a string"           | True      | False*       | True        |
| [1, 2]               | True      | False        | True        |
| [null, null])        | True      | False        | True        |
| [NaN, NaN]           | True      | False        | True        |

* True for values that have been declared no data in the metadata; STAC Extension tbd.

soxofaan:
> True for values that have been declared no data in the metadata

[...]

PR #190, opened 2020-09-15
Fix issue 189

m-mohr:
Fixes issue #189 and adds new processes. Let me know what you think.

### DIFF
--- a/is_infinite.json
+++ b/is_infinite.json
@@ -1,7 +1,7 @@
 {
-    "id": "is_nodata",
-    "summary": "Value is not a no-data value",
-    "description": "Checks whether the specified data is a missing data, i.e. equals to `null` or any of the no-data values specified in the metadata.",
+    "id": "is_infinite",
+    "summary": "Value is an infinite number",
+    "description": "Checks whether the specified value `x` is an infinite number. The definition of infinite numbers follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935).",
     "categories": [
         "comparison"
     ],
@@ -15,7 +15,7 @@
         }
     ],
     "returns": {
-        "description": "`true` if the data is a no-data value, otherwise `false`.",
+        "description": "`true` if the data is an infinite number, otherwise `false`.",
         "schema": {
             "type": "boolean"
         }
@@ -37,13 +37,14 @@
             "arguments": {
                 "x": null
             },
-            "returns": true
-        },
-        {
-            "arguments": {
-                "x": [null, null]
-            },
             "returns": false
+        }
+    ],
+    "links": [
+        {
+            "rel": "about",
+            "href": "https://ieeexplore.ieee.org/document/4610935",
+            "title": "IEEE Standard 754-2008 for Floating-Point Arithmetic"
         }
     ]
 }
--- a/is_infinite.json
+++ b/is_infinite.json
@@ -1,7 +1,7 @@
 {
     "id": "is_infinite",
     "summary": "Value is an infinite number",
-    "description": "Checks whether the specified value `x` is an infinite number. The definition of infinite numbers follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935).",
+    "description": "Checks whether the specified value `x` is an infinite number. The definition of infinite numbers follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). The special numerical value `NaN` (not a number) as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935) is not an infinite number and must return `false`.",
     "categories": [
         "comparison"
     ],
--- a/is_nan.json
+++ b/is_nan.json
@@ -1,7 +1,7 @@
 {
     "id": "is_nan",
     "summary": "Value is not a number",
-    "description": "Checks whether the specified value `x` is not a number (often abbreviated as `NaN`). The definition of `NaN` follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",
+    "description": "Checks whether the specified value `x` is not a number. Returns `true` for numeric values (integers and floating point numbers), except for the special value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",
     "categories": [
         "comparison",
         "math > constants"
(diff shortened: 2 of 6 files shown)


is_nan.json
```diff
@@ -1,9 +1,10 @@
 {
     "id": "is_nan",
     "summary": "Value is not a number",
-    "description": "Checks whether the specified value `x` is not a number (often abbreviated as `NaN`). The definition of `NaN` follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`.",
+    "description": "Checks whether the specified value `x` is not a number (often abbreviated as `NaN`). The definition of `NaN` follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",
     "categories": [
```
soxofaan, line 5: > is not a number (often abbreviated as `NaN`)

this is a bit misleading because there is a difference between "not a number" (object is not an int or float) and "NaN" (which is a special IEEE float value).

soxofaan: proposal

> Checks whether the specified value `x` is not a number. Returns `True` for numeric values (integers and floats), except for the special value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.

is_nodata.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "is_nodata",
     "summary": "Value is not a no-data value",
-    "description": "Checks whether the specified data is a missing data, i.e. equals to any of the no-data values / `null`.",
```
soxofaan, line 4: add that "NaN" is not considered no-data (so return false)?

nan.json
```diff
@@ -0,0 +1,27 @@
+{
+    "id": "nan",
```
soxofaan, line 2: also add "infinity" process as you defined "is_infinity"?

is_valid.json
```diff
@@ -38,6 +38,12 @@
                 "x": null
             },
             "returns": false
+        },
```
soxofaan, line 41: comment about the process description above (I can only comment on diffs, not on unchanged lines):

The numbered list layout with embedded if in the second item is bit hard to understand.

proposal, something like:

>  Checks whether the specified value x is valid, which is defined based on the type of the value as follows:
> * for a numeric value: return true if value is a finite number and false otherwise. The definition ...
> * for no-data value (`null`): return false
> * any other value: return true (even for arrays only containing `null`, empty strings, string `"null"`, ...)

is_infinite.json
```diff
@@ -0,0 +1,50 @@
...
+            },
+            "returns": false
+        },
+        {
+            "arguments": {
+                "x": null
+            },
+            "returns": false
+        }
+    ],
```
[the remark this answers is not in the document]
m-mohr: No, unfortunately no infinite / NaN support in JSON.

is_nan.json
```diff
@@ -1,9 +1,10 @@
 {
     "id": "is_nan",
     "summary": "Value is not a number",
-    "description": "Checks whether the specified value `x` is not a number (often abbreviated as `NaN`). The definition of `NaN` follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`.",
+    "description": "Checks whether the specified value `x` is not a number (often abbreviated as `NaN`). The definition of `NaN` follows the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",
```
m-mohr, line 4:
```suggestion
    "description": "Checks whether the specified value `x` is not a number. Returns `True` for numeric values (integers and floats), except for the special value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",
```

is_nan.json
m-mohr: Sounds good.

is_nodata.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Checks whether the specified data is a missing data, i.e. equals to `null` or any of the no-data values specified in the metadata.",
```
m-mohr, line 4:
```suggestion
    "description": "Checks whether the specified data is a missing data, i.e. equals to `null` or any of the no-data values specified in the metadata. The special numerical value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935) is not considered no-data and must return `false`.",
```

nan.json
m-mohr: I could think of use cases where you want to set values to NaN as a user, but couldn't find use cases for infinity. I feel like infinity's (there's two of them) are only really the result of computations.

is_nodata.json
m-mohr: Done.

is_valid.json
m-mohr: I agree that it is not easy to understand. I wasn't 100% happy with the proposal though and have submitted another proposal. Is is better now?

nan.json
soxofaan: well, I even wonder if infinity will happen at all in general openeo usage. Things like division by zero probably cause some kind of ZeroDivision error instead of assigning infinity. Also: most users will probably be fine with receiving NaNs where infinity would theoretically be more correct.

m-mohr: Could be. We reference infinity values in divide, ln, log and mod as possible results in supporting environments.

is_infinite.json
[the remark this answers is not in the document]
soxofaan: something like this is invalid openeo syntax, right?

      "x": {"process_id": "nan", "arguments": {}}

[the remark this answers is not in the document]
m-mohr: Yes, we'd need to use from_node... Would be useful for examples, but is invalid.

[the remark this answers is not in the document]
m-mohr: If we can't provide a positive examples, maybe remove the examples completely?

is_valid.json
soxofaan: yes that's easier to parse

minor suggestion:

> All arrays, objects and strings are valid, regardless of their content.

soxofaan (approved):

is_infinite.json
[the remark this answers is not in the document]
soxofaan: yes it looks a bit silly like that

m-mohr:
Woops, the approval notification came earlier via e-mail then the comments...

Issue #187, opened 2020-08-21
Which process for mean over a window with no overlap


clausmichele:
I would like to take the mean over a window of a specified size. However, apply_kernel doesn't have a stride value to set, so it's not possible to have no overlap theoretically. I would like to know how you would handle this using the given processes and maybe if you can give an example.

Input:
[[1, 1, 1, 1, 1, 1],
           [2, 2, 2, 2, 2, 2],
           [3, 3, 3, 3, 3, 3]]

Output mean over a 3x3 window with no overlap (stride=3) : [[ 2, 2 ]]

m-mohr:
apply* processes don't change the number of dimension labels / resolution. In openEO terms this is an aggregation. What you want is basically aggregate_spatial, but there you can only specify geometries. Another process that might somewhat work is resample(_cube)_spatial with the "average" method. (Interesting, there it's called average while the process is called mean.)
I doubt we have a really good way to process what you want. I guess for your use case we'd need a new aggregate_spatial_stride or so.

Disclaimer: I've never heard about strides before, so I had to read into this, but may have understood it incorrectly.

clausmichele:
I was thinking about a process which could be named aggregate_spatial_window, where x and y dimensions will be different in output and the process will be applied over a window with dimensions specified by the user.
For example, input 9x9, window size 3x3, output 3x3
Or input 10x20, window 5x10, output 2x2
It would also need a parameter similar to apply_kernel for the border values, which specifies what to do when the window size is not multiple of the data size.

jdries:
I would definitely go for resample_spatial  with the "average' method. Only downside seems to be that the user needs to compute the correct target resolution himself, but isn't that possible?

clausmichele:
As I see it, I would still need to compute the window size from the data dimensions, which is something I shouldn't care as a user. I want to apply an operation on a given window size whatever I have as input, in my opinion the back-end should care about the rest.
Anyway, let's wait for the opinion of @m-mohr

m-mohr:
On the one hand it feels like resample_cube_spatial can handle that, but it's not convenient at all and convenience is the aim of the processes. On the other hand, we would then have quite a lot of processes that do something with windows (apply_kernel, apply_neighborhood, aggregate_spatial_window, ... maybe more in the future?) and that bothers me a bit. So I have no strong preference here. The best way to propose new processes is to just create a PR with a JSON file to discuss.

jdries:
[...]

PR #191, opened 2020-09-16
Create aggregate_spatial_window

clausmichele:
I created this pull request to discuss about the new process I would like to add as discussed here https://github.com/Open-EO/openeo-processes/issues/187


aggregate_spatial_window.json
```diff
@@ -0,0 +1,96 @@
...
+                "type": [
+                    "number",
+                    "null"
+                ]
+            }
+        },
+        {
+            "name": "boundary",
+            "description": "Operation to apply if X/Y are not multiple of xDim/yDim. Options are:\n\n- trim: trim the datacube to fit window dimensions. \n\n- pad: pad the datacube with NaNs to fit window dimensions.",
+            "schema": [
```
m-mohr, line 41: I guess you want to do the following here:

```
"schema": {
  "type": "string",
  "enum": [
    "pad",
    "trim"
  ]
}
```

aggregate_spatial_window.json
```diff
@@ -0,0 +1,96 @@
+{
+    "id": "aggregate_spatial_window",
+    "summary": "Zonal statistics for rectangular windows",
+    "description": "Aggregates statistics over the spatial dimensions. This process passes a list of values to the reducer. \n\n The results are computed dividing the input X/Y grid into xDim/yDim non-overlapping windows and applying the reducer process to each window.\n If X and Y are not multiple of xDim/yDim NaNs padding before the computation is applied.",
```
m-mohr, line 4: You need to remove all spaces around \n, otherwise the linter fails.

aggregate_spatial_window.json
```diff
@@ -0,0 +1,96 @@
...
+        {
+            "name": "data",
+            "description": "A raster data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "xDim",
```
m-mohr, line 19: Parameter names in openEO are snake_case. Could we come up with a better name and description? For me it's not clear what this parameter excepts exactly. (Also applies for yDim.)

What's the behavior for null?

aggregate_spatial_window.json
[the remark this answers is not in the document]
m-mohr: I guess this parameter could also be optional with a sensible default. Therefore, the parameter should be after all required parameters.

m-mohr (commented):
Some first things to work on.

aggregate_spatial_window.json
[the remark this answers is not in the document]
clausmichele: Ok modified as you suggested!

aggregate_spatial_window.json
clausmichele: Spaces removed.

aggregate_spatial_window.json
clausmichele: Modified variable names to snake_case. x_size would be ok? It specifies the size of the window along the X dimension (or longitude). I also added the default to 1, making both x_size and y_size optional. So, the default window dimension would be 1x1 and it should produce the same result as apply.

m-mohr:
Thanks for the changes. I tried to parse it, but the JSON is invalid. Could you make it valid JSON and run it through https://jsonformatter.org/ so that it's better indented and easier understandable? Must be made before merging anyway.

aggregate_spatial_window.json
m-mohr: Let's check whether we can align with other processes regarding naming. 
Is this process only meant to work on horizontal spatial dimensions (x,y)?
Is 1x1 the most suitable default? It seems that's actually the most uncommon case as we have apply...

clausmichele: Yes, it is meant to work on spatial dimensions (x,y).
I don't see a general default dimension, it should be the user to set it...we at Eurac would need it to apply a boxcar filtering operation on Sentinel-1 data, which could have dimensions 4x19 or 1x5 for example-

m-mohr:
Closing in favor of #192

PR #192, opened 2020-09-17
Add aggregate_spatial_window

clausmichele:
I created a new pull request from draft to draft, with the corrected json file, which now should be correctly formatted. Let's continue the discussion here.

### DIFF
--- a/aggregate_spatial_window.json
+++ b/aggregate_spatial_window.json
@@ -46,7 +46,7 @@
         },
         {
             "name": "size",
-            "description": "Window sizes (pixels) along each dimension.\n\nFirst value correspond to Window size along X or Longitude and second to Window size along Y or Latitude.",
+            "description": "Window sizes in pixels along the horizontal spatial dimensions.\n\nThe first value corresponds to the `x` axis, the second values corresponds to the `y` axis.",
             "schema": {
                 "type": "array",
                 "items": {
--- a/aggregate_spatial_window.json
+++ b/aggregate_spatial_window.json
@@ -1,6 +1,7 @@
 {
   "id": "aggregate_spatial_window",
   "summary": "Zonal statistics for rectangular windows",
+  "description": "Aggregates statistics over the spatial dimensions. This process passes a list of values to the reducer.\n\nThe results are computed dividing the input X/Y grid into x_size/y_size non-overlapping windows and applying the reducer process to each window.\nIf X and Y are not multiple of x_size/y_size NaNs padding before the computation is applied.",
   "categories": [
     "cubes",
     "aggregate & resample"
@@ -16,6 +17,7 @@
     },
     {
       "name": "reducer",
+      "description": "A reducer to be applied on all values of each geometry. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes.",
       "schema": {
         "type": "object",
         "subtype": "process-graph",
@@ -42,7 +44,6 @@
         ]
       }
     },
-<<<<<<< HEAD
 	{
 		"name": "size",
 		"description": "Window sizes (pixels) along each dimension.\n\nFirst value correspond to Window size along X or Longitude and second to Window size along Y or Latitude.",
@@ -65,28 +66,6 @@
 			}
 		}
 	},
-=======
-    {
-      "name": "x_size",
-      "description": "Window size along X or Longitude. The number of values used as input to the reducer along X dimension.",
-      "schema": {
-        "type": "integer",
-        "minimum": 1
-      },
-      "optional": true,
-      "default": 1
-    },
-    {
-      "name": "y_size",
-      "description": "Window size along Y or Latitude. The number of values used as input to the reducer along Y dimension.",
-      "schema": {
-        "type": "integer",
-        "minimum": 1
-      },
-      "optional": true,
-      "default": 1
-    },
->>>>>>> e2fa4060c12f133096b332df2d10f37e0503b365
     {
       "name": "boundary",
       "description": "Operation to apply if X/Y are not multiple of x_size/y_size. Options are:\n\n- pad (default): pad the datacube with NaNs to fit window dimensions.\n\n- trim: trim the datacube to fit window dimensions.",
@@ -114,5 +93,7 @@
     "description": "A data cube with the newly computed values and the same dimensions.\n\nThe resolution will change, depending on the chosen parameters. It decreases for the dimensions which have the corresponding parameter `size` = [`x_size`,`y_size`] set to values greater than 1.\n\nThe dimension labels will be set to the center of the window. The other dimension properties (name, type and reference system) remain unchanged.",
     "schema": {
       "type": "object",
+      "subtype": "vector-cube"
     }
   }
+}
--- a/aggregate_spatial_window.json
+++ b/aggregate_spatial_window.json
@@ -67,7 +67,7 @@
         },
         {
             "name": "boundary",
-            "description": "Behavior to apply if the number of values for the axes `x` and `y` is not a multiple of the corresponding `size` = [`x_size`,`y_size`] parameter. Options are:\n\n- `pad` (default): pad the data cube with `NaN` (not a number) values to fit the required window size.\n\n- `trim`: trim the data cube to fit the required window size.",
+            "description": "Behavior to apply if the number of values for the axes `x` and `y` is not a multiple of the corresponding value in the `size` parameter. Options are:\n\n- `pad` (default): pad the data cube with `NaN` (not a number) values to fit the required window size.\n\n- `trim`: trim the data cube to fit the required window size.",
             "schema": {
                 "type": "string",
                 "enum": [
--- a/aggregate_spatial_window.json
+++ b/aggregate_spatial_window.json
@@ -1,8 +1,7 @@
 {
   "id": "aggregate_spatial_window",
   "summary": "Zonal statistics for rectangular windows",
-  "description": "Aggregates statistics over the spatial dimensions. This process passes a list of values to the reducer.\n\nThe results are computed dividing the input X/Y grid into x_size/y_size non-overlapping windows and applying the reducer process to each window.\nIf X and Y are not multiple of x_size/y_size NaNs padding before the computation is applied.",
-  "categories": [
+  "description": "Aggregates statistics over the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nThe pixel grid for the axes `x` and `y` is divided into non-overlapping windows with the size specified in the parameters `size` = [`x_size`,`y_size`]. If the number of values for the axes `x` and `y` is not a multiple of the corresponding `x_size` or `y_size` parameter, the behavior specified in `boundary` parameter is applied.\nFor each of these windows, the reducer process computes the result.",  "categories": [
     "cubes",
     "aggregate & resample"
   ],
--- a/aggregate_spatial_window.json
+++ b/aggregate_spatial_window.json
@@ -1,7 +1,6 @@
 {
   "id": "aggregate_spatial_window",
   "summary": "Zonal statistics for rectangular windows",
-  "description": "Aggregates statistics over the spatial dimensions. This process passes a list of values to the reducer.\n\nThe results are computed dividing the input X/Y grid into x_size/y_size non-overlapping windows and applying the reducer process to each window.\nIf X and Y are not multiple of x_size/y_size NaNs padding before the computation is applied.",
   "categories": [
     "cubes",
     "aggregate & resample"
@@ -17,7 +16,6 @@
     },
     {
       "name": "reducer",
-      "description": "A reducer to be applied on all values of each geometry. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes.",
       "schema": {
         "type": "object",
         "subtype": "process-graph",
@@ -44,6 +42,7 @@
         ]
       }
     },
+<<<<<<< HEAD
 	{
 		"name": "size",
 		"description": "Window sizes (pixels) along each dimension.\n\nFirst value correspond to Window size along X or Longitude and second to Window size along Y or Latitude.",
@@ -66,6 +65,28 @@
 			}
 		}
 	},
+=======
+    {
+      "name": "x_size",
+      "description": "Window size along X or Longitude. The number of values used as input to the reducer along X dimension.",
+      "schema": {
+        "type": "integer",
+        "minimum": 1
+      },
+      "optional": true,
+      "default": 1
+    },
+    {
+      "name": "y_size",
+      "description": "Window size along Y or Latitude. The number of values used as input to the reducer along Y dimension.",
+      "schema": {
+        "type": "integer",
+        "minimum": 1
+      },
+      "optional": true,
+      "default": 1
+    },
+>>>>>>> e2fa4060c12f133096b332df2d10f37e0503b365
     {
       "name": "boundary",
       "description": "Operation to apply if X/Y are not multiple of x_size/y_size. Options are:\n\n- pad (default): pad the datacube with NaNs to fit window dimensions.\n\n- trim: trim the datacube to fit window dimensions.",
@@ -93,7 +114,5 @@
     "description": "A data cube with the newly computed values and the same dimensions.\n\nThe resolution will change, depending on the chosen parameters. It decreases for the dimensions which have the corresponding parameter `size` = [`x_size`,`y_size`] set to values greater than 1.\n\nThe dimension labels will be set to the center of the window. The other dimension properties (name, type and reference system) remain unchanged.",
     "schema": {
       "type": "object",
-      "subtype": "vector-cube"
     }
   }
-}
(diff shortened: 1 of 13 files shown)


aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
+{
+  "id": "aggregate_spatial_window",
+  "summary": "Zonal statistics for rectangular windows",
+  "description": "Aggregates statistics over the spatial dimensions. This process passes a list of values to the reducer.\n\nThe results are computed dividing the input X/Y grid into x_size/y_size non-overlapping windows and applying the reducer process to each window.\nIf X and Y are not multiple of x_size/y_size NaNs padding before the computation is applied.",
```
m-mohr, line 4: Trying to align a bit more with other processes; can probably still be improved:
```suggestion
  "description": "Aggregates statistics over the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nThe pixel grid for the axes `x` and `y` is divided into non-overlapping windows with the size specified in the parameters `x_size` and `y_size`. If the number of values for the axes `x` and `y` is not a multiple of the corresponding `x_size` or `y_size` parameter, the behavior specified in `boundary` parameter is applied.\nFor each of these windows, the reducer process computes the result.",
```

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+  "categories": [
+    "cubes",
+    "aggregate & resample"
+  ],
+  "parameters": [
+    {
+      "name": "data",
+      "description": "A raster data cube.",
```
m-mohr, line 12: Does it have constraints? It seems like the input raster data cube must have only two dimensions? x and y...

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      "description": "A reducer to be applied on all values of each geometry. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes.",
+      "schema": {
+        "type": "object",
+        "subtype": "process-graph",
+        "parameters": [
+          {
+            "name": "data",
+            "description": "An array with elements of any type.",
+            "schema": {
+              "type": "array",
```
m-mohr, line 29: This passes an flat array, right? So for x_size = 2 and y_size = 2, it could be [5,4,1,7], right? I think this could make the descriptions easier.

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+            },
+            "optional": true,
+            "default": null
+          }
+        ]
+      }
+    },
+    {
+      "name": "x_size",
+      "description": "Window size along X or Longitude. The number of values used as input to the reducer along X dimension.",
```
m-mohr, line 49:
```suggestion
      "description": "Window size along the axis `x` in pixels.",
```

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      "schema": {
+        "type": "number"
+      },
+      "optional": true,
+      "default": 1
+    },
+    {
+      "name": "y_size",
+      "description": "Window size along Y or Latitude. The number of values used as input to the reducer along Y dimension.",
```
m-mohr, line 58:
```suggestion
      "description": "Window size along the axis `y` in pixels.",
```

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+              "description": "Any data type."
+            },
+            "optional": true,
+            "default": null
+          }
+        ]
+      }
+    },
+    {
+      "name": "x_size",
```
m-mohr, line 48: In `apply_neighborhood` we use `size = [2,5]` instead of x_size = 2 / y_size = 5. Should we align?

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      "description": "Window size along X or Longitude. The number of values used as input to the reducer along X dimension.",
+      "schema": {
+        "type": "number"
```
m-mohr, line 51: We should add `"minimum": 1`

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      },
+      "optional": true,
+      "default": 1
+    },
+    {
+      "name": "y_size",
+      "description": "Window size along Y or Latitude. The number of values used as input to the reducer along Y dimension.",
+      "schema": {
+        "type": "number"
```
m-mohr, line 60: We should add `"minimum": 1`

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      },
+      "optional": true,
+      "default": 1
+    },
+    {
+      "name": "boundary",
+      "description": "Operation to apply if X/Y are not multiple of x_size/y_size. Options are:\n\n- pad (default): pad the datacube with NaNs to fit window dimensions.\n\n- trim: trim the datacube to fit window dimensions.",
```
m-mohr, line 67:
```suggestion
      "description": "Behavior to apply if the number of values for the axes `x` and `y` is not a multiple of the corresponding `x_size` or `y_size` parameter. Options are:\n\n- `pad` (default): pad the data cube with `NaN` (not a number) values to fit the required window size.\n\n- `trim`: trim the data cube to fit the required window size.",
```

Open questions: What is trimmed? Equally on each side? At the "end"? At the "beginning"?

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      "description": "Additional data to be passed to the reducer.",
+      "schema": {
+        "description": "Any data type."
+      },
+      "optional": true,
+      "default": null
+    }
+  ],
+  "returns": {
+    "description": "A vector data cube with the computed results.\n\nX and Y dimensions will be smaller compared to the input datacube.",
```
m-mohr, line 89: There are quite some information about the resulting data cube missing. I'm just assuming some things in my suggestion, feel free to change whatever makes more sense:

```suggestion
    "description": "A data cube with the newly computed values and the same dimensions. The resolution will change, depending on the chosen parameters. It decreases for the dimensions which have the corresponding parameter (`x_size`/`y_size`) set to a value greater than 1. The dimension labels will be set to the center of the window. The other dimension properties (name, type and reference system) remain unchanged.",
```

m-mohr (commented):
Made a couple of detailed review comments. Afterwards, have a look at the CI results (currently fails) and change the JSON file accordingly, please.

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+    "schema": {
+      "type": "object",
+      "subtype": "vector-cube"
```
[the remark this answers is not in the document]
clausmichele: My starting point was the json of aggregate_spatial https://github.com/Open-EO/openeo-processes/blob/master/aggregate_spatial.json maybe that's also wrong?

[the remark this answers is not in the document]
m-mohr: No, that is correct. aggregate_spatial gets vectors passed and returns the values for them...

aggregate_spatial_window.json
clausmichele: In the implementation I am working on, it just needs to have x and y, if it is a timeseries then it will be applied on the full timeseries and return another timeseries. So the datacube dimensions could be (time,x,y) or (x,y) or (band,x,y) or (time,band,x,y) and so on. But we can also choose to limit it to 2d datacubes to ease the implementation.

m-mohr: Let's discuss later in the meeting.

aggregate_spatial_window.json
clausmichele: Well, it depends if the array has the first two dimensions which corresponds to x and y.
If the input has dimension in this order (x,y,band,time) and size [5,4,1,7] and we apply a window with x_size = 2 and y_size = 2 the output would be [2,2,1,7] for boundary 'trim' option or [3,2,1,7] for 'pad'. Is it more clear now?

aggregate_spatial_window.json
clausmichele: yes, we can use the same as apply_neighborhood.

aggregate_spatial_window.json
clausmichele: Good question. Actually I'm using xarray coarsen method, which has also the option to align top-left or top-right the windows http://xarray.pydata.org/en/v0.15.1/generated/xarray.DataArray.coarsen.html
So, if aligned to the left it will trim the excess on the right and vice versa.

aggregate_spatial_window.json
m-mohr: Yeah, maybe we should just take over the definition from https://github.com/Open-EO/openeo-processes/blob/75059c6566a7c8951d04467665252e74d5bb9c3a/apply_neighborhood.json#L45

It has its own subtype `chunk-size` and thus I envision good client support. We can remove the duration of course and may want to remove the null behavior (all values).

aggregate_spatial_window.json
m-mohr: Yeah, maybe we should just take over the definition from https://github.com/Open-EO/openeo-processes/blob/master/apply_neighborhood.json#L44-L91

It has its own subtype `chunk-size` and thus I envision good client support. We can remove the duration of course and may want to remove the null behavior (all values).

aggregate_spatial_window.json
m-mohr: resample_spatial has the align parameter, which is very similar.

aggregate_spatial_window.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "aggregate_spatial_window",
   "summary": "Zonal statistics for rectangular windows",
-  "description": "Aggregates statistics over the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nThe pixel grid for the axes `x` and `y` is divided into non-overlapping windows with the size specified in the parameters `x_size` and `y_size`. If the number of values for the axes `x` and `y` is not a multiple of the corresponding `x_size` or `y_size` parameter, the behavior specified in `boundary` parameter is applied.\nFor each of these windows, the reducer process computes the result.",
+  "description": "Aggregates statistics over the spatial dimensions. This process passes a list of values to the reducer.\n\nThe results are computed dividing the input X/Y grid into x_size/y_size non-overlapping windows and applying the reducer process to each window.\nIf X and Y are not multiple of x_size/y_size NaNs padding before the computation is applied.",
```
m-mohr, line 4: Why have you reverted this?

clausmichele: My mistake

clausmichele: Is it ok now?

aggregate_spatial_window.json
clausmichele: I updated the description as you have suggested, we still need to decide what to do for the alignment.

clausmichele:
I don't know why the build fails, it tells me that the word 'behavior' is misspelled

m-mohr:
@clausmichele I guess it wants you to use 'behaviour'. Other than that it's failing because the JSON needs to be formatted with 4 spaces. Running the JSON through https://jsonformatter.org/ with 4 Tab Spaces should make it work.

aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
+{
+    "id": "aggregate_spatial_window",
+    "summary": "Zonal statistics for rectangular windows",
+    "description": "Aggregates statistics over the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nThe pixel grid for the axes `x` and `y` is divided into non-overlapping windows with the size specified in the parameters `size` = [`x_size`,`y_size`]. If the number of values for the axes `x` and `y` is not a multiple of the corresponding `x_size` or `y_size` parameter, the behaviour specified in `boundary` parameter is applied.\nFor each of these windows, the reducer process computes the result.",
```
m-mohr, line 4: Simplified a bit:
```suggestion
    "description": "Aggregates statistics over the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nThe pixel grid for the axes `x` and `y` is divided into non-overlapping windows with the size specified in the parameter `size`. If the number of values for the axes `x` and `y` is not a multiple of the window size, the behaviour specified in `boundary` parameter is applied.\nFor each of these windows, the reducer process computes the result.",
```

aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
...
+            },
+            "optional": true,
+            "default": null
+        }
+    ],
+    "returns": {
+        "description": "A data cube with the newly computed values and the same dimensions.\n\nThe resolution will change, depending on the chosen parameters. It decreases for the dimensions which have the corresponding parameter `size` = [`x_size`,`y_size`] set to values greater than 1.\n\nThe dimension labels will be set to the center of the window. The other dimension properties (name, type and reference system) remain unchanged.",
+        "schema": {
+            "type": "object",
+            "subtype": "vector-cube"
```
m-mohr, line 95: It seems a couple of things have been reverted. Please go through all the comments again and check whether there are still changes to be done. I found a couple of them...

```suggestion
            "subtype": "raster-cube"
```

aggregate_spatial_window.json
m-mohr: This still needs clarification

aggregate_spatial_window.json
```diff
@@ -0,0 +1,95 @@
...
+      "name": "data",
+      "description": "A raster data cube.",
+      "schema": {
+        "type": "object",
+        "subtype": "raster-cube"
+      }
+    },
+    {
+      "name": "reducer",
+      "description": "A reducer to be applied on all values of each geometry. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes.",
```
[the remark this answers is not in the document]
m-mohr: This hasn't been applied yet

aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
...
+                            "minimum": 1,
+                            "description": "The unit the values are given in pixels (`px`)"
+                        }
+                    }
+                }
+            }
+        },
+        {
+            "name": "boundary",
+            "description": "Behavior to apply if the number of values for the axes `x` and `y` is not a multiple of the corresponding `size` = [`x_size`,`y_size`] parameter. Options are:\n\n- `pad` (default): pad the data cube with `NaN` (not a number) values to fit the required window size.\n\n- `trim`: trim the data cube to fit the required window size.",
```
m-mohr, line 70:
```suggestion
            "description": "Behavior to apply if the number of values for the axes `x` and `y` is not a multiple of the corresponding value in the `size` parameter. Options are:\n\n- `pad` (default): pad the data cube with `NaN` (not a number) values to fit the required window size.\n\n- `trim`: trim the data cube to fit the required window size.",
```

aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
...
+                        },
+                        "optional": true,
+                        "default": null
+                    }
+                ]
+            }
+        },
+        {
+            "name": "size",
+            "description": "Window sizes (pixels) along each dimension.\n\nFirst value correspond to Window size along X or Longitude and second to Window size along Y or Latitude.",
```
m-mohr, line 49:
```suggestion
            "description": "Window sizes in pixels along the horizontal spatial dimensions.\n\nThe first value corresponds to the `x` axis, the second values corresponds to the `y` axis.",
```

m-mohr (changes requested):

aggregate_spatial_window.json
m-mohr: Yes, I'm unsure yet. I'll put this into proposal stage first (see #196) and then we can elaborate during development whether a paramter should be added. For now we should just specify a default, either upper-left or lower-right. I think I'd default to pad the lower-right. By the way, why do we fill with NaN instead of null/no-data? That seems the approach more aligned with openEO. NaN is actually rarely used here...

clausmichele: I was reporting what the Xarray library does, but we are free to choose something else.

m-mohr: I think no-data / null makes more sense in the context of openEO. no data is implemented into several processes, but NaN is not (really).

m-mohr:
I completely lost track in this PR of what still needs to be done. Too many comments and commits. Let's merge this into a new branch here and make a new PR for more detailed discussions that then merges into draft.

aggregate_spatial_window.json
m-mohr: No, I don't understand this. I assume it's flat, as this is how the spec works.

clausmichele: With [5,4,1,7] you mean an array with values 5,4,1 and 7? Which has shape (y,x) = (1,4) ? Then you could use a window with size = [2,2] but you must use _pad_  for applying it.

aggregate_spatial_window.json
clausmichele: what about:
 "description": "A raster data cube with two spatial dimensions and arbitrary number of bands and times. If the datacube has multiple bands the process will be applied to all of them. Same for a timeseries."

Sorry I'm not good in writing this kind of descriptions.

aggregate_spatial_window.json
clausmichele: Ok I'll put no-data instead of Nan.

PR #197, opened 2020-09-22
Establish proposal procedure #196

### LINKED ISSUE #196: Establish proposal procedure
With new processes coming in, it's sometimes not easy to decide whether to make them core and release them as part of the "stable contract" or not. Currently an example is #192, which seems like a useful process, but alternatives are also discussed. Nevertheless, an openEO partner wants such a process and thus it's good to give implementers a broader choice for use cases, which they can built on as this leads to less dispersion. Also, the process should be implemented first, so that we can make changes if required and then afterwards we could push into core. Surely, these processes should all make clear by "experimental: true" that changes may occur. 

Anyway, long story short: We should make a folder or branch for proposals or experimental processes that are in development and have some procedures around it. Do you agree?

m-mohr:
Establish a proposal procedure, see #196

Closes #196


mkadunc (approved):

README.md
```diff
@@ -25,7 +27,8 @@ See also the [changelog](CHANGELOG.md) for the changes between versions and the
 
 This repository contains a set of files formally describing the openEO Processes:
 
-* The `*.json` files provide the process specifications as defined by the openEO API.
+* The `*.json` files provide the stable process specifications as defined by openEO. New processes need at least two implementations.
+* The `*.json` files in the [`proposals`](proposals/) folder provide proposed new process specifications that are only implemented by one organization and may still change. Everyone is encouraged to base their work on the proposals and give feedback so that eventually the processes evolve into stable proces specifications.
```
soxofaan, line 31: This reads like a very strong coupling between the API and implementing backends: stable=2 or more backends, experimental=1 backend. Is it necessary to be that strict about that? It seems like the decision process is reduced to a simple count of implementations.

As far as I experienced process discussions it was most of the times pretty clear where a process was on the scale "no-brainer" vs "needs more thought or real world testing", regardless of number of actual implementations.

README.md
```diff
@@ -25,7 +27,8 @@ See also the [changelog](CHANGELOG.md) for the changes between versions and the
 
 This repository contains a set of files formally describing the openEO Processes:
 
-* The `*.json` files provide the process specifications as defined by the openEO API.
+* The `*.json` files provide the stable process specifications as defined by openEO. New processes need at least two implementations.
```
m-mohr, line 30:
```suggestion
* The `*.json` files provide the stable process specifications as defined by openEO. New processes need at least two implementations or consensus from the openEO PSC.
```

README.md
m-mohr: In the future discussions will be different due to the end of the "core" openEO project. A TSC will be formed (and that will lead to a different ruleset again). I added "or consensus from the openEO PSC".

soxofaan: > Proposals don't need an implementation, by the way.

then maybe also relax:

> files in the proposals folder provide proposed new process specifications that are only implemented by one organization and may still change

e.g: files in the proposals folder are proposals for new process specifications that are still experimental and subject to change.

soxofaan (approved):

README.md
m-mohr:
```suggestion
* The `*.json` files in the [`proposals`](proposals/) folder provide proposed new process specifications that are still experimental and subject to change. Ideally, each specification is backed by an implementation. Everyone is encouraged to base their work on the proposals and give feedback so that eventually the processes evolve into stable process specifications.
```

PR #198, opened 2020-09-22
Tweak aggregate_spatial_window

### LINKED ISSUE #187: Which process for mean over a window with no overlap

m-mohr:
Further tweaks for aggregate_spatial_window, based on #192. The old PR got messy and this tries to do the final tweaks...

Closes #187.

~Will be moved to proposals once #197 got merged.~ Done.


aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
...
+    "summary": "Zonal statistics for rectangular windows",
+    "description": "Aggregates statistics over the horizontal spatial dimensions (axes `x` and `y`) of the data cube.\n\nThe pixel grid for the axes `x` and `y` is divided into non-overlapping windows with the size specified in the parameter `size`. If the number of values for the axes `x` and `y` is not a multiple of the window size, the behaviour specified in `boundary` parameter is applied.\nFor each of these windows, the reducer process computes the result.",
+    "categories": [
+        "cubes",
+        "aggregate & resample"
+    ],
+    "parameters": [
+        {
+            "name": "data",
+            "description": "A raster data cube.",
```
m-mohr, line 13: Clarification needed, see https://github.com/Open-EO/openeo-processes/pull/192#discussion_r490087458

> Does it have constraints? It seems like the input raster data cube must have only two dimensions? x and y...

Proposal from @clausmichele:
> A raster data cube with two spatial dimensions and arbitrary number of bands and times. If the datacube has multiple bands the process will be applied to all of them. Same for a timeseries.

aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
...
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "reducer",
+            "description": "A reducer to be applied on all values of each geometry. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes.",
```
m-mohr: Do we need to specify that the array is flattened? E.g. a 2x2 windows results in [1,2,3,4] instead of [[1,2], [3,4]]?

clausmichele (approved):
Ok for me, thanks for polishing it!

aggregate_spatial_window.json
```diff
@@ -0,0 +1,98 @@
...
+                            "type": "number",
+                            "minimum": 1,
+                            "description": "The unit the values are given in pixels (`px`)"
+                        }
+                    }
+                }
+            }
+        },
+        {
+            "name": "boundary",
```
m-mohr, line 62: Open questions: What is trimmed? Where is the padding added? Equally on each side? At the "end"? At the "beginning"? See https://github.com/Open-EO/openeo-processes/pull/192#discussion_r490122368

@clausmichele 
> Good question. Actually I'm using xarray coarsen method, which has also the option to align top-left or top-right the windows http://xarray.pydata.org/en/v0.15.1/generated/xarray.DataArray.coarsen.html
So, if aligned to the left it will trim the excess on the right and vice versa.

@m-mohr 
>  I'll put this into proposal stage first (see #196) and then we can elaborate during development whether a paramter should be added. For now we should just specify a default, either upper-left or lower-right. I think I'd default to pad the lower-right.

aggregate_spatial_window.json
```diff
@@ -0,0 +1,91 @@
...
+            "name": "data",
+            "description": "A raster data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "reducer",
+            "description": "A reducer to be applied on the list of values, which contain all pixels covered by the window. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes.",
```
m-mohr, line 21: Do we need to specify that the array is flattened? E.g. a 2x2 windows results in [1,2,3,4] instead of [[1,2], [3,4]]?

m-mohr: > Ok for me, thanks for polishing it!

@clausmichele Thanks! Can you check the remaining comments here and add your opinion?

m-mohr:
> Ok for me, thanks for polishing it!

@clausmichele Thanks! Can you check the remaining comments here and add your opinion?

aggregate_spatial_window.json
clausmichele: I don't think its necessary since the result remains the same. How you pass/process the actual data it's just an implementation detail in my opinion.

m-mohr: Agreed. Just want to make sure there's no misunderstanding here => resolved.

aggregate_spatial_window.json
clausmichele: I would say that pad or trim on the right/bottom could be default, with the window aligned top-left at the beginning.

m-mohr: Sorry for leaving this open so long. I've now added a parameter `align` as it is available in the resample process. This should now clarify all alignment issues. Please review the changes.

aggregate_spatial_window.json
m-mohr: @clausmichele Do you think the following wording would work?

> A raster data cube with exactly two horizontal spatial dimensions and an arbitrary number of additional dimensions. The process is applied to all additional dimensions individually.

clausmichele: @m-mohr yes the description is fine!

clausmichele (approved):

clausmichele (approved):

Issue #199, opened 2020-09-22
null handling in `all`


soxofaan:
Some questions about description/definition of `all` process:

-   > if ... the array is empty ... the process returns null.

    what is the background reasoning here? FYI: in python `all([]) == True` (see .e.g. https://stackoverflow.com/questions/19601802/how-does-all-in-python-work-on-empty-lists)

-   > If only one value is given, the process evaluates to the given value.
   
    is it necessary to explicitly document the single value case? It conflicts a bit with the `ignore_nodata=True` case: if given `[null]`, should one use that single value, or ignore the `null` and handle it as empty list?  I know that it is the same (currently, see higher), but it makes it a bit confusing.

-   > .. so that null is a valid logical object

    what does this mean? Like a "third boolean value"?

-  > If a component is null, the result will be null if the outcome is ambiguous

    I don't know to handle the two if's in that sentence. And what does "ambiguous" mean here (how would the result be ambiguous)? 

(the same probably applies to `any`, but I haven't read it in that much detail yet)

m-mohr:
> Some questions about description/definition of `all` process:
> 
> * > if ... the array is empty ... the process returns null.
>   
>   
>   what is the background reasoning here?

I can't remember the discussion, but I think the reasoning was: No data is given => no data value should be returned. For me this sounds more logical than assigning empty arrays to true or false, which both could be argued to be valid return values. It's likely also to be consistent with your example below `[null] => null`.


>   is it necessary to explicitly document the single value case?

What are you referring to?

> * > .. so that null is a valid logical object
>   
>   what does this mean? Like a "third boolean value"?

That just means that null can occur in a context that usually only expects booleans and need to be handled correctly. Indeed this sounds a bit strange. What is a "logical object"? Maybe it was meant to be "entity" or "value". Not sure. Maybe we should just remove " so that `null` is a valid logical object"? What do you think?


> * > If a component is null, the result will be null if the outcome is ambiguous
>   
>   
>   I don't know to handle the two if's in that sentence. And what does "ambiguous" mean here (how would the result be ambiguous)?
> 
> (the same probably applies to `any`, but I haven't read it in that much detail yet)

This refers to the following examples:
* `all(data = [false,null], ignore_nodata = false) => false` -> A single false will always lead to false independently of no-data (unambiguous). false/true and false/false are both false.
[...]

PR #200, opened 2020-09-24
Clarify all and any #199

m-mohr:
see #199


all.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Checks if **all** of the values in `data` are true. Evaluates all values from the first to the last element and stops once the outcome is unambiguous. If no value is given (i.e. the array is empty) the process returns `null`.\n\nBy default all no-data values are ignored so that the process returns `null` if all values are no-data, `true` if all other values are true and `false` otherwise. Setting the `ignore_nodata` flag to `false` considers no-data values. If a component is `null`, the result will be `null` if the outcome is ambiguous. See the following truth table for any two consecutive values in the array:\n\n```\n      || null  | false | true\n----- || ----- | ----- | -----\nnull  || null  | false | null\nfalse || false | false | false\ntrue  || null  | false | true\n```",
```
soxofaan, line 4: there is still this hard to read "if .., the result will be ... if ..." construct here

soxofaan: Proposed replacement for 

>  Setting the `ignore_nodata` flag to `false` considers no-data values. If a component is `null`, the result will be `null` if the outcome is ambiguous. See the following truth table for any two consecutive values in the array:

->

>  Setting the `ignore_nodata` flag to `false` takes no-data values into account, handling them as an "ambiguous" value. The array values are reduced pairwise according to the truth table below. A `null` array item makes the result ambiguous, unless there is a `false` in the data, which makes the result unambiguously `false`.

any.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Checks if **any** (i.e. at least one) value in `data` is `true`. Evaluates all values from the first to the last element and stops once the outcome is unambiguous. If no value is given (i.e. the array is empty) the process returns `null`.\n\nBy default all no-data values are ignored so that the process returns `null` if all values are no-data, `true` if at least one of the other values is true and `false` otherwise. Setting the `ignore_nodata` flag to `false` considers no-data values. If a component is `null`, the result will be `null` if the outcome is ambiguous. See the following truth table for any two consecutive values in the array:\n\n```\n      || null | false | true\n----- || ---- | ----- | ----\nnull  || null | null  | true\nfalse || null | false | true\ntrue  || true | true  | true\n```",
```
soxofaan, line 4: same remark as with `all`

soxofaan (changes requested):
see suggested changes about explaining "ambiguity" in the truth table

all.json
m-mohr: Thanks for the proposal. I tried to be more concise and leave it mostly up to the table to make it clear how things resolve. The "reducing" part of your proposal helps with that. Not we don't really need to explain and just let users check the table. See https://github.com/Open-EO/openeo-processes/pull/200/commits/747868f5052f3d2269a09477a2c1d9746cf32217 What do you think?

all.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Checks if **all** of the values in `data` are true. Evaluates all values from the first to the last element and stops once the outcome is unambiguous. A result is ambiguous unless a value is `false`. If no value is given (i.e. the array is empty) the process returns `null`.\n\nBy default all no-data values are ignored so that the process returns `null` if all values are no-data, `true` if all values are true and `false` otherwise. Setting the `ignore_nodata` flag to `false` takes no-data values into account and the array values are reduced pairwise according to the following truth table:\n\n```\n      || null  | false | true\n----- || ----- | ----- | -----\nnull  || null  | false | null\nfalse || false | false | false\ntrue  || null  | false | true\n```",
```
soxofaan, line 4: That is indeed already better.

One note though about  this in `all` description:

> A result is ambiguous unless a value is `false`.

I think this is a misleading statement: if all input values are `true`, the result is `true`, without ambiguity.

We could move the whole thing about ambiguity based short circuiting ("evaluates ... from first ... and stops once ... unambiguous ...") to the end as a implementation detail note (or drop it all together). For the user it is not that relevant after all. And in some implementations it might not be possible to stop as early as possible (e.g. because of  parallelization  aspects).

m-mohr: Huh, there's something missing. 
My intention was to write: "A result is ambiguous unless a value is `false` or all values have been taken into account."

I agree that we should move it to the end.

m-mohr: Hey @soxofaan,

sorry, it took me a while to get back to this. I moved the ambiguity details to the end and fixed the all description. Is this now ready to be merged?

soxofaan: yes fine for me

soxofaan (approved):

PR #206, opened 2020-10-28
Clarify contradicting statements in filter_temporal

### LINKED ISSUE #203: Ambiguous filter_temporal process definition
While implementing `filter_temporal` process we have stumbled across some inconsistencies in the [process definition](https://processes.openeo.org/#filter_temporal). Note that none of it is a blocker for us, but it is probably best to document such issues when we encounter them. Let us know if you would prefer some other channel of communication for this. 

Also, if you would prefer a PR (IIUC, it should be fixed [here](https://github.com/Open-EO/openeo-processes/blob/master/filter_temporal.json)) I'd be happy to do it, but would appreciate a confirmation that these suggestions make sense. 

-----
The [filter_temporal process](https://processes.openeo.org/#filter_temporal) documentation is not clear on whether the `dimension` parameter specifies a single or multiple dimensions:

> If the dimension is set to null (it's the default value), the data cube is expected to only have one temporal dimension.

> If the dimension is not set or is set to null, the filter applies to all temporal dimensions.

> Return value: ... remain unchanged, except that the given temporal dimension(s) have less (or the same) dimension labels.

Since the data types do not suggest a way to provide multiple extents and/or dimensions, I am assuming that this process can only operate on a **single** temporal dimension. If this is so, I would suggest changing the texts to read:

> If the dimension is set to null (it's the default value), the data cube is expected to have **exactly** one temporal dimension.

> If the dimension is not set or is set to null, the filter applies to **the only temporal dimension that exists (and must exist) in data cube**.

> Return value: ... remain unchanged, except that the given temporal **dimension has less (or the same) dimension labels**.

-----

Additionally, this exception is not available in [the list of errors](https://github.com/Open-EO/openeo-api/blob/master/errors.json):

> Fails with a DimensionNotAvailable error if the specified dimension does not exist.

> Errors/Exceptions
>    DimensionNotAvailable
>      *Message: A dimension with the specified name does not exist.*

I assume that `ProcessParameterInvalid` should be used instead and that other possible messages should also be included. One option would be:

> Fails with a `ProcessParameterInvalid` error if the specified dimension does not exist.

> Errors/Exceptions
>    ProcessParameterInvalid
>      *Message: A dimension with the specified name does not exist.*
>      *Message: A dimension with the specified name is not temporal.*
>      *Message: More than one temporal dimension available, please specify dimension.*
>      *Message: No temporal dimension is available.*

m-mohr:
Clarify contradicting statements in filter_temporal for the default value of the dimension parameter. By default all temporal dimensions are affected by the process. Fixes #203

### DIFF
--- a/filter_temporal.json
+++ b/filter_temporal.json
@@ -73,7 +73,7 @@
         }
     ],
     "returns": {
-        "description": "A data cube restricted to the specified temporal extent. The dimensions and dimension properties (name, type, labels, reference system and resolution) remain unchanged, except the temporal dimensions may have less dimension labels depending on the value for the `dimension` parameter.",
+        "description": "A data cube restricted to the specified temporal extent. The dimensions and dimension properties (name, type, labels, reference system and resolution) remain unchanged, except that the temporal dimensions (determined by `dimensions` parameter) may have less dimension labels.",
         "schema": {
             "type": "object",
             "subtype": "raster-cube"
--- a/filter_temporal.json
+++ b/filter_temporal.json
@@ -1,7 +1,7 @@
 {
     "id": "filter_temporal",
     "summary": "Temporal filter for a temporal intervals",
-    "description": "Limits the data cube to the specified interval of dates and/or times.\n\nMore precisely, the filter checks whether the temporal dimension label is greater than or equal to the lower boundary (start date/time) and the temporal dimension label is less than the value of the upper boundary (end date/time). This corresponds to a left-closed interval, which contains the lower boundary but not the upper boundary.\n\nIf the dimension is set to `null` (it's the default value), the data cube is expected to only have one temporal dimension.",
+    "description": "Limits the data cube to the specified interval of dates and/or times.\n\nMore precisely, the filter checks whether each of the temporal dimension labels is greater than or equal to the lower boundary (start date/time) and less than the value of the upper boundary (end date/time). This corresponds to a left-closed interval, which contains the lower boundary but not the upper boundary.",
     "categories": [
         "cubes",
         "filter"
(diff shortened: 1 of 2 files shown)


filter_temporal.json
```diff
@@ -73,7 +73,7 @@
         }
     ],
     "returns": {
-        "description": "A data cube restricted to the specified temporal extent. The dimensions and dimension properties (name, type, labels, reference system and resolution) remain unchanged, except that the given temporal dimension(s) have less (or the same) dimension labels.",
+        "description": "A data cube restricted to the specified temporal extent. The dimensions and dimension properties (name, type, labels, reference system and resolution) remain unchanged, except the temporal dimensions may have less dimension labels depending on the value for the `dimension` parameter.",
```
sinergise-anze, line 76:
```
... except the temporal dimensions may have less dimension labels depending on the value for the `dimension` parameter.",
```

Everything else looks good, I would just rephrase this sentence, because it could be understood as if the labels depend on `dimension` parameter (they depend on `extent`).

I would suggest:
```
...except the temporal dimensions (determined by `dimensions` parameter) may have less dimension labels.",
```

Not a big deal though, it is clear from context what is meant.

sinergise-anze (approved):

filter_temporal.json
m-mohr: Good suggestion!

```suggestion
        "description": "A data cube restricted to the specified temporal extent. The dimensions and dimension properties (name, type, labels, reference system and resolution) remain unchanged, except that the temporal dimensions (determined by `dimensions` parameter) may have less dimension labels.",
```

PR #209, opened 2020-11-11
add definition of atmospheric correction process

jdries:
Starting a PR to gather discussions on atmospheric correction process!

One open question from me: what about cloud detection?

- Create separate process?
- Can we allow for immediately applying a cloud mask? In on-the-fly use cases, users may not require a separate mask, but simply want to remove all clouded pixels immediately.
- In a lot of other scenarios, users do want to further play with cloud masks, like applying erosion/dilation before applying it to the data.

My proposal is purely to get discussion started, feel free to comment!


m-mohr:
Some general comments: 

- Please issue the PR against the "draft" branch and there put it into the "proposals" folder.
- Parameter names are not allowed to contain dashes (-), use underscore instead (_).
- From experience with the other processes, it's much easier to specify and implement processes that do exactly one thing, not multiple things at once (regarding the clouds)
- What exactly is expected to be passed to elevation-model?
- I guess in algorithm (I think our other processes usually call it "method") should provide a list of values in an enum, at least on the back-end side. 
- Not going into details with other spec issues here, let's first draft an idea and then go into details about spec conformance.

m-mohr:
Thanks @jdries. Just changing the base branch here is not enough. You still need to merge draft into your branch. See the "Files changed", which now lists more files than expected.

jdries:
Ok, seems better now, there's a few more changed files, from a commit on master that was not yet merged into draft. Perhaps we can merge master into draft?

m-mohr:
Well, that's just version number changes that we don't need in draft and then the merge is empty. I think the easiest solution is to close this PR, create a new branch, copy paste your file (or cherry-pick) and create a new PR.

m-mohr:
Closed in favor of #210

PR #210, opened 2020-11-11
atmospheric_correction, cloud_detection, sar_backscatter process

jdries:
Starting a PR to gather discussions on atmospheric correction process!

One open question from me: what about cloud detection?

    Create separate process?
    Can we allow for immediately applying a cloud mask? In on-the-fly use cases, users may not require a separate mask, but simply want to remove all clouded pixels immediately.
    In a lot of other scenarios, users do want to further play with cloud masks, like applying erosion/dilation before applying it to the data.

My proposal is purely to get discussion started, feel free to comment!


clausmichele:
A cloud-removal process would be useful. Should it refer to a specific dataset with cloud masks automatically? Should the loaded data have a band for the cloud mask? The less setting it has, the better it is in my opinion. Then everything has to be handled by the back-end.

jdries:
I added processes for backscatter and cloud detection.

An, optical ARD process would now look like:
```
#default params, mask pixels in-place
cube.cloud_detection().atmospheric_correction()

#add cloud mask band, leave clouds in the data
cube.cloud_detection(apply_mask=False).atmospheric_correction()

#specify a method
cube.cloud_detection().atmospheric_correction(method="FORCE")
```

m-mohr:
Is there a reasonable way to test the processes in the (hopefully upcoming) test suite when no method is set?

jdries:
> Is there a reasonable way to test the processes in the (hopefully upcoming) test suite when no method is set?

This could get rather difficult:
- atmospheric correction can be tested to not crash, but there's no unambiguously defined output for a given input
- backscatter depends on metadata that you probably will not have in a simply array, so needs to be tested in combination with load_collection, and as we know, not all backends may have all Sentinel-1 data.
- cloud_detection: also not unambiguous, but here you could choose an input combination that is absolutely clouded/cloud free, in which all detectors should have a consistent output.

jdries:
Here's a first (somewhat correct) image generated on the fly with our new atmospheric_correction process:
image

Still a way to go for performance, validation and DIAS deployment!

przell:
A comment from the meeting with Patrick:
CARD4L Per Pixel Data are e.g. Cloud Mask, Shadow, Aerosol Optical Thickness, Viewing Angles... 
Could we implement the production and provision of some of these via the processes discussed here?
Goes somewhat in the direction what has been mentioned by @jdries here:
> #default params, mask pixels in-place
> cube.cloud_detection().atmospheric_correction()
> 
> #add cloud mask band, leave clouds in the data
> cube.cloud_detection(apply_mask=False).atmospheric_correction()
> 
> #specify a method
> cube.cloud_detection().atmospheric_correction(method="FORCE")

jdries:
Also requestesd: BRDF and topographic correction.
FORCE includes them:
https://github.com/davidfrantz/force/blob/master/src/lower-level/topo-ll.c
https://github.com/davidfrantz/force/blob/master/src/lower-level/brdf-ll.c

Question: topographic correction appears to be part of atmospheric correction, but what about brdf? Can that also be a separate process?

lforesta:
> Question: topographic correction appears to be part of atmospheric correction, but what about brdf? Can that also be a separate process?

@jdries BRDF processing is part of the atmospheric correction in FORCE, like topographic correction (both can be flagged True/False). As for running these algorithms as independent processes, I did not have time yet to test, hopefully next week.

jdries:
These are processing options available at SentinelHub:
https://docs.sentinel-hub.com/api/latest/data/sentinel-1-grd/#processing-options

m-mohr:
I've made a couple of formal (non-content) changes to the processes. 

FYI:
- If `null` is the default value, it must be listed as a data type in `type`
- Pre-defined values must be listed as `enum`.
- Added preliminary summaries and categories (probably needs more work)
- Reformatted to be consistent with other processes
- Added `experimental` flag for the proposals
- Will also add a couple of suggestions directly via the GH review process.
- Will likely also change the wording a bit to be more consistent with other processes (later?)

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,78 @@
...
+            "schema": {
+                "type": "string",
+                "enum": [
+                    "sigma0",
+                    "gamma0"
+                ]
+            }
+        },
+        {
+            "description": "By default, orthorectification is disabled. Set to `true` to enables orthorectification.\n\nThe non-orthorectified products use a simple earth model as provided in the products themselves. This may be sufficient for very flat target areas and is faster to process.",
```
m-mohr, line 33: > The non-orthorectified products use a simple earth model as provided in the products themselves.

This also sounds like a (very basic) orthorectification, but it says also that false disables orthorectification. That is a bit confusing.

Also, I added a typo here: ;-)
```suggestion
            "description": "By default, orthorectification is disabled. Set to `true` to enable orthorectification.\n\nThe non-orthorectified products use a simple earth model as provided in the products themselves. This may be sufficient for very flat target areas and is faster to process.",
```

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,78 @@
...
+            "name": "orthorectify",
+            "optional": true,
+            "default": false,
+            "schema": {
+                "type": "boolean"
+            }
+        },
+        {
+            "description": "The (collection) name of the digital elevation model to use, leave empty to allow the back-end to make a suitable choice. Not all back-ends may offer all digital elevation models. Elevation models are listed as collections, or can be found in back-end specific documentation.",
```
m-mohr, line 42: Hmm... in openEO I'd expect this to be a load_collection that is then passed to the method. If this really should be just the collection identifier, we should add `"subtype": "collection-id"` (and split the schemas) so that client can help with choosing the id.

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,78 @@
...
+    ],
+    "returns": {
+        "description": "Backscatter values in linear power corresponding to the chosen backscatter coefficient. For decibel values, an additional conversion is needed.",
+        "schema": {
+            "subtype": "raster-cube",
+            "type": "object"
+        }
+    },
+    "exceptions": {
+        "UnsupportedParameter": {
```
m-mohr, line 62: For all three processes: I don't think these exception are required. The acceptable values should be listed in the enum and thus this can be detected in validation already and doesn't need to be thrown during runtime.

proposals/cloud_detection.json
```diff
@@ -0,0 +1,65 @@
+{
+    "id": "cloud_detection",
+    "summary": "Cloud detection",
+    "description": "Detects clouds and cloud shadow in optical imagery. By default, the cloud mask is applied to the input data cube. Set `apply_mask` to `false` to add the cloud mask as a new band.",
```
m-mohr, line 4: This description previously mentioned that a band is (always) added, but that conflicted with the apply_mask parameter. I changed it to comply with the parameter. If the desired behavior should be different, please change the description and parameter accordingly.

proposals/cloud_detection.json
```diff
@@ -0,0 +1,65 @@
...
+                ],
+                "enum": [
+                    "sen2cor",
+                    "fmask",
+                    "s2cloudless"
+                ]
+            }
+        },
+        {
+            "description": "By default, the cloud mask is applied to the input data cube. If set to `false`, the cloud mask information is added as one or more additional bands. Some implementations may add additional bands.",
```
m-mohr, line 37: I'm wondering whether this should be two processes as the behavior is somewhat different and in one case the data cube dimension labels change while in the other the data cube stays exactly as it is.

proposals/atmospheric_correction.json
```diff
@@ -0,0 +1,67 @@
...
+                    "null"
+                ],
+                "enum": [
+                    "FORCE",
+                    "iCor"
+                ]
+            }
+        },
+        {
+            "description": "The name of the digital elevation model to use, leave empty to allow the back-end to make a suitable choice. Not all back-ends may offer all digital elevation models.",
```
m-mohr, line 36: The spec of this parameter is different to the one in sar_backscatter. I'd propose to align them. Also the note that "Not all back-ends may offer all digital elevation models" should not be in here. If a back-end doesn't support DEMs then the parameter should be removed or so by the implementinf back-end. But just the generic info is confusing.

proposals/sar_backscatter.json
jdries: This option is fairly complicated:
- our sentinelhub backed backscatter offers different DEM's, that we do not have as a collection
- our orfeo based implementation will have access to the DEM's on DIAS, these could also be offered as collection, but orfeo has a DEM loading mechanism of its own

@lforesta  how will it be on EODC?

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,78 @@
...
+            "description": "Data cube containing raw backscatter input.",
+            "name": "data",
+            "schema": {
+                "subtype": "raster-cube",
+                "type": "object"
+            }
+        },
+        {
+            "description": "The backscatter coefficient to compute, either `sigma0` or `gamma0` (the default).",
+            "name": "backscatter_coefficient",
```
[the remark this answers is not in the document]
jdries: true, was copied from sentinelhub parameters.
@przell  any opinion?

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,78 @@
...
+    "summary": "Computes backscatter from georeferenced SAR input",
+    "description": "Computes backscatter from georeferenced SAR input.\n\nNote that backscatter computation may require instrument specific metadata that is tightly coupled to the original SAR products. As a result, this process may only work in combination with loading data from specific collections, not with general data cubes.",
+    "categories": [
+        "cubes",
+        "sar"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "description": "Data cube containing raw backscatter input.",
```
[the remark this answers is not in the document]
jdries: That's one for the experts, @przell  ?

proposals/sar_backscatter.json
m-mohr: A schema could also be something like:

```json
[
  {
    "type": "string",
    "subtype": "collection-id"
  },
  {
    "type": "string",
    "enum": [
      "dem1",
      "dem2"
    ]
  },
  {
    "type": null
  }
]
```

Then a back-end can choose. What to implement, but that may decrease portability. The best (general-purpose) approach would be "load_collection" (subtype: collection-id).

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,78 @@
...
+            "name": "elevation_model",
+            "optional": true,
+            "default": null,
+            "schema": {
+                "type": [
+                    "string",
+                    "null"
+                ]
+            }
+        }
```
soxofaan, line 52: Maybe also add a generic `options`  parameter (like in `save_result`) to allow adding various backscatter tuning parameters that are too implementation dependent to be standardized.

m-mohr: I'm not sure about this. In principle the idea of allowing additional parameters is good. The difference with options in save_result is that the potential values are defined in /file_formats and clients (e.g. Web Editor) can help with filling the fields. Here, there's no equivalent so users might struggle to give actual values. A back-end could add a schema, but from experience that's likely not happening in most cases.

proposals/sar_backscatter.json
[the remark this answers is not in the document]
przell: I think in combination with the possible choices "sigma0" and "gamma0" it is quite clear.

proposals/sar_backscatter.json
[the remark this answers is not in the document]
przell: It depends on the input data we are using. I know for EODC it is S1 GRD_IW. I guess for vito it's the same? 
The SNAP process we are currently implementing can only use S1 GRD_IW. For S1 SLC we would need a different processing chain. Would it make sense to specify this explicitly?
S1 GRD (Ground Range Detected) is already projected to the earths surface into square pixels based on WGS84 ellipsoid, but I wouldn't call it georeferenced yet. The pixel values are the amplitude directly taken from the measurement file. 
Here is the official description of esa: https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/product-types-processing-levels/level-1
To avoid going into details in the process description maybe we could word it like this:
3:  "summary": "Computes backscatter from Sentinel1 Level1 GRD input",
4: "description": "Computes backscatter from Sentinel1 Level1 GRD input. ..."
12: "description": "Data cube containing Sentinel1 Level1 GRD data."

@lforesta @clausmichele @aljacob: Do you think that makes sense?

przell:
Hi all, 
I will start bringing up possible parameters for the **sar_backscatter** process. They will need discussion for sure. But here's a starting point taking into account the CARD4L requirements.
I have taken the CARD4L requirements from here: http://ceos.org/ard/files/PFS/NRB/v5.0/CARD4L-PFS_Normalised_Radar_Backscatter-v5.0.pdf
Unfortunately the page can't be reached at the moment. You can find the list below in tabular form in the ard document on the eodc teams.

**Per Pixel**
These are layers/bands added to the output. To be CARD4L Compliant they should be there. So default should be true.

- data_mask_image: true/false
- scattering_area_image: true/false
- local_incidence_angle: true/false

**Radiometric Terrain Corrected Measurements**
These are very dependent on the software used at the backend. So it will be hard to give generic choices to the user. One solution would be to use the options parameter as proposed by @soxofaan. But every backend would have to implement and document this independently. Otherwise we could fix the choices per backend and only report the choice in the metadata.

- backscatter_measurements: gamma0 (default, sigma0 is not card4l compliant) / sigma0
- scaling_conversion: can we fix this across backends and not give a choice? SNAP: linear, db, dB / Sentinelhub: 10*log10(DN)
- noise_removal: SNAP Themal Noise and Image Border Noise / Sentinelhub [this method](https://sentinel.esa.int/web/sentinel/radiometric-calibration-of-level-1-products)
- radiometric_terrain_correction: SNAP range doppler or SAR sim cross calibration / Sentinelhub [this method](https://www.geo.uzh.ch/microsite/rsl-documents/research/publications/peer-reviewed-articles/201108-TGRS-Small-tcGamma-3809999360/201108-TGRS-Small-tcGamma.pdf). Needs discussion. 

**Geometric Terrain Corrections**

- dem: EODC prefers to use the global 30m dem by default atm. Since for the Copernicus DEM (with 10m resolution in Europe) we would have to go through a licensing procedure. Maybe we could add this after SRR1. @lforesta  please correct me if I have mistaken.
- geometric_accuracy: I guess this should be only reported in the metadata.
- gridding_convention: I will have to look up the exact definition in the requirements again. Will be of interest of how to handle this if the data is reprojected later on in the process graph.

m-mohr:
we will enable it? Basically these are extra bands that are added to the datacube by the ARD process. So the parameter could be (for an atmospheric correction example):
> add_bands: ['AOT', 'WVP']
> (good naming suggestion are more than welcome!)

Could we align that with the STAC naming?

przell:
> These are layers/bands added to the output. To be CARD4L Compliant they should be there. So default should be true.

> So for me, false is a better default.

I agree, from a user perspective this makes absolutely sense! Probably it's enough to say that the process is capable of generating ARD compliant data. But doesn't necessarily do so if the user has other requirements.

m-mohr:
Indeed, but we should make it easy enough to generate CARD4L compliant data. Document it, have an alias process that sets the corresponding parameters or something else.

proposals/atmospheric_correction.json
jdries: Align with sar_backscatter
Keep a string, backend can specialize to make it an enum or collection-id or union of both. (Depending on specific backend.)

jdries: Default schema can use collection id by default?

jdries:
CARD4L: create predefined process graph with parameters configured for compliance.
e.g. card4l_optical = cloud_detection + atmospheric_correction + brdf? (with specific parameters)

proposals/cloud_detection.json
jdries: So: cloud_detection and cloud_mask as two processes.
cloud_detection could return only cloud mask bands?

cloud_mask can be a user defined process, defined in terms of cloud_detection?

What about method specific parameters?
- common denominator
- custom backend-specific object (e.g. context parameter)

proposals/sar_backscatter.json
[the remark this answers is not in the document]
jdries: Proposal to keep it vague:
Computes backscatter from SAR input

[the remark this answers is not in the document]
m-mohr: Done.

proposals/sar_backscatter.json
m-mohr: Done.

proposals/sar_backscatter.json
m-mohr: Removed and replaced with a DigitalElevationModelInvalid exception in two of the processes.

proposals/atmospheric_correction.json
m-mohr: Done.

proposals/cloud_detection.json
m-mohr: Just discussed with Peter: Why do we actually need cloud_mask? It feels like we just can use our regular mask function. Any objections to remove/not have cloud_mask and clouc_detection just return a new data cube with the masks?

m-mohr: Made a new attempt at making cloud_detection more general, more useful and easier to use. It doesn't allow applying the masks anymore, instead we propose to use the mask process. Please see the new version of the process and give feedback. We are pretty sure it's not the last iteration, e.g. we need to discuss which aspects to name (water vapor?) and how to name the dimension labels.

proposals/sar_backscatter.json
m-mohr: @lforesta could you help here?

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,167 @@
...
+            "description": "If set to `true`, a ellipsoidal incidence angle band is added. The values are given in degrees.",
+            "optional": true,
+            "default": false,
+            "schema": {
+                "type": "boolean"
+            }
+        },
+        {
+            "name": "noise_power",
+            "description": "",
```
m-mohr, line 94: Can anyone help what this parameter can be described?

The CARD4L doc says:

> Estimated noise equivalent σ0 (or β0), as applicable) used for Noise Removal, if applied, for each channel.<
> [...] Sample Type [NESZ or NEBZ]

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,167 @@
...
+        {
+            "name": "scaling",
+            "description": "",
+            "optional": true,
+            "default": "dB",
+            "schema": {
+                "type": "string",
+                "enum": [
+                    "dB",
+                    "db",
```
m-mohr, line 128: Can someone explain the difference between db and dB?

proposals/sar_backscatter.json
lforesta: @m-mohr I'm not sure what's the question, if this should be `true` or `false` by default? The S1 ARD data we generate with SNAP/pyroSAR have orthorectification (actually two different topographic corrections are used, maybe @clausmichele can add details), so I would set `true` as default.

m-mohr: @lforesta Thank you. The question is whether this part of the process description "The non-orthorectified products use a simple earth model as provided in the products themselves." is also some kind of orthorectification? Then it would be a bit strange to say "orthorectification" is disabled.

lforesta: Ah I see. If it's disabled, no DEM is used to take into account the topography, and hence I would not call the data orthorectified. Imho it would be geocoding/georeferencing without orthorectification.
What do @clausmichele @aljacob @przell think about this?

przell: I guess orthorectification is called terrain-correction in SNAP. So disabling should be possible also in SNAP. Then it will not be CARD4L compliant though. So in the genereic sar_backscatter process we can keep it as a parameter and in the ard_generation process we will set it to "true".

m-mohr: Thank you. Solved.

proposals/sar_backscatter.json
m-mohr: We've removed it for now

proposals/sar_backscatter.json
m-mohr: We've removed it for now.

proposals/sar_backscatter.json
```diff
@@ -89,15 +89,6 @@
                 "type": "boolean"
             }
         },
-        {
-            "name": "noise_power",
```
przell: It's a target requirement. So not critical. Still we have added all the other target requirements. Here we don't know what this item means. So we removed it for now. 
So we'll have a incomplete list of the target requirements.

m-mohr: On the other hand, CARD4L has no direct connection to sar_backscatter. It's only really relevant to the ard_* processes.

przell: **mask:** create manually via no-data and data
**contributing_area**: double check what it means. required.
**noise_power**: double check what it means. could be left out and added later
**date**: find better naming

 **rtc**: radiometric correction currently fixed to the method chosen by the backend and reported to metadata.

m-mohr: I did the renaming from incidence_angle to local_incidence_angle and from date to date_offset. I'll also apply the changes in STAC.

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,157 @@
...
+                "type": "string",
+                "enum": [
+                    "sigma0",
+                    "gamma0"
+                ]
+            }
+        },
+        {
+            "name": "orthorectify",
+            "description": "By default, orthorectification is disabled. Set to `true` to enable orthorectification.\n\nThe non-orthorectified products use a simple earth model as provided in the products themselves. This may be sufficient for very flat target areas and is faster to process.",
```
soxofaan, line 34: Discription says "By default, orthorectification is disabled" but default value in spec is `true`

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,157 @@
...
+            "description": "If set to `true`, a band is added that provides the acquisition date for each pixel. Values are day offset to the first observation date (in UTC) as specified in the metadata.",
+            "optional": true,
+            "default": false,
+            "schema": {
+                "type": "boolean"
+            }
+        },
+        {
+            "name": "scaling",
+            "description": "By default, all values are given in logarithmic decibel (dB) scale. If set to `linear`, the values are given in linear scale. Values are converted from linear to dB using *dB = 10*log₁₀(l)* and vice-versa.",
```
soxofaan, line 112: `*dB = 10*log₁₀(l)*`

I think the three asterisks here will generate wrong markup. The subscript 10 also looks risky in terms of font/tool support, I think `log10` is just fine. The `l` argument of log10 is also to avoid: it's too close to upper case `i` or `1` in some fonts, e.g. just use `linear` (as you use `dB` at the left hand side).

proposals/ard_surface_reflectance.json
```diff
@@ -0,0 +1,150 @@
...
+                {
+                    "type": "null"
+                }
+            ]
+        },
+        {
+            "description": "The minimum probability required for atmospheric disturbances like clouds for being removed (greater-than-equal).",
+            "name": "threshold",
+            "optional": true,
+            "default": 0.5,
```
jdries, line 39: For ARD, I would dare to go for a lower probability, maybe 0.3. Cloud pixels are very bad for analysis, and the general consensus seems to be that it's better to throw out more good pixels than keeping in clouded pixels. @sinergise-anze  What threshold value do you use to convert cloud probabilities into masks?

proposals/ard_surface_reflectance.json
```diff
@@ -0,0 +1,150 @@
...
+            "schema": {
+                "type": "number",
+                "minimum": 0,
+                "maximum": 1
+            }
+        },
+        {
+            "description": "The name of the band dimension.",
```
jdries, line 47: Not sure if users will understand what this is about. Would it be possible to actually avoid the reduce, and use apply instead? If cloud_detection only returns one band, then why the reduce?

proposals/ard_surface_reflectance.json
```diff
@@ -0,0 +1,150 @@
...
+            "arguments": {
+                "data": {
+                    "from_node": "mask"
+                },
+                "elevation_model": {
+                    "from_parameter": "elevation_model"
+                }
+            }
+        },
+        "merge_cubes": {
```
jdries, line 137: This is a second spatiotemporal join, after the mask. Backend implementors have already indicated that this is one of the most complex processes to implement, so we could consider simplifying this.
One solution is to achieve masking through array operations:
* merge_cubes to add cloudmask band to the cube
* apply that returns an array: [  if (mask <0.5) then band1 else nodata, if (mask <0.5) then band2 else nodata, cloudmask ]

The list of bands to process are basically the band labels of the original L1C cube. Not sure of these are available?

The other option is that openEO backends have a sufficiently smart process graph evaluator, that can figure out that this process graph can be converted into something rather linear, without a need for joins at all. Only for new backend imlementors, this may look a bit daunting.

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,157 @@
...
+            "optional": true,
+            "default": "dB",
```
soxofaan, line 114: Is it a good idea to use dB scaling by default? As far as I know it is mathematically wrong to apply calculations, like aggregation or averaging, directly on dB values. By using `dB` scaling by default we would promote bad practices by default.
However I'm not that familiar with SAR at the moment, so it could be common practice there to do so.

przell: @soxofaan, thanks for pointing this out. I had a deeper look at the CARD4L specification. And you're right. The conversion is not required. 
We can stay with linear values: 

> Note: transformation to the logarithm decibel scale is not required or desired as this step can be easily completed by the user if necessary

For more details look here: 
https://ceos.org/ard/files/PFS/NRB/v5.0/CARD4L-PFS_Normalised_Radar_Backscatter-v5.0.pdf
Point 3.1 and 3.2

We could think about offering the conversion function afterwards. But isn't required for now.

proposals/ard_surface_reflectance.json
przell: Being more strict towards possible disturbances in the designated ARD processes also aligns well with the ARD Concept CEOS stated in their first webinar. Pixel values should represent physically comparable measurements across timesteps, space (and across sensors).

m-mohr: Sounds good. Is 0.3 a "random" value or is there some science behind it? Or is it common denominator between a subset of the tools we use? Then we could maybe link to something and make it look less random (like it was with 0.5, too).

proposals/sar_backscatter.json
m-mohr: Thank you, fixed.

proposals/ard_surface_reflectance.json
m-mohr: We probably need to better document/describe it. By default it's bands and is not required to be set and works if the back-end follows the recommended dimension names. It's an edge case we are handing here or we need to add a constraint that the input data cube has a dimension with name or of type bands. I'm fine with either way.

proposals/ard_surface_reflectance.json
m-mohr: This is actually the simplest of the merges we'd need. The process graph somewhat makes the "algorithm" work, but doesn't output any of the additional bands CARD4L requires us to add. We stopped here to ask whether we actually really want to give the process graph for this process. It is really complicated (and sometimes unknown) where the additional required bands for CARD4L come from and how to compute them...

proposals/sar_backscatter.json
m-mohr: If it's actually not desired, I'd say let's completely remover the parameter and make it linear always. Doing multiply(10, log(x, 10)) is trivial to do in openEO processes.

proposals/sar_backscatter.json
m-mohr: The formula has been removed due to another review comment.

proposals/ard_normalized_radar_backscatter.json
```diff
@@ -0,0 +1,122 @@
...
+            "description": "If set to `false`, no noise removal is applied. Defaults to `true`, which removes noise.",
+            "optional": true,
+            "default": true,
+            "schema": {
+                "type": "boolean"
+            }
+        }
+    ],
+    "returns": {
+        "description": "Backscatter values corresponding to the chosen backscatter coefficient. The data returned is CARD4L compliant and contains metadata.\n\nBy default the backscatter values are given in linear scale. To convert them to decibel scale (dB) use the processes ``multiply()`` and ``log()``.",
```
m-mohr, line 73: I was not sure to which extent to speak about the dB conversion.

We could (1) go the way given above or choose to make it either 

(2) very concrete
```suggestion
        "description": "Backscatter values corresponding to the chosen backscatter coefficient. The data returned is CARD4L compliant and contains metadata.\n\nBy default, the backscatter values are given in linear scale. To convert them to decibel scale (dB) use the processes ``multiply()`` and ``log()``: *dB = multiply(10, log(linear, 10)*.",
```

or (3) not care about it at all and assume it's common knowledge:
```suggestion
        "description": "Backscatter values corresponding to the chosen backscatter coefficient. By default, the values are given in linear scale. The data returned is CARD4L compliant and contains metadata.",
```

proposals/sar_backscatter.json
m-mohr: After discussion with @przell, we thought we may just add it later if required and leave the options out for now.

proposals/ard_surface_reflectance.json
sinergise-anze: I think the value depends a lot on the intended use of the mask. Since the process can't know what the caller expects, maybe it would be a good idea if this was a required parameter without a default value?

proposals/sar_backscatter.json
soxofaan: fair enough.

FYI: in the POC implementation of VITO backend we currently use this `options` argument for example to pass an implementation detail `tile_size`, which allows us to benchmark and finetune performance.

m-mohr: Good to know. We were about to ask whether anyone needs it for SRR1 / has a use case now (then we'd include it now, too).  So if you need it now, we'll add it.

proposals/ard_surface_reflectance.json
jdries: But you have a CLM and CLP band right? I assume CLM is CLP with a threshold applied, or how else do you get it?

jdries: This example mentions 0.4 as a default:
https://github.com/sentinel-hub/sentinel2-cloud-detector/blob/master/examples/sentinel2-cloud-detector-example.ipynb

m-mohr: I agree that if there's no reasonable default then we should make the parameter required.

proposals/sar_backscatter.json
soxofaan: > if you need it ...

We need it in the sense that we currently support it for development/benchmarking reasons. It's not clear if we will need it in the end for normal end users and use cases.

As a backend we can easily add support for this of course. But I also added this `options` parameter to the `sar_backscatter` helper (flagged experimental) in the python client: https://github.com/Open-EO/openeo-python-client/blob/daa440f57ec50ef37f09abd68ab409fa18a3fb5d/openeo/rest/datacube.py#L1494-L1497

m-mohr: Okay, I meant needed in a (user-oriented) use case. So for internal reasons indeed you are fine to just add it to the process anyway.

soxofaan: and I forgot, we currently also have a `to_dB` option in there too, comparable to the `scaling` argument which is still under discussion

proposals/ard_surface_reflectance.json
sinergise-anze: > But you have a CLM and CLP band right? I assume CLM is CLP with a threshold applied, or how else do you get it?

Sorry, didn't understand what you meant - yes, CLM uses a predefined threshold value of `0.4`.

m-mohr:
We have another preliminary draft up, please all have a look.
The SAR part should be near final and the Optical will need some additional fine-tuning.

proposals/ard_surface_reflectance.json
m-mohr: Removed for now.

proposals/ard_surface_reflectance.json
m-mohr: Removed for now. With the process graph removed it's not required any longer and the description just says a single dimension of type bands is required.

proposals/ard_surface_reflectance.json
m-mohr: We've removed probability completely for now as we got into discussions whether the probability actually has to be per "aspect" (e.g. cloud or shadows) separately. It seems that's more useful, but on the other hand, then we need an object that maps the probability values to the respective aspect and that is complicated. Thus, we decided, for now, to remove it completely and just use the defaults of the chosen method.

proposals/sar_backscatter.json
m-mohr: @przell Is this all resolved?

proposals/sar_backscatter.json
m-mohr: Scaling has been removed in favor of just computing dB yourself using apply, multiply and log.

proposals/ard_normalized_radar_backscatter.json
m-mohr: I forgot that it also needs `apply`, of course. I've removed the dB wording for now, hoping that people know what to do to convert from linear to dB

lforesta (approved):

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,127 @@
+{
+    "id": "sar_backscatter",
+    "summary": "Computes backscatter from SAR input",
+    "description": "Computes backscatter from SAR input.\n\nNote that backscatter computation may require instrument specific metadata that is tightly coupled to the original SAR products. As a result, this process may only work in combination with loading data from specific collections, not with general data cubes.",
```
jdries, line 4: This description does not yet mention the mapping of the parameters to sigma0 and gamma0. We can no longer specify these directly, so I believe it would be convenient for the user, and the backend implementor, to at least document what set of parameters corresponds to sigma0 and gamma0.

m-mohr: Can someone provide the mapping? I was overwhelmed by the options in the call and can't recall how the mapping is meant to be. @aljacob

proposals/sar_backscatter.json
```diff
@@ -0,0 +1,127 @@
...
+        {
+            "name": "data",
+            "description": "The source data cube containing SAR input.",
+            "schema": {
+                "subtype": "raster-cube",
+                "type": "object"
+            }
+        },
+        {
+            "name": "elevation_model",
```
soxofaan, line 20: Isn't it cleaner to move parameter `elevation_model` after `orthorectify` and `rtc`?

m-mohr: Depends a bit how often we think the DEM will be chosen in contract to the other parameters. I assumed choosing a DEM would be a "best-practice", then it's good where it is. But my assumption could be wrong.
Another reason was to align with the other processes in this PR. For all of them which have such an option, it's the first optional parameter.

soxofaan: As far as I currently understand `elevation_model` only makes sense when `orthorectify` and/or `rtc` are enabled, so it looks more intuitive if that parameter comes after these.
Also, elevation_model has a default of null to let backend pick a model, so it is less critical to specify than `orthorectify` or `rtc`

m-mohr: That sounds reasonable.

proposals/ard_surface_reflectance.json
```diff
@@ -0,0 +1,77 @@
...
+            "description": "The cloud detection method to use.",
+            "name": "cloud_detection_method",
+            "schema": {
+                "type": [
+                    "string"
+                ],
+                "enum": [
+                    "sen2cor",
+                    "fmask",
+                    "s2cloudless"
```
jdries, line 43: force is missing?

proposals/ard_surface_reflectance.json
```diff
@@ -0,0 +1,77 @@
...
+            "description": "The source data cube containing multi-spectral optical top of the atmosphere (TOA) reflectances. There must be a single dimension of type bands available.",
+            "name": "data",
+            "schema": {
+                "subtype": "raster-cube",
+                "type": "object"
+            }
+        },
+        {
+            "description": "The atmospheric correction method to use.",
+            "name": "athmosperic_correction_method",
```
jdries, line 22: typo: athmosperic -> atmospheric

proposals/ard_surface_reflectance.json
m-mohr: Not sure, I have not compiled this list of options. If you can replace the "?" with a "!" I'll add it ;-)

jdries: @lforesta  I guess you can confirm this?

proposals/sar_backscatter.json
m-mohr: Moved behind the two options.

proposals/ard_surface_reflectance.json
m-mohr: I've just added it for now. If @lforesta can't confirm, we'll remove it again.
I also change the casing of the options to conform to the official casing (e.g. "Sen2Cor" instead of "sen2cor", "iCOR" instead of "iCor" etc)

m-mohr: I've also added FORCE to cloud_detection, although it's likely not yet possible to use it standalone?!

lforesta: FORCE uses fmask for cloud detection (see https://force-eo.readthedocs.io/en/latest/howto/l2-ard.html#tut-ard , point 5). I have not seen a way to use this independently from the utility that generates ARD.
It must be possible to isolate the function that is used internally for this, but it would require quite some extra work I guess.

m-mohr: I'll remove FORCE again, thank you.

lforesta (approved):

m-mohr:
Merging for now after we had the telco today, any additional changes may be discussed in a separate PR.

PR #215, opened 2020-12-21
Adds return value schemas for child processes

m-mohr:
Specifying the return values processes receive from child processes.

This adapts to a recent API change proposed in https://github.com/Open-EO/openeo-api/pull/351

The corresponding API issue is https://github.com/Open-EO/openeo-api/issues/350

A small addition is that is also clarifies the behavior for data cubes returned by the child processes and for that add the exception `DataCubePropertiesImmutable`.


aggregate_spatial.json
```diff
@@ -49,7 +49,13 @@
                         "optional": true,
                         "default": null
                     }
-                ]
+                ],
+                "returns": {
+                    "description": "Return the value that should be stored in the data cube.",
+                    "schema": {
+                        "description": "Any data type."
```
soxofaan, line 56: The "any" here looks like you can do anything, but shouldn't we at least put in the description that the output data type should be related to the input (numerical raster cube in -> numerical out)?

soxofaan: (these comments on the `aggregate_spatial` diff probably relate to all other diffs below too)

array_apply.json
```diff
@@ -63,7 +63,13 @@
                         "optional": true,
                         "default": null
                     }
-                ]
+                ],
+                "returns": {
+                    "description": "Return the value that should be stored in the data cube.",
```
soxofaan, line 68: > data cube

array_apply works on and returns array

aggregate_spatial.json
```diff
@@ -49,7 +49,13 @@
                         "optional": true,
                         "default": null
                     }
-                ]
+                ],
+                "returns": {
+                    "description": "Return the value that should be stored in the data cube.",
```
soxofaan, line 54: > Return the value that should be stored in the data cube.

The "Return" verb at the beginning seems a bit out of place to me, I think the description works without it.

And maybe for clarity: "stored in the output data cube" (or "resulting data cube")

aggregate_spatial.json
m-mohr: Well, if the back-end supports it, you can do anything. Usually you don't want to change the data type, but that's nothing I could enforce here.

array_apply.json
m-mohr: Good catch, will be fixed in the next commit.

aggregate_spatial.json
m-mohr: Thanks, indeed this can be written a bit more compact. Will be fixed in the next commit.

It seems we have not used anything like "output ..." or "resulting ..." yet in the processes. In the "returns" of the process, it also just says "A vector data cube". So there's not really a consistent phrase we can "refer" to. In array_apply it says "An array with the newly computed values.", which also doesn't make a short referable phrase. On the other hand, I'm not even sure the clarification is really needed.

aggregate_spatial.json
soxofaan: I wasn't talking about enforcing through `schema`, because that is probably very hard indeed. I just thought about a bit of clarification through `description`, e.g. that for most use cases the return type should be scalar instead of (multi)array or cube.

aggregate_spatial.json
soxofaan: yeah, maybe what triggered me is the verb "store" as well. "Store in the data cube" somewhat sounds as in-place replacement to me for some reason. While more generic, "use" is less suggestive maybe?

aggregate_spatial.json
m-mohr: That feels like something that should be clarified in general (e.g. in the glossary, although I'd like to have a general "working with data cubes" section in the documentation at some point) and per back-end. Some back-ends only support numbers for example, some may allow for example numbers -> strings or boolean -> numbers. Maybe there should be a place for each back-end to document what restrictions on working with data cubes are available?

aggregate_spatial.json
m-mohr: Indeed, store sounds like an in-place replacement.
I'm not exactly sure why, but "The value to be used in the data cube/array" sounds strange. "use" seems to generic and it could be unclear what that actually means. Would "The value to be stored in a new data cube/array" work? That would be somewhat aligned with the process return value description ("A data cube with the newly computed values.") in most cases.

m-mohr: Is commit https://github.com/Open-EO/openeo-processes/pull/215/commits/ec5f84585bdf8c61df66806d5e8548c04aa6509b fine for you? I'm pretty happy with the improvement.

soxofaan: yes that's already better

another issue with "store" is that it also sound like "persist or "store to disk" (like `save_result`), which is not what you to suggest I guess. 

I had a quick search on thesaurus.com and kind of prefer these alternatives: "gather", "assemble", "collect"

aggregate_spatial.json
soxofaan: I don't mind "usually", it allows to describe a common use case in more concrete terms while still allowing going beyond that. 
As developer I would prefer that over a reference to some part of the glossary to be honest. 

> Maybe there should be a place for each back-end to document what restrictions on working with data cubes are available

I can only speak for VITO backend, but concerning aggregate/reduce it feels overkill at this point to discuss exotic extensions on top of what we already support (number -> number).

aggregate_spatial.json
m-mohr: I think store (here: in memory) is fine for me and it should be clear from the context, but a simple alternative would be "set", I think.

aggregate_spatial.json
m-mohr: Maybe that is a mis-understanding, but it became obvious several times in openEO history that people often are not aware of how data cubes and their methods (reduce, apply, ...) work. So I was thinking that a guide how to work with them in some use cases would help the understanding. That could document that it's usually not a good idea to change data types and for all others that should usually be clear anyway. The thing is that this is a back-end specific thing what is allowed and thus it feels not the right place to put it all over the processes... If you only support numbers anyway, then it would actually be better to change the process definition and enforce the "number" type and now allow "any". That would then be very specific and can be read by clients, too.

soxofaan: well, a fair amount of repetition/duplication in documentation can be beneficial, so it shouldn't be "one `xor` the other", it can be explained in both places: in depth in the glossary and succinctly in a process description I think.

> to change your process definitions and enforce the "number" type and now allow "any". That would then be very specific and can be read by clients, too.

While a backend can indeed "overwrite" a spec (and clients can attempt follow that), I would not underestimate the gravity of references like https://openeo.org/documentation/1.0/processes.html however: people will end up there when googling for documentation and be confused when a backend does things differently.

m-mohr: Sure that are all valid issues, although the exact definition will be given by /processes to clients and can be visualized in the same workflow. That's hopefully where people look first as they are in those environments anyway, but surely they can also google and find the general docs. 

It seems there are multiple potential ToDos:
1. Clarify on processes.openeo.org etc that these are just the general specifications and back-ends can derive from it. So users should be redirected to the clients to figure out what processes support. -> opened https://github.com/Open-EO/openeo.org/issues/27
2. Write the data cube guide with the notes mentioned -> opened https://github.com/Open-EO/openeo.org/issues/26
3. Check what needs to be added in the processes itself. -> opened https://github.com/Open-EO/openeo-processes/issues/216

I've opened issues for all of them so that we can keep track so that I can merge this.

aggregate_spatial.json
m-mohr: @soxofaan I've changed most occurances of 'store' to set, used or added, see https://github.com/Open-EO/openeo-processes/pull/215/commits/8be67ff27b1b29933d4df683ffec97735ade052e

aggregate_spatial.json
soxofaan: > given by /processes to clients and can be visualized in the same workflow. That's hopefully where people look first as they are in those environments anyway

I'm afraid this is your webeditor/js bias speaking here :smile: 
In the webeditor it is natural by design to visualize this kind of  "dynamic" documentation, because you have to be logged in on a backend before you can start "programming".

Workflows with the Python client (and I guess R client too) are generally more disconnected from a backend during development, so you can not work with "dynamic" documentation. You could say that (jupyter) notebooks enable webeditor-like environments allowing "dynamic" documentation and discovery, but I would still bet by money on most end users opening a disconnected browser tab to look up for information.

array_find.json
```diff
@@ -26,7 +26,7 @@
         }
     ],
     "returns": {
-        "description": "Returns the index of the first element with the specified value. If no element was found, `null` is returned.",
+        "description": "The index of the first element with the specified value. If no element was found, `null` is a.",
```
soxofaan, line 29: > is a.

typo?

soxofaan (approved):

aggregate_spatial.json
m-mohr: The R client using R Studio can visualize the same "user interface" as it's available on processes.openeo.org and it will also be available in Jupyter. All guides teach you to first connect, so I doubt many people will work "offline" without connecting first?So, it's not only my Web Editor bias here, but there's also truth in your comment for sure (e.g. working in a non-IDE env or opening browser tabs). :-) What I'd like to see from back-ends at some point is to make their own "processes.provider.com" page rendering their processes with the docgen so that people can use that easily...

array_find.json
m-mohr: How on earth did that get in here? it should still be `returned`. Thanks for spotting this.

Issue #212, opened 2020-11-27
merge_cubes: order of dimension labels after merge


m-mohr:
merge_cubes doesn't specify directly which order the dimension labels have been after the merge. It's implicitly clear that temporal and spatial likely have an order, but for band names for example it's not clear whether the bands are just concatenated or - for example - sorted. Behavior probably depends on the ref sys and if there's none than it concatenation, but it should be clarified.

Example 1:
- temporal dimension labels cube1: 2018-..., 2020-...
- temporal dimension labels cube2: 2019-...
- I'd guess the result is 2018-..., 2019-..., 2020-...

Example 2:
- band dimension labels cube1: B1, B4
- band dimension labels cube2: B2, B5
- I'd guess the result is B1, B4, B2, B5, but others could think it's B1, B2, B4, B5 as for spatial and temporal it's also having an implicit order

Opinions? How is it implemented?

soxofaan:
I don't have a strong opinion which one feels best. For spatial and temporal it seems most logical to auto-sort. For band/other dimensions: is it that big of an issue to leave it as an implementation detail? 

Side question: can labels in a  dimensions have mixed type (number and string)? If yes: how to define order for mix of number and strings? If no: what with merge_cube where first cube has number labels and second cube has string labels?

lforesta:
@soxofaan it's not an implementation detail because `array_element` allows to get an element by index, so it has to be clear for the user if it's B1, B2, B4, B5 or B1, B4, B2, B5.

In our case it is sorted in the temporal dimension, but not on the spectral one, but indeed the user cannot know that, so I'm in favour of specifying that in the process definition.

As to what specify, I don't have a strong opinion. In principle we have `sort` to sort a dimension, but it feels clunky to leave it to the user to having to use sort on every dimension, after having used `merge_cubes`...
I guess I would expect the 'continuous' dimensions to be sorted in ascending order (this includes spectral if you think of the wavelenght), and categorical ones to be in alphabetical order.

m-mohr:
I have written the processes with the following expectations in mind:

- spatial and temporal have a natural order that is always in place
- bands (not necessarily spectral) are ordered as specified in metadata or processes, e.g. load_collection and filter_bands explicitly state that the order of the given array matters
- all other dimension types are also ordered as specified in metadata or processes
- For both bands and other dimension types, new labels are appended, e.g. see my merge_cubes example above

In theory dimension labels could be mixed types, yes.

[...]

PR #219, opened 2021-01-07
merge_cubes: order of dimension labels after merge

m-mohr:
This clarifies what has been discussed in #212. In addition, I updated the openEO glossary, too. See commit https://github.com/Open-EO/openeo.org/commit/5764eb8f740270f32a49863db16795dd1cf6435f


merge_cubes.json
```diff
@@ -1,7 +1,7 @@
...
...cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.\n   2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.\n4. Merging a data cube with dimensions `x`, `y`, `t` with another cube with dimensions `x`, `y` will join on the `x`, `y` dimension, so the lower dimension cube is merged with each time step in the higher dimensional cube. This can for instance be used to apply a digital elevation model to a spatiotemporal data cube.\n\nAfter the merge, the dimensions with a natural/inherit label order (with a reference system this is each spatial and temporal dimensions) still have all dimension labels sorted. For other dimensions where there is no inherit order, including bands, the dimension labels keep the order in which they are present in the original data cubes and the dimension labels of `cube2` are appended to the dimension labels of `cube1`.",
```
m-mohr, line 4: @soxofaan I took over your wording from the issue but did you mean to use 'inherit' or 'inherent'?

```suggestion
    "description": "The data cubes have to be compatible. A merge operation without overlap should be reversible with (a set of) filter operations for each of the two cubes. The process performs the join on overlapping dimensions, with the same name and type.\n\nAn overlapping dimension has the same name, type, reference system and resolution in both dimensions, but can have different labels. One of the dimensions can have different labels, for all other dimensions the labels must be equal. If data overlaps, the parameter `overlap_resolver` must be specified to resolve the overlap.\n\n**Examples for merging two data cubes:**\n\n1. Data cubes with the dimensions `x`, `y`, `t` and `bands` have the same dimension labels in `x`,`y` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the first cube and `B3` and `B4`. An overlap resolver is *not needed*. The merged data cube has the dimensions `x`, `y`, `t` and `bands` and the dimension `bands` has four dimension labels: `B1`, `B2`, `B3`, `B4`.\n2. Data cubes with the dimensions `x`, `y`, `t` and `bands` have the same dimension labels in `x`,`y` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the first data cube and `B2` and `B3` for the second. An overlap resolver is *required* to resolve overlap in band `B2`. The merged data cube has the dimensions `x`, `y`, `t` and `bands` and the dimension `bands` has three dimension labels: `B1`, `B2`, `B3`.\n3. Data cubes with the dimensions `x`, `y` and `t` have the same dimension labels in `x`,`y` and `t`. There are two options:\n   1. Keep the overlapping values separately in the merged data cube: An overlap resolver is *not needed*, but for each data cube you need to add a new dimension using ``add_dimension()``. The new dimensions must be equal, except that the labels for the new dimensions must differ by name. The merged data cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.\n   2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all pixels. The merged data cube has the same dimensions and labels as the original data cubes, but all pixel values have been processed by the overlap resolver.\n4. Merging a data cube with dimensions `x`, `y`, `t` with another cube with dimensions `x`, `y` will join on the `x`, `y` dimension, so the lower dimension cube is merged with each time step in the higher dimensional cube. This can for instance be used to apply a digital elevation model to a spatiotemporal data cube.\n\nAfter the merge, the dimensions with a natural/inherent label order (with a reference system this is each spatial and temporal dimensions) still have all dimension labels sorted. For other dimensions where there is no inherent order, including bands, the dimension labels keep the order in which they are present in the original data cubes and the dimension labels of `cube2` are appended to the dimension labels of `cube1`.",
```

soxofaan: I indeed meant "inherent"

soxofaan (approved):

PR #221, opened 2021-01-27
proposal for resolution merge process

jdries:
Resolution merging can be considered part of ARD, and is supported by the FORCE framework:
https://force-eo.readthedocs.io/en/latest/components/higher-level/l2i/index.html#l2i

The concrete implementation at VITO uses the FORCE approach to increase the resolution of Sentinel-2 20M to 10M.

I named it 'resolution_merge', because 'pansharpening' refers to the usage of a panchromatic band, which is not the case in the Sentinel-2 algorithm. Resolution merging also seems to be used in publications and by force, although sometimes you also see references to pansharpening in that context.

The usage with the Python client is shown here:
https://github.com/Open-EO/openeo-geopyspark-integrationtests/blob/master/tests/test_integration.py#L934

The core of the backend implementation is here:
https://github.com/Open-EO/openeo-geotrellis-extensions/blob/develop/openeo-geotrellis/src/main/scala/org/openeo/geotrellis/ard/Improphe.scala#L62

### DIFF
--- a/proposals/resolution_merge.json
+++ b/proposals/resolution_merge.json
@@ -1,8 +1,7 @@
 {
     "id": "resolution_merge",
-    "summary": "Resolution merging or pansharpening: increase the spatial resolution of low-resolution bands, based on higher resolution bands.",
-    "description": "Resolution merging algorithms try to improve the spatial resolution of lower resolution bands (e.g. Sentinel-2 20M) based on higher resolution bands. (e.g. Sentinel-2 10M).\n ",
-
+    "summary": "Resolution merging / pan-sharpening",
+    "description": "Resolution merging algorithms try to improve the spatial resolution of lower resolution bands (e.g. Sentinel-2 20M) based on higher resolution bands (e.g. Sentinel-2 10M).",
     "categories": [
         "cubes",
         "optical"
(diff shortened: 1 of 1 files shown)


proposals/resolution_merge.json
```diff
@@ -0,0 +1,73 @@
...
+            "schema": {
+                "type": "array",
+                "items": {
+                    "type": "string",
+                    "subtype": "band-name"
+                }
+            }
+        },
+        {
+            "name": "method",
```
m-mohr, line 42: @przell  and I have removed the method parameter in #210. I guess we should align here or there.

proposals/resolution_merge.json
```diff
@@ -0,0 +1,73 @@
...
+    "description": "Resolution merging algorithms try to improve the spatial resolution of lower resolution bands (e.g. Sentinel-2 20M) based on higher resolution bands (e.g. Sentinel-2 10M).",
+    "categories": [
+        "cubes",
+        "optical"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "Data cube containing multiple spectral bands, with different source resolutions.",
```
m-mohr, line 13: Is this actually possible? It seems like our current data cube model doesn't allow different resolution bands in a single data cube (i.e. resolution is part of the spectral dimension, not the bands)

m-mohr (commented):
@jdries I made a couple of changes (textual, validity), which you can see in commit https://github.com/Open-EO/openeo-processes/pull/221/commits/aec7cb34cc033602974fddc42052ef178af007da - I hope they are all fine for you. I've also added some comments here for things that we should discuss.

proposals/resolution_merge.json
jdries: Indeed, my current implementation actually already upsamples 20m to 10m resolution when loading the data. This process then applies a 'smarter' resolution_merge algorithm.
But the statement that the source resolutions differ is actually true, it's not literally saying that the bands have different resolutions.
So not sure what to do with this in fact...

m-mohr: I think my point is that you can't have different resolution bands in reality in our data cubes and thus this process doesn't work as specified here, right? You'd probably need to load two different data cubes with different resolutions and then resolution merge the bands in those data cubes, I guess...

jdries: Or the same solution as for multiple UTM zones: using additional dimensions to resolve the issue.
Basically what this process does is resolving the issue of loading bands with different resolutions. Backends already do this in fact, but usually the method is basic upsampling. What we do here is simply allowing another method to be specified.

A solution with a merge feels again like pushing more complexity to both users and backends.

m-mohr: > Or the same solution as for multiple UTM zones: using additional dimensions to resolve the issue.

Yes, that could be an alternative, I guess...

> Basically what this process does is resolving the issue of loading bands with different resolutions. Backends already do this in fact, but usually the method is basic upsampling. What we do here is simply allowing another method to be specified.

The issue is that it needs some intelligent logic in the process graph parsing, but those side effects are not clear and make the process not very universal. So then the result of the process depends on the underlying "smartness" of the back-end... it would be better if users could actually understand what happens on a step-by-step basis.

m-mohr:
Closing as stale, we can re-open whenever this gets some attention again.

PR #224, opened 2021-03-10
CARD4L: Don't require contributing area by default #223

### LINKED ISSUE #223: ard_normalized_radar_backscatter: do not require contributing_area?
The current definition of ard_normalized_radar_backscatter sets contributing_area to true by default.
However, CEOS says:

> This is also something the WG is looking to change. Apparently SNAP is not the only tool that doesn’t provide this output. 

And neither SNAP nor Sentinelhub support it. So we do not have any software that can actually produce it. Hence, a default invocation of the process will either fail, or not comply with the process spec.

Should we already change the default, in anticipation of a change made by CEOS to CARD4L? Other option is to wait, but we certainly have to address it before making the process final.

m-mohr:
Don't require contributing area by default as most tools don't support it and also CEOS is likely to remove it from the spec.

Closes #223

### DIFF
--- a/proposals/ard_normalized_radar_backscatter.json
+++ b/proposals/ard_normalized_radar_backscatter.json
@@ -33,6 +33,15 @@
             ]
         },
         {
+            "name": "contributing_area",
+            "description": "If set to `true`, a DEM-based local contributing area band named `contributing_area` is added. The values are given in square meters.",
+            "optional": true,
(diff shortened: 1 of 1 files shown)


jdries (approved):
Approved!

m-mohr:
Will merge @ 16:00 today.

Issue #228, opened 2021-04-02
Documentation of (CARD4L) ARD processes


jdries:
Based on review remark, which mentions that the documentation is rather sparse.
- ard_normalized_radar_backscatter does not mention the bands that are added by default, and their names
- ard_surface_reflectance : CARD4L specifies metadata bands to be added, but the process doesn't mention any of this

In general: maybe it can link to code samples? For instance:
https://open-eo.github.io/openeo-python-client/ard.html#atmospheric-correction
We point to processes.openeo.org a lot, which in it current form never mentions how to actually run things. 

Can we also extend the description, perhaps by even copying from the CARD4L description, so people don't have to go and read the PDF?

m-mohr:
Yeah, we did not copy the information over from sar_backscatter for example. There it is defined how the bands will be named etc. We can simply copy that over.

It's the same for surface reflectance, many of the missing information is in the underlying processes for atmospheric_correction and cloud_detection. Here it's not 100% clear though what is returned as we don't know to what extend different cloud masks are available.

m-mohr:
@jdries What do you want me to add from the CARD4L spec?

Linking to the library docs: I'm a bit hesitant as things are still in flux, in theory, processes are independent of implementations in clients and we should probably discuss a structure to not just make this yet another "about" link.

I've opened PR #229 for improvements. For now just SAR and I'll work on SR next. Is SAR how you expected it to be?

m-mohr:
> ard_surface_reflectance : CARD4L specifies metadata bands to be added, but the process doesn't mention any of this

@jdries Could you please point me to where this is specified? I can't find it in the CARD4L spec.

jdries:
I believe this is the full list:
2.2 No Data -> threshold
2.3 Incomplete Testing  -> threshold
2.4 Saturation  -> threshold
2.5 Cloud  -> threshold
2.6 Cloud Shadow  -> threshold
2.7 Land/Water Mask
2.8 Snow/Ice Mask
2.9 Terrain Shadow Mask
2.10 Terrain Occlusion
2.11 Solar and Viewing Geometry -> threshold: provide average
2.12 Terrain Illumination Correction
2.13 Aerosol Optical Depth Parameters

So looking at this, it the cloud bitmask seems to be the most relevant.

m-mohr:
@jdries @lforesta What bands names have you implemented for all these? Otherwise I'll just use the names that correspond to STAC CARD4L.

[...]

PR #229, opened 2021-04-02
Better documentation of (CARD4L) ARD processes #228

m-mohr:
Fixes for #228:
- ard_normalized_radar_backscatter does not mention the bands that are added by default, and their names
- ard_surface_reflectance : CARD4L specifies metadata bands to be added, but the process doesn't mention any of this
- Minor clarifications and improvements to wording


m-mohr:
ToDo: Add `options` parameter to new processes that back-ends can extend it based on methods. Will likely also add implementation notes...

proposals/ard_normalized_radar_backscatter.json
```diff
@@ -61,7 +61,7 @@
         }
     ],
     "returns": {
-        "description": "Backscatter values expressed as gamma0. The data returned is CARD4L compliant and contains metadata.\n\nBy default, the backscatter values are given in linear scale.",
+        "description": "Backscatter values expressed as gamma0 in linear scale.\n\nIn addition to the bands `contributing_area` and `ellipsoid_incidence_angle` that can optionally be added with corresponding parameters, the following bands are always added to the data cube:\n\n- `mask`: A data mask that indicates which values are valid (1), invalid (0) or contain no-data (null).\n- `local_incidence_angle`: A band with DEM-based local incidence angles in degrees.\n\nThe data returned is CARD4L compliant with corresponding metadata.",
```
lforesta, line 64: local_incidence_angle -> @clausmichele you mentioned that SNAP/pyroSAR can produce this as a separate output, do I remember this correctly?

proposals/ard_surface_reflectance.json
```diff
@@ -61,7 +61,7 @@
...
...ssessed as being snow/ice (1) or not (0). All values describes the probability and must be between 0 and 1.\n- `land-water` (optional): Indicates whether a pixel is assessed as being land (1) or water (0).  All values describes the probability and must be between 0 and 1.\n- `incidence-angle` (optional): Specifies per-pixel incidence angles in degrees.\n- `azimuth` (optional): Specifies per-pixel azimuth angles in degrees.\n- `sun-azimuth:` (optional): Specifies per-pixel sun azimuth angles in degrees.\n- `sun-elevation` (optional): Specifies per-pixel sun elevation angles in degrees.\n- `terrain-shadow` (optional): Indicates with a value of 1 whether a pixel is not directly illuminated due to terrain shadowing. Otherwise the value is 0.\n- `terrain-occlusion`  (optional): Indicates with a value of 1 whether a pixel is not visible to the sensor due to terrain occlusion during off-nadir viewing. Otherwise the value is 0.\n- `terrain-illumination`  (optional): Contains coefficients used for terrain illumination correction are provided for each pixel.\n\nThe data returned is CARD4L compliant and contains metadata.\n\nThe data returned is CARD4L compliant with corresponding metadata.",
```
lforesta, line 64: per-pixel tests -> is this left very vague on-purpose? Clearly it is method/algorithm dependent. In case one method provides no suh tests, should a back-end return all 0 by default?

m-mohr: This is simply copy/pasted from CARD4L. As far as I understand, a method that doesn't support such tests can't be CARD4L compliant? But maybe it's also just all 0. 🤷‍♀️

proposals/ard_normalized_radar_backscatter.json
clausmichele: yes, see here https://pyrosar.readthedocs.io/en/latest/pyroSAR.html?highlight=localIncidenceAngle#pyroSAR.snap.util.geocode

proposals/ard_surface_reflectance.json
m-mohr: I've changed it to be a bit more precise: 

> Identifies pixels with a value of 1 for which the per-pixel tests (at least saturation, cloud and cloud shadows, see CARD4L specification for details) have not all been successfully completed. Otherwise the value is 0.

m-mohr:
I've added the `options` parameters for Optical SR and also started an "implementation guide" for back-ends, which we can fill with details about implementing processes in the next time. @lforesta @jdries

proposals/ard_surface_reflectance.json
lforesta: ah ok, I wasn't sure those were the "tests"

m-mohr: I'm still not sure, but that's my understanding of the CARD4L spec.

meta/implementation.md
```diff
@@ -0,0 +1,76 @@
...
+
+```json
+{
+	"description": "Proprietary options for the atmospheric correction method.",
+	"name": "options",
+	"optional": true,
+	"default": {},
+	"schema": [
+		{
+			"title": "FORCE options",
```
lforesta, line 50: Should the properties' names be unique? Or are they valid within their scope/object?
e.g. imagine a back-end supports both iCor and FORCE and both have a property `erase_cloud`

m-mohr: Don't need to be unique. It's clear from the method chosen which options apply.

aljacob (commented):
fine for me

m-mohr:
Merged as discussed in the dev telco. We can continue working on improvements in additional PRs, but it seems most additional changes will likely need to be done in the specific process definitions by the back-ends.

PR #232, opened 2021-04-13
Linear interpolation for arrays

m-mohr:
Related issue: #173

Implements a linear interpolation process that replaces null values in arrays.
Hope this is close to what is required for Platform UC3. It is meant to be used with apply_dimension in most cases.
Feedback would be appreciated.
Inspired by [xarray.interpolate_na](xarray.DataArray.interpolate_na).


proposals/array_interpolate_linear.json
```diff
@@ -0,0 +1,57 @@
+{
+    "id": "array_interpolate_linear",
+    "summary": "One-dimensional linear interpolation for arrays",
+    "description": "Performs a linear interpolation for each of the no-data values (`null`) in the array given, except for the first and last value.",
```
soxofaan, line 4: > except for the first and last value.

except for leading and trailing no-data values ?

soxofaan: Another option is to provide the `fill_value` argument like in `xarray.interpolate_na` or `scipy.interpolate.interp1d`

```python
>>> nan = float("NaN")
>>> data = xarray.DataArray([nan, 1, nan, 2, nan, 10, nan, nan])

>>> data.interpolate_na("dim_0").data
array([ nan,  1. ,  1.5,  2. ,  6. , 10. ,  nan,  nan])

>>> data.interpolate_na("dim_0", fill_value=-1).data
array([-1. ,  1. ,  1.5,  2. ,  6. , 10. , -1. , -1. ])

>>> data.interpolate_na("dim_0", fill_value="extrapolate").data
array([ 0.5,  1. ,  1.5,  2. ,  6. , 10. , 14. , 18. ])
```

proposals/array_interpolate_linear.json
```diff
@@ -0,0 +1,57 @@
...
+    "categories": [
+        "arrays",
+        "math",
+        "math > interpolation"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "An array of numbers and no-data values.\n\nIf the given array is a labeled array, the labels must have a natural/inherent label order and it expects the labels to be sorted accordingly. This is the default behavior in openEO for spatial and temporal dimensions.",
```
soxofaan, line 14: > and it expects the labels to be sorted accordingly

Why is this necessary? It would be more user friendly to make the sorting  the responsibility of the backend implementation, I think

soxofaan: I think there should be also explanation that not only the order of labels will be used, but also the value (if numerical).
e.g. that `interpolate({0: 10, 2: NaN, 10: 20})` results in `{0: 10, 2: 12, 10: 20}` and not `{0: 10, 2: 15, 10: 20}`

proposals/array_interpolate_linear.json
m-mohr: I had something similar in the first commit (first/last parameters). That seemed to make the process definition much more complex so I decided to start with a minimum and add features for additional requirements once they come up.

proposals/array_interpolate_linear.json
m-mohr: > Why is this necessary? It would be more user friendly to make the sorting the responsibility of the backend implementation, I think

This is meant to be mostly a back-end responsibility. Spatial and temporal dimensions are sorted by default (temporal will likely be the most common use-case), all other dimensions are usually not usable for interpolation or must be sorted though. The reason is that the process on its own just from the array can't really know how to sort properly.

> I think there should be also explanation that not only the order of labels will be used, but also the value (if numerical).
> e.g. that `interpolate({0: 10, 2: NaN, 10: 20})` results in `{0: 10, 2: 12, 10: 20}` and not `{0: 10, 2: 15, 10: 20}`

Yes, I agree. It should be implicitly clear that the label values are used, but it can be made explicit. Unfortunately, I currently have no good idea how to describe that (without just providing an example).

m-mohr: Maybe something like the folloowing?

> The linear interpolants are defined by the dimension labels (x) and the values in the array (y).

m-mohr: See commit https://github.com/Open-EO/openeo-processes/pull/232/commits/7127bbe110b1faaf48c77c75020f178915a62768 for an attempt at solving your issues.

proposals/array_interpolate_linear.json
soxofaan: a problem with this basic spec might be that the user expects that all NaN are (or can be) replaced by `array_interpolate_linear`, regardless of their location. But maybe I'm overthinking it

proposals/array_interpolate_linear.json
soxofaan: > This is meant to be mostly a back-end responsibility

the "it expects the labels to be sorted accordingly" somewhat reads like it is responsibility of the user (because of the "expects"). Grammatical side note: what does the "it" refer to in that sentence?

I think if we can explain the x-y-interpolation details properly (for non-labeled and labeled arrays), we don't have to talk about label order in the first place.

proposals/array_interpolate_linear.json
m-mohr: Let's wait for the first implementation / use-case and then try to figure out what to add. :-)

proposals/array_interpolate_linear.json
m-mohr: it = the process, will fix

Yes, it's the user responsibility, but I added the "This is the default behavior in openEO for spatial and temporal dimensions.", which was meant to make clear that for spatial and temporal dimensions usually no further action is required.

m-mohr:
@jdries You mentioned you implemented interpolation. Do you have any feedback for this PR?

jdries:
For now, it seems fine, but I didn't do any deep testing yet.
I also still do not support labels, so to be seen how that works out.
From my side, it would be fine to merge this as an experimental process.

clausmichele:
I have a use case where we would need to interpolate some data having zeros instead of NaNs, would it be possible to add a parameter and let us choose the no-data value to use?

m-mohr:
Could you just replace the 0's with NaN before?

`apply(data, if(eq(x, 0), nan(), x))` should work.

clausmichele:
yes, this could be an option, thanks

Issue #234, opened 2021-04-13
dimension properties in apply_dimension after labels have been discarded


m-mohr:
So if the callback in apply_dimension doesn't return the same array length as the source array, the labels get discarded (i.e. enumerated). That may work well for some dimension types like bands, but could an issue for the temporal dimension as the temporal reference is not available anymore. Should the dimension type be changed to `other` in this case?

soxofaan:
While enumeration does not fit the usual calendar date/time structure, it could still be relative or symbolic time in some sense, so I don't think it is a problem to still call it a "temporal" dimension

m-mohr:
Yes, but then you also need to reset the TRS (ISO 8601 / Gregorian calendar / UTC), too. And after renaming the labels you'd need to set the TRS again.

soxofaan:
I'm not sure if I understand you correctly, but  you probably mean that the labels of a temporal dimension "MUST be strings compliant to ISO 8601"? So an dimension with enumerated labels can not be of type "temporal"?


in openapi.yaml I also found this https://github.com/Open-EO/openeo-api/blob/81dd248b9e8b8caeedbdd96c6a9dd8f3f56d0eb7/openapi.yaml#L4386-L4389 :
>  Data not compliant    with ISO 8601 can be represented as an             *Additional Dimension Object* with `type` set to             `temporal`.

I'm not sure though how to interpret this. Maybe this is a  leftover that became inconsistent?

m-mohr:
I'm not sure myself what the implications of a temporal dimension are, but it probably must refer somehow to a specific instance in time that can be resolved, otherwise the default TRS (gregorian calendar) can't apply. A dimension with enumerated labels can surely be temporal if the TRS defines how to map them to time.

The part you found in the Collection spec is referring to the Data Cube Extension in STAC and has nothing to do with our data cube model. In STAC a dimension that is compliant to the temporal dimension json schema must be specified in metadata using ISO8601, but that doesn't imply anything to our data cubes. It's just a schema thing. An "Additional Dimension Object" can still specify a temporal dimension using another trs.

PR #236, opened 2021-04-21
Fix reference system handling in apply_dimension + improvements

m-mohr:
Fixed description of `apply_dimension` with regards to reference systems. Made description easier to understand (hopefully)

See also #234.


apply_dimension.json
```diff
@@ -83,7 +83,7 @@
...
...rence system and the resolution are preserved only if the number of pixel values in the source dimension is equal to the number of values computed by the process. Otherwise, all other dimension properties change as defined in the list below.\n2. The source dimension is not the target dimension and the latter exists:\n   - The number of dimensions decreases by one as the source dimension is dropped.\n   - The target dimension properties name and type remain unchanged. All other dimension properties change as defined in the list below.\n3. The source dimension is not the target dimension and the latter does not exist:\n   - The number of dimensions remain unchanged, but the source dimension is replaced with the target dimension.\n   - The target dimension has the specified name and the type other. All other dimension properties are set as defined in the list below.\n\nUnless otherwise stated above, for the given (target) dimension the following applies:\n\n- the number of dimension labels is equal to the number of values computed by the process,\n- the dimension labels are incrementing integers starting from zero,\n- the resolution changes, and\n- the reference system is undefined.",
```
m-mohr, line 86:
```suggestion
        "description": "A data cube with the newly computed values.\n\nAll dimensions stay the same, except for the dimensions specified in corresponding parameters. There are three cases how the dimensions can change:\n\n1. The source dimension is the target dimension:\n   - The (number of) dimensions remain unchanged as the source dimension is the target dimension.\n   - The source dimension properties name and type remain unchanged.\n   - The dimension labels, the reference system and the resolution are preserved only if the number of pixel values in the source dimension is equal to the number of values computed by the process. Otherwise, all other dimension properties change as defined in the list below.\n2. The source dimension is not the target dimension and the latter exists:\n   - The number of dimensions decreases by one as the source dimension is dropped.\n   - The target dimension properties name and type remain unchanged. All other dimension properties change as defined in the list below.\n3. The source dimension is not the target dimension and the latter does not exist:\n   - The number of dimensions remain unchanged, but the source dimension is replaced with the target dimension.\n   - The target dimension has the specified name and the type other. All other dimension properties are set as defined in the list below.\n\nUnless otherwise stated above, for the given (target) dimension the following applies:\n\n- the number of dimension labels is equal to the number of values computed by the process,\n- the dimension labels are incrementing integers starting from zero,\n- the resolution changes, and\n- the reference system is undefined unless the previous reference system is defined for zero-based indices.",
```

In had something like that in a previous draft, but thought that's too much of a special cases and couldn't come up with any reasonable examples. Thoughts?

soxofaan: :frowning_face:   it's hard to find and review changes in this  one-line diff

m-mohr: Yes, obviously - thanks GitHub. ;-)

The only change is actually that at the end I've added 

> unless the previous reference system is defined for zero-based indices.

m-mohr:
I'll merge this end of the week if no one has objected until then.

Issue #231, opened 2021-04-09
Additional use cases


m-mohr:
Several capabilities could be useful in other projects:

- [x] Time-series fitting (see #240)
- [ ] Segmentation
- [ ] Spectral unmixing
- [ ] Regression
- [ ] Unsupervised classification
- [ ] Supervised classification (see #295)
- [ ] Spectral sharpening (see #221)
- [x] Convolutions (apply_kernel and apply_neighborhood)
- [x] Morphologic operations (apply_kernel and apply_neighborhood)
- [x] Spatial filtering (filter_spatial / load_collection)

We could also include image processing algorithms that allow transforming EO data collections into temporally equidistant and consistent time series datasets. Note: These techniques shall include as a minimum compositing (e.g. best-pixel, weighted averaging, max value), statistical aggregation (e.g. percentile, mean, standard deviation) and might additionally include capabilities for deriving coefficients or statistics of time series fits (e.g. integral, amplitude, inflection points, etc.).

req. 44 + 45

jdries:
convolutions == apply_kernel?
spectral_sharpening == resolution_merge (have a pull request for that!)
morhpological operations => dilation, erosion => apply_kernel

jdries:
supervised classification => random forest => use case 9@SRR3
regression => use case 8 @SRR3

jonathom:
**Spectral Unmixing**

I had a look and came up with two use cases that could be helpful towards developing processes. First, a **quick overview** of what I understand is contained in an unmixing process:
1. Preprocessing from spectral radiance to corrected reflectance values (important, but not part of the unmixing itself)
2. selecting or providing endmembers (spectral values of the materials or land covers in question)
3. unmixing, linear or non-linear, both with multiple possible approaches (finding the combination of endmembers that lead to the observed spectrum of pixels)
4. outputs are maps per endmember, describing the fraction of the pixels that is covered by the endmember

**UC1**: I want to know the fractions of materials/land cover that each pixel in my datacube contains. For this, I have a labeled dataset (geojson etc.) with endmember polygons that I confirmed as containing pure areas (material-wise). For preparation I need to aggregate mean values per endmember class from the satellite collection that I will be using. I want to use a standard linear unmixing method, in which the combined reflectances should sum to one. The result should be a datacube with as many bands as endmembers that I defined, with an optional time dimension remaining untouched.

[...]

PR #240, opened 2021-04-26
Curve Fitting

m-mohr:
A first draft for the curve fitting process. Name `fit_curve` and highly inspired by [scipy.curve_fit](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html). This is mostly to spur discussions rather than a  final proposal.

Related to #231


m-mohr:
The process is written from a use-case perspective based on an example from @clausmichele.

My question is whether this can be implemented by our back-ends, so for Platform, this is especially aimed at @jdries and @lforesta.

proposals/fit_curve.json
```diff
@@ -0,0 +1,83 @@
...
+    "summary": "Curve fitting",
+    "description": "Use non-linear least squares to fit a model function `y = f(x, parameters)` to data.",
+    "categories": [
+        "math"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "A labeled array of numbers. The array labels are the values for `x` (independent variable where the data is measured) and the array elements are the values for `y` (dependent data).",
```
soxofaan, line 12: I wonder how easy it will be for a user to get the input data in this labeled array format.

Wouldn't it be easier to have just two input array parameters: `x` and `y`? I think that will feel more familiar for most users. And it;s more self-documenting.

soxofaan: Also, the schema does not allow you to specify that the labels should be numeric, right?

m-mohr: Indeed, good points. It's probably easier and more generic to support x and y separately, but this would basically require the implementation of array_labels in back-ends to be able to support this process.

m-mohr: I've drafted a new proposal based on the discussion above. What do you think, @soxofaan ?

clausmichele: In my opinion, the ideal case would be to just have a datacube as input: the back-end translates the time labels and takes care of them. The input datacube should have dimensions: time (which will be fitted to x in fit_curve) and bands (which will be fitted to y) and the fitting will be computed on all the available bands.

soxofaan: @m-mohr  you use parameter names `data` for the dependent data (y) and `labels` for the independent variables (x).
Why not just `x` and `y`, which is standard terminology in machine learning literature and tools as far as I know.
Note that especially "labels" is tricky because that term is also often used as the target "y" you train with/for (in classification problems).

soxofaan: > the ideal case would be to just have a datacube as input

That is also an interesting solution, and probably the shortest path for the user

> The input datacube should have dimensions: time (which will be fitted to x in fit_curve) and bands (which will be fitted to y) and the fitting will be computed on all the available bands.

you seem to have a particular use case in mind here: multiple timeseries to be fitted at the same time.
I wasn't thinking about more general use cases, not just  time based x'es, For example: fit EVI vs NDVI (so band vs band).

I'm not sure if we should directly aim for the most general solution (e.g. general `fit_curve`), or focus on just timeseries fitting (e.g. `fit_timeseries`).

clausmichele: Well, yes, the specific use case is UC6, where we need to fit a seasonal function to each pixel in the data timeseries.

m-mohr: > In my opinion, the ideal case would be to just have a datacube as input: the back-end translates the time labels and takes care of them.

If the only feasible way is to compute this on a temporal dimension, this could be a nice shortcut, but we should ensure that we don't end up with two similar processes, one defined on arrays and one defined on data cubes.

> Why not just `x` and `y`, which is standard terminology in machine learning literature and tools as far as I know.

In openEO we use x and y for numerical scalar values in mostly all (math related) processes. I'm not sure whether there are implementations that actually use these conventions for anything, but that may also be misleading for users then. FWIW, scipy uses ydata and xdata...

> Note that especially "labels" is tricky because that term is also often used as the target "y" you train with/for (in classification problems).

I've simply used this to make it work well with openEO terminology. In the normal workflow you'd get a labeled array in apply_dimension or so and then pass the "data" parameter to "data" and then call array_labels(data) and pass it to the labels parameter. That seems pretty intuitive (ignoring any background people may have).

> I'm not sure if we should directly aim for the most general solution (e.g. general `fit_curve`), or focus on just timeseries fitting (e.g. `fit_timeseries`).

Me neither, but we should ensure that we don't end up with two similar processes, one defined on arrays and one defined on data cubes. What other dimensions could you do curve fitting for?

clausmichele:
Hi @m-mohr, I'm trying to implement fit_curve, but I don't understand how to write the fitting function.
`parameters+(parameters*cos(2*pi/362.25*x))+(parameters*sin(2*pi/362.25*x))` ? How do I specify which parameters to use if it's an array?

I will use dimension_labels to get the timestamps, is it ok in your opinion?
image

clausmichele:
@m-mohr could you please have a look here? I can't proceed with the implementation otherwise

m-mohr:
What is parameters in your case? Is that a single parameter?

clausmichele:
parameters is an array of parameters, for this specific case is [a0,a1,a2], because the function I need to fit is:
a0+(a1*cos(2*pi/362.25*x))+(a2*sin(2*pi/362.25*x))

m-mohr:
I've pushed an update to the process specification which adds a "parameters" parameter, which we (may?) need to define the number of parameters.

Here's how a process graph could look like:

````
{
  "process_graph": {
    "load": {
      "process_id": "load_collection",
      "arguments": {
        "id": "S2_L2A_T32TPS",
        "spatial_extent": null,
        "temporal_extent": null
      }
    },
    "labels": {
      "process_id": "dimension_labels",
      "arguments": {
        "data": {
          "from_node": "load"
        },
        "dimension": "temporal"
      }
    },
    "fit": {
      "process_id": "fit_curve",
      "arguments": {
        "data": {
          "from_node": "load"
        },
        "labels": {
          "from_node": "labels"
        },
        "parameters": [
          1,
          1,
          1
        ],
        "process": {
          "process_graph": {
            "765a9m5jr": {
              "process_id": "pi",
              "arguments": {}
            },
            "a0": {
              "process_id": "array_element",
              "arguments": {
                "data": {
                  "from_parameter": "parameters"
                },
                "index": 0
              }
            },
            "a1": {
              "process_id": "array_element",
              "arguments": {
                "data": {
                  "from_parameter": "parameters"
                },
                "index": 1
              }
            },
            "a2": {
              "process_id": "array_element",
              "arguments": {
                "data": {
                  "from_parameter": "parameters"
                },
                "index": 2
              }
            },
            "viadlidxt": {
              "process_id": "multiply",
              "arguments": {
                "x": 2,
                "y": {
                  "from_node": "765a9m5jr"
                }
              }
            },
            "dzcmqylf5": {
              "process_id": "multiply",
              "arguments": {
                "x": 2,
                "y": {
                  "from_node": "765a9m5jr"
                }
              }
            },
            "az83u1fgl": {
              "process_id": "divide",
              "arguments": {
                "x": {
                  "from_node": "viadlidxt"
                },
                "y": 362.25
              }
            },
            "l3wirzu5v": {
              "process_id": "divide",
              "arguments": {
                "x": {
                  "from_node": "dzcmqylf5"
                },
                "y": 362.25
              }
            },
            "4kt3ezdz5": {
              "process_id": "multiply",
              "arguments": {
                "x": {
                  "from_node": "az83u1fgl"
                },
                "y": {
                  "from_parameter": "x"
                }
              }
            },
            "4kceskraz": {
              "process_id": "multiply",
              "arguments": {
                "x": {
                  "from_node": "l3wirzu5v"
                },
                "y": {
                  "from_parameter": "x"
                }
              }
            },
            "o7s99j9pd": {
              "process_id": "cos",
              "arguments": {
                "x": {
                  "from_node": "4kt3ezdz5"
                }
              }
            },
            "kydj4bwy3": {
              "process_id": "sin",
              "arguments": {
                "x": {
                  "from_node": "4kceskraz"
                }
              }
            },
            "e19f54kwn": {
              "process_id": "multiply",
              "arguments": {
                "x": {
                  "from_node": "a1"
                },
                "y": {
                  "from_node": "o7s99j9pd"
                }
              }
            },
            "82rlj4j02": {
              "process_id": "multiply",
              "arguments": {
                "x": {
                  "from_node": "a2"
                },
                "y": {
                  "from_node": "kydj4bwy3"
                }
              }
            },
            "f9xv6wbqn": {
              "process_id": "add",
              "arguments": {
                "x": {
                  "from_node": "a0"
                },
                "y": {
                  "from_node": "e19f54kwn"
                }
              }
            },
            "rjb4pmv5a": {
              "process_id": "add",
              "arguments": {
                "x": {
                  "from_node": "f9xv6wbqn"
                },
                "y": {
                  "from_node": "82rlj4j02"
                }
              },
              "result": true
            }
          }
        }
      },
      "result": true
    }
  }
}
```

PS: This is purely theoretical and I've never actually tried or implemented it, so happy to take any suggestions that may make it easier or better etc.

clausmichele:
Thanks. I was actually thinking that the fit_curve process takes care of translating the temporal labels/strings (or whatever format they are) into the format required by the process (numerical).

m-mohr:
If that's a common use case we could think about this special case and support it, yes.

m-mohr:
Hmm, I'm just thinking whether it would be easier if this process returns an actual process graph instead of just the parameters?

clausmichele:
With the computed parameters [a0,a1,a2] and new time steps coming from another "testing" datacube (load_collection with a subsequent time extent) I need to predict the values the fitted function would take with the new time steps. This would basically return the a datacube with the same temporal extent as the "test" one, but different values: no more real data but predicted values.

m-mohr:
Results from a call with @clausmichele:
- There's no way yet to properly pass the "data" parameter - we were at a point to pass a raster-cube and the dimension name to work on, but decided against it as it's not very generic. Instead, I'll add an openEO process to get a (labeled?) array with values for a dimension.
- We may want to make the "labels" parameter optional and default to the labels/indices from the array.
- We want to allow labels to be date-times for convenience
- The process should still return the computed parameters so that they can be compared, stored or be re-used for later calls to fit_curve as defaults for the parameters.
- I should highlight in the process that the function should be stored as user-defined function so that users can re-use it for both the curve fitting and later applying the function with the computed parameters to the data cube.

m-mohr:
@clausmichele I made an attempt to cover what we discussed in a single process, which returns 2(!) data cubes. I have not had the time to fine-tune the descriptions yet, but would you think it would work this way for you?

clausmichele:
How would I have to use the result? How do I select the "parameters" datacube to predict values for a different datacube with different timesteps (same x,y,bands)?
Up to this point it's clear to me how to feed the data to the process and compute the output, but I'm not sure about what could follow.

m-mohr:
> How would I have to use the result?

You get the data cube with `array_element(result, 0)`

> How do I select the "parameters" datacube

As above, but use 1 instead of 0.

> to predict values for a different datacube with different timesteps (same x,y,bands)?

Not possible. That use case didn't came up yet or I did not understand it.

> Up to this point it's clear to me how to feed the data to the process and compute the output, but I'm not sure about what could follow.

I'm not sure what you want to have. You input a data cube (e.g. S2 with dimensions x,y,t,b) and get returned a data cube (x,y,t,b) with the predicted dates on the temporal dimension instead of the original dates. Could you give some more use-case examples?
I just tried to make the use-case from yesterday working.

clausmichele:
> > to predict values for a different datacube with different timesteps (same x,y,bands)?
> 
> Not possible. That use case didn't came up yet or I did not understand it.

Some time ago I've shared with you the python code of the use case implementation, which is here https://github.com/SARScripts/SAR2Cube_use_cases/blob/main/SAR2Cube_Forest_Change.ipynb

You can see that we need to fit the curve function over a certain period (2 years of data in this case) but then we need to predict values of a different period (the "testing" one) as well.

Maybe this sketch can clarify the pipeline:
image

m-mohr:
> Some time ago I've shared with you the python code of the use case implementation, which is here https://github.com/SARScripts/SAR2Cube_use_cases/blob/main/SAR2Cube_Forest_Change.ipynb

Yes, but the code is mostly undocumented and has a lot of other things in it. I can't exactly figure out what the code does and what extract of the code is needed for the openEO use case, sorry.

> You can see that we need to fit the curve function over a certain period (2 years of data in this case) but then we need to predict values of a different period (the "testing" one) as well.
> 
> Maybe this sketch can clarify the pipeline:
> image

Okay, so if I understand correctly the issue is that the "parameters" parameter is not a datacube, but an array, right? So the node "array_element (parameters)" is meant to be passed into the "parameters" parameter of fit_curve, right?

m-mohr:
Or do you want to predict values without an actual curve_fitting just based on the parameters that are returned for the first fit_curve?

clausmichele:
> 
> 
> Or do you want to predict values without an actual curve_fitting just based on the parameters that are returned for the first fit_curve? Then we'd need two processes: fit + predict. The first fit_curve would be replaced with fit + predict and the second fit_curve would only use predict.

This! It is indeed what we would need

m-mohr:
Okay, thanks for clarifying, that makes things likely a bit more complicated, but I'll come up with another proposal then.

m-mohr:
I've committed a new proposal that splits the fit_curve process into two parts: fit_curve and predict_curve (names TBD). I have to polish the descriptions a lot, but for now it's meant to discuss whether the proposal is feasible and makes sense for many use cases.

Here's an example:

First of all, the model function should be stored separately, here as an example as `example_fitting_function_eurac`. We'll use this function in the other process below.
```json
{
  "id": "example_fitting_function_eurac",
  "parameters": [
    {
      "name": "x",
      "schema": {
        "type": "number"
      }
    },
    {
      "name": "parameters",
      "schema": {
        "type": "array",
        "items": {
          "type": "number"
        }
      }
    }
  ],
  "process_graph": {
    "765a9m5jr": {
      "process_id": "pi",
      "arguments": {}
    },
    "a0": {
      "process_id": "array_element",
      "arguments": {
        "data": [
          "parameters"
        ],
        "index": 0
      }
    },
    "a1": {
      "process_id": "array_element",
      "arguments": {
        "data": {
          "from_parameter": "parameters"
        },
        "index": 1
      }
    },
    "a2": {
      "process_id": "array_element",
      "arguments": {
        "data": {
          "from_parameter": "parameters"
        },
        "index": 2
      }
    },
    "dzcmqylf5": {
      "process_id": "multiply",
      "arguments": {
        "x": 2,
        "y": {
          "from_node": "765a9m5jr"
        }
      }
    },
    "viadlidxt": {
      "process_id": "multiply",
      "arguments": {
        "x": 2,
        "y": {
          "from_node": "765a9m5jr"
        }
      }
    },
    "l3wirzu5v": {
      "process_id": "divide",
      "arguments": {
        "x": {
          "from_node": "dzcmqylf5"
        },
        "y": 362.25
      }
    },
    "az83u1fgl": {
      "process_id": "divide",
      "arguments": {
        "x": {
          "from_node": "viadlidxt"
        },
        "y": 362.25
      }
    },
    "4kceskraz": {
      "process_id": "multiply",
      "arguments": {
        "x": {
          "from_node": "l3wirzu5v"
        },
        "y": {
          "from_parameter": "x"
        }
      }
    },
    "4kt3ezdz5": {
      "process_id": "multiply",
      "arguments": {
        "x": {
          "from_node": "az83u1fgl"
        },
        "y": {
          "from_parameter": "x"
        }
      }
    },
    "kydj4bwy3": {
      "process_id": "sin",
      "arguments": {
        "x": {
          "from_node": "4kceskraz"
        }
      }
    },
    "o7s99j9pd": {
      "process_id": "cos",
      "arguments": {
        "x": {
          "from_node": "4kt3ezdz5"
        }
      }
    },
    "82rlj4j02": {
      "process_id": "multiply",
      "arguments": {
        "x": {
          "from_node": "a2"
        },
        "y": {
          "from_node": "kydj4bwy3"
        }
      }
    },
    "e19f54kwn": {
      "process_id": "multiply",
      "arguments": {
        "x": {
          "from_node": "a1"
        },
        "y": {
          "from_node": "o7s99j9pd"
        }
      }
    },
    "f9xv6wbqn": {
      "process_id": "add",
      "arguments": {
        "x": {
          "from_node": "a0"
        },
        "y": {
          "from_node": "e19f54kwn"
        }
      }
    },
    "rjb4pmv5a": {
      "process_id": "add",
      "arguments": {
        "x": {
          "from_node": "f9xv6wbqn"
        },
        "y": {
          "from_node": "82rlj4j02"
        }
      },
      "result": true
    }
  }
}
```

Here's an example workflow based on @clausmichele's image above (if I understood it correctly as the actual values/inputs are missing partially from the image):
```json
{
  "process_graph": {
    "1": {
      "process_id": "load_collection",
      "arguments": {
        "id": "COPERNICUS/S2",
        "spatial_extent": null,
        "temporal_extent": [
          "2015-01-01T00:00:00Z",
          "2018-01-01T00:00:00Z"
        ],
        "bands": null
      }
    },
    "2": {
      "process_id": "filter_temporal",
      "arguments": {
        "data": {
          "from_node": "1"
        },
        "extent": [
          "2015-01-01T00:00:00Z",
          "2017-01-01T00:00:00Z"
        ]
      }
    },
    "3": {
      "process_id": "filter_temporal",
      "arguments": {
        "data": {
          "from_node": "1"
        },
        "extent": [
          "2017-01-01T00:00:00Z",
          "2018-01-01T00:00:00Z"
        ]
      }
    },
    "4": {
      "process_id": "fit_curve",
      "arguments": {
        "data": {
          "from_node": "2"
        },
        "function": {
          "process_graph": {
            "1": {
              "process_id": "example_fitting_function_eurac",
              "arguments": {
                "x": {
                  "from_parameter": "x"
                },
                "parameters": {
                  "from_parameter": "parameters"
                }
              },
              "result": true
            }
          }
        },
        "parameters": [
          1,
          1,
          1
        ],
        "dimension": "t"
      }
    },
    "5": {
      "process_id": "predict_curve",
      "arguments": {
        "data": {
          "from_node": "3"
        },
        "parameters": {
          "from_node": "4"
        },
        "function": {
          "process_graph": {
            "1": {
              "process_id": "example_fitting_function_eurac",
              "arguments": {
                "x": {
                  "from_parameter": "x"
                },
                "parameters": {
                  "from_parameter": "parameters"
                }
              },
              "result": true
            }
          }
        },
        "dimension": "t"
      },
      "result": true
    }
  }
}
```

image

The predict_curve process may need a "prediction_labels" parameter or there needs to be a way to add "no-data" labels to the t dimension to be able to actually predict values other than the ones that are already in the 2017 cube?!

m-mohr:
This seems to work, so shall we merge, @clausmichele ?

clausmichele:
Yes, for me it's fine!

Issue #237, opened 2021-04-21
MultiPolygon support for mask_polygon


soxofaan:
https://github.com/Open-EO/openeo-processes/blob/75059c6566a7c8951d04467665252e74d5bb9c3a/mask_polygon.json#L20
> A GeoJSON object containing a polygon. The provided feature types can be one of the following:
>    A Polygon geometry,
>    a GeometryCollection containing Polygons,
>    a Feature with a Polygon geometry or
>    a FeatureCollection containing Features with a Polygon geometry.

Why is MultiPolygon not allowed, but GeometryCollection is, while they are equivalent in the context of masking?

Related: `aggregate_spatial` describes behavior for points and lines as well, so it might make sense to also copy that to `mask_polygon`. I vaguely remember we discussed this before, but I can't reference right now.

As far as I could recover from git, the whitelist was there from the start (added in #20), but I can't find a reason for the excluded types.

m-mohr:
MultiPolygon should indeed be allowed, I think.
I'm actually thinking, why not just make the process mask_spatial and allow all geometries as defined also in aggregate_spatial?

m-mohr:
Probably something to do together with the vector data cube update for #68 as this would change the process anyway.

soxofaan:
> why not just make the process mask_spatial and allow all geometries as defined also in aggregate_spatial?

I'm not sure, for masking you also have `mask` for masking with a raster mask, which is also spatial, so `mask_spatial` would be confusing. It could make sense to go for `mask_geometry` though.

jdries:
Don't really see why this moved from a very simple, backwards compatible, update to the spec, to a breaking change?
Breaking changes really require a lot more motivation than what we have here. (The last rename of masking processes caused quite a bit of breakage, time loss and subsequent complaints from my users, hence the concern.)

soxofaan:
yes indeed, I just wanted to raise the question about why certain geometry types are excluded for `mask_polygon` (in VITO backend we do support MultiPolygon for example).

The discussion about rename is something else and a lot lower priority for me

m-mohr:
Strictly speaking, adding Multipolygon is also a breaking change. It "breaks" interoperability with back-ends that do not update to follow to implement the new version with MultiPolygon. Nevertheless, it feels like we have somehow converged towards calling only user-facing changes breaking?

[...]

PR #242, opened 2021-04-27
mask_polygon: Support multi polygons #237

m-mohr:
This is a proposal to support multi polygons instead of just polygons in `mask_polygon`. See #237 for context.

It's not clear yet which target version this is for as it's not clear whether this is a breaking change.


soxofaan (approved):

m-mohr:
No further feedback has been received, so I'll merge this.

Issue #252, opened 2021-05-18
Clarify that Multi* geometries are a single entities in computations


m-mohr:
Originates from https://github.com/Open-EO/openeo-geopyspark-driver/issues/76:

It's not clear what the implications of using e.g. a MultiPoligon over a FeatureCollection containing Polygons is.

We should clarify that *each geometry is a single entity in processing*. For example, if you provide a MultiPolygon in aggregate_spatial, it will result in a single value computed for the MultiPolygon. On the other hand, if you specify a FeatureCollection with two Polygons, this will result in two values, one per polygon.

cc @jonathom

PR #253, opened 2021-05-18
Clarified use of geometries in aggregate_spatial processes

m-mohr:
Improvements for aggregate_spatial processes, see #252:

- Clarified: One value will be computed per GeoJSON geometry, which means that, for example, a single value will be computed for a `MultiPolgon`, but two values will be computed for a `FeatureCollection` or `GeometryCollection` containing two polygons.
- Moved the details about geometries from the process description to the description of the parameter `geometries` to have related descriptions in a single place.

This behavior of MultiPolygons / Polygons is aligned with the VITO implementation AFAIK. Not sure about GeometryCollections.

### DIFF
--- a/aggregate_spatial.json
+++ b/aggregate_spatial.json
@@ -1,7 +1,7 @@
 {
     "id": "aggregate_spatial",
     "summary": "Zonal statistics for geometries",
-    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions. This process passes a list of values to the reducer. In contrast, ``aggregate_spatial_binary()`` passes two values, which may be better suited especially for UDFs in case the number of values gets too large to be processed at once.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.\n\nThe data cube must have been reduced to only contain two spatial dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise, this process fails with the `TooManyDimensions` exception.\n\nThe number of total and valid pixels is returned together with the calculated values.",
+    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions. This process passes a list of values to the reducer. In contrast, ``aggregate_spatial_binary()`` passes two values, which may be better suited especially for UDFs in case the number of values gets too large to be processed at once.\n\nThe data cube must have been reduced to only contain two spatial dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise, this process fails with the `TooManyDimensions` exception.\n\nThe number of total and valid pixels is returned together with the calculated values.",
     "categories": [
         "cubes",
         "aggregate & resample"
@@ -17,7 +17,7 @@
         },
         {
             "name": "geometries",
-            "description": "Geometries as GeoJSON on which the aggregation will be based.",
+            "description": "Geometries as GeoJSON on which the aggregation will be based. One value will be computed per GeoJSON geometry, which means that, for example, a single value will be computed for a `MultiPolgon`, but two values will be computed for a `FeatureCollection` or `GeometryCollection` containing two polygons.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.",
             "schema": {
                 "type": "object",
                 "subtype": "geojson"
--- a/proposals/aggregate_spatial_binary.json
+++ b/proposals/aggregate_spatial_binary.json
@@ -1,7 +1,7 @@
 {
     "id": "aggregate_spatial_binary",
     "summary": "Zonal statistics for geometries by binary aggregation",
-    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions. This process consecutively passes a pair of values to the reducer. This may be better suited especially for UDFs in case the number of values gets too large to be processed at once. In contrast, ``aggregate_spatial()`` passes a list of values.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise, this process fails with the `TooManyDimensions` exception.\n\nThe number of total and valid pixels is returned together with the calculated values.",
+    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions. This process consecutively passes a pair of values to the reducer. This may be better suited especially for UDFs in case the number of values gets too large to be processed at once. In contrast, ``aggregate_spatial()`` passes a list of values.\n\nThe data cube must have been reduced to only contain two raster dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise, this process fails with the `TooManyDimensions` exception.\n\nThe number of total and valid pixels is returned together with the calculated values.",
     "categories": [
         "cubes",
         "aggregate & resample"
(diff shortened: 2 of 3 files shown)


edzer:
Hmm. According to the [simple feature access](https://www.ogc.org/standards/sfa) standard, a `GeometryCollection` is the geometry for a single feature, and it sounds more logical to me that you'd like the aggregated value over that collection. For a `FeatureCollection` (not a concept in sfa), it makes sense to return the aggregate value for each feature.

mkadunc:
I agree with Edzer - GeomCollection is a single geometry in the same way as a MultiPolygon.

I propose we require the parameter to always be an array, e.g. a FeatureCollection, and the aggregation happens over all elements of the array (FeatureCollection).

I would try to avoid API with branching behaviour based on (runtime) type of parameter objects - such things tend to be difficult to document and cause weird issues.

m-mohr:
Indeed, I was wondering about that and unsure about GeometryCollection. I should have had a look into the spec. So instead of making it "per Geometry", the wording should be about "per Feature". I'll change it accordingly.

We can't change the actual input data type right now to not introduce breaking changes, but I think the actual input data type is okay, it just needs a clarification. @mkadunc

mkadunc:
It’s OK indeed. I guess we can’t formally specify geojson FeatureCollection as param type, so we could just clarify in text that this is the kind of geojson that is expected.

m-mohr:
Okay, I rewrote that part:

> One value will be computed per GeoJSON Feature or Geometry, multiple values will be computed for a `FeatureCollection` containing multiple features. For example, a single value will be computed for a `MultiPolgon`, but two values will be computed for a `FeatureCollection` containing two polygons.

aggregate_spatial.json
```diff
@@ -17,7 +17,7 @@
         },
         {
             "name": "geometries",
-            "description": "Geometries as GeoJSON on which the aggregation will be based.",
+            "description": "Geometries as GeoJSON on which the aggregation will be based.\n\nOne value will be computed per GeoJSON `Feature`, `Geometry` or `GeometryCollection`. For a `FeatureCollection` multiple values will be computed, one value per contained `Feature`. For example, a single value will be computed for a `MultiPolgon`, but two values will be computed for a `FeatureCollection` containing two polygons.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.",
```
jonathom, line 20:
```suggestion
            "description": "Geometries as GeoJSON on which the aggregation will be based.\n\nOne value will be computed per GeoJSON `Feature`, `Geometry` or `GeometryCollection`. For a `FeatureCollection` multiple values will be computed, one value per contained `Feature`. For example, a single value will be computed for a `MultiPolygon`, but two values will be computed for a `FeatureCollection` containing two polygons.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.",
```

proposals/aggregate_spatial_binary.json
```diff
@@ -18,7 +18,7 @@
         },
         {
             "name": "geometries",
-            "description": "Geometries as GeoJSON on which the aggregation will be based.",
+            "description": "Geometries as GeoJSON on which the aggregation will be based.\n\nOne value will be computed per GeoJSON `Feature`, `Geometry` or `GeometryCollection`. For a `FeatureCollection` multiple values will be computed, one value per contained `Feature`. For example, a single value will be computed for a `MultiPolgon`, but two values will be computed for a `FeatureCollection` containing two polygons.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.",
```
jonathom, line 21:
```suggestion
            "description": "Geometries as GeoJSON on which the aggregation will be based.\n\nOne value will be computed per GeoJSON `Feature`, `Geometry` or `GeometryCollection`. For a `FeatureCollection` multiple values will be computed, one value per contained `Feature`. For example, a single value will be computed for a `MultiPolygon`, but two values will be computed for a `FeatureCollection` containing two polygons.\n\n- For **polygons**, the process considers all pixels for which the point at the pixel center intersects with the corresponding polygon (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nThus, pixels may be part of multiple geometries and be part of multiple aggregations.",
```

jonathom (commented):
Nice, seems logically straightforward and understandable to me.

aggregate_spatial.json
m-mohr: Ouch, fixed...

soxofaan:
So this would make the current VITO GeometryCollection handling invalid ? That would be annoying as we have internal use cases depending on it, and there is also a lot of documentation/demo's/webinars showing this "wrong way".

m-mohr:
How do you handle GeometryCollection? @soxofaan

soxofaan:
we return a separate aggregation for each geometry in the collection

m-mohr:
> we return a separate aggregation for each geometry in the collection

Hmm... the issue here is that GeometryCollection is defined to be a Geometry itself, which makes sense to me if you'd, for example, want to run a use case, which e.g. aggregates over multiple geometry types at once. There's no other way to achieve this except for using a geometry collection.

Unfortunately, this was not defined well enough, but looking at SFA there's probably no other way to interpret it in a consistent way with SFA. And for interoperability, we should define it now rather than leaving it open...

jdries:
Perhaps we can also repeat the warning that's included in the geojson spec:

>    To maximize interoperability, implementations SHOULD avoid nested
>    GeometryCollections.  Furthermore, GeometryCollections composed of a
>    single part or a number of parts of a single type SHOULD be avoided
>    when that single part or a single object of multipart type
>    (MultiPoint, MultiLineString, or MultiPolygon) could be used instead.

And reflect that in our own examples by not using geometrycollection at all. It's anyway a type that you would use for features composed of a combination of polygons and lines or points, so rather exotic.
That way, our implementation can stay as-is for now, and we gradually move away from this annoying type.
I also believe that if the user sends a featurecollection, which uses geometrycollection to define feature geometry, we would still do the right thing.

edzer:
> I also believe that if the user sends a featurecollection, which uses geometrycollection to define feature geometry, we would still do the right thing.

@jdries according to https://datatracker.ietf.org/doc/html/rfc7946#page-12 , a featureCollection is an array of feature objects; each feature object needs to have a geometry (section 3.2). Sets of geometries for multiple features should never be put in a GeometryCollection, that is how I read this. If a single feature has a GeometryCollection as geometry type, so be it.

I agree that we can discourage use of GeometryCollection in general. They are part of a formalism, and hence "needed", but almost always a nuisance, I see them rarely used intentionally.

m-mohr:
I've added the following: 

> To maximize interoperability, a nested `GeometryCollection` should be avoided. Furthermore, a `GeometryCollection` composed of a single type of geometries should be avoided in favour of the corresponding multi-part type (e.g. `MultiPolygon`).

jonathom:
> GEE: has two processes, reduceRegion ...

@m-mohr I can't seem to find that process (in the web editor), what am I missing here?

@soxofaan I've been failing in passing multiple geometries to VITO back-end. Could you kindly explain how to do this? Passing geojson FeatureCollection (js client) doesn't work, passing shapely MultiPolygon results in only one output (which is expected I think, but shapely talks about Multi~ Collections as geometry collections, which completes my confusion). Already mentioned [here](https://github.com/Open-EO/openeo-geopyspark-driver/issues/76).

m-mohr:
@jonathom I was referring to the native GEE functions, not openEO processes.

mkadunc:
> Sentinel Hub: Seems to require passing an array. Have not looked into details, maybe Miha can sound in how they implement it?

Sentinel Hub service natively handles either a single `bbox`, a single `Polygon` or a single `MultiPolygon`; all these produce just one aggregation (i.e. MultiPolygon is treated as a single Geometry for the purpose of calculating statistics).

The sentinel-hub-python-driver does indeed expect the `geometries` argument to be an array of geometry objects. This is the relevant part of the code:
```Python
geometries = self.validate_parameter(arguments, "geometries", required=True, allowed_types=[list])
```

soxofaan:
> I've been failing in passing multiple geometries to VITO back-end. Could you kindly explain how to do this?

@jonathom it is illustrated in this notebook: https://github.com/Open-EO/openeo-python-client/blob/master/examples/notebooks/openeo-terrascope-webinar.ipynb   (scroll down half way, to the section about "time series")
In the Python client you pass a GeometryCollection object, and the VITO backend will calculate a separate aggregation for each geometry in that collection

m-mohr:
Is anyone up for a final review?

m-mohr:
Merging for now as no new feedback came in. If there's anything to discuss, please open a new issue.

Issue #258, opened 2021-05-31
Remove binary versions of processes?


m-mohr:
We seem to conclude that having a process `reduce_spatial(raster-cube, process, context) -> raster-cube` is the way to go. [...]

One thing that came to mind is that for aggregate_spatial and reduce_dimension we have the binary variants (e.g. `reduce_dimension_binary`). We'd need a binary variant also for reduce_spatial, but on the other hand, I'm wondering whether we should remove these binary variants for now? There's no implementation yet and I've also not heard that anyone is actually planning to implement one of them.

_Originally posted by @m-mohr in https://github.com/Open-EO/openeo-processes/issues/226#issuecomment-851453918_

They were experimental before, so we can easily remove them without breaking change.

Issue #226, opened 2021-03-31
How to compute the average over spatial dimensions?


clausmichele:
I would like to compute the average over the data I'm loading from a small bbox, to reduce noise and get a smoother signal. It is what I did in python in [User Story 2](https://rawcdn.githack.com/openEOPlatform/SRR1_notebooks/1178ef77a557a5054c609fcb7a599dbe22958232/OpenEO%20Platform%20-%20User%20story%202.html) using the Xarray method `.mean(['x','y'], skipna=True)`

How can we perform this task with opneEO processes? I considered various options:
1. reduce_dimension over x and then y? Is this supported by the back-ends?
2. aggregate_spatial without a polygon but over the whole area?
3. resample_spatial where the target is 1 pixel and covers the input area?

jdries:
Did you try something like the example below already? In principle, if you use aggregate_spatial, you can leave out other spatial filtering. So you only need to specify your small bbox once, which is most easily done with the shapely 'box' function.
```
from shapely.geometry import box
cube = c.load_collection("TERRASCOPE_S2_NDVI_V2",bands=["NDVI_10M"]).filter_temporal(["2019-05-24T00:00:00Z", "2019-05-30T00:00:00Z"])
cube.polygonal_mean_timeseries(box(minx=5.027, maxx=5.0438, miny=51.1974, maxy=51.2213)).execute()
```

clausmichele:
Thanks @jdries. I've just tried and the result is the following as a netCDF. I put here the output without applying polygonal_mean_timeseries and with it:
Original result:
```
<xarray.Dataset>
Dimensions:      (t: 524, x: 4, y: 5)
Coordinates:
    spatial_ref  int64 ...
  * t            (t) datetime64[ns] 2017-01-05 2017-01-06 ... 2019-12-28
  * x            (x) float64 6.855e+05 6.855e+05 6.855e+05 6.855e+05
  * y            (y) float64 5.135e+06 5.135e+06 5.135e+06 5.135e+06 5.135e+06
Data variables:
    VH_gamma0    (t, y, x) float32 ...
    VV_gamma0    (t, y, x) float32 ...
Attributes:
    nodata:        0.0
    crs:           +proj=utm +zone=32 +datum=WGS84 +units=m +no_defs 
    grid_mapping:  spatial_ref
```
Using aggregate_spatial (polygonal_mean_timeseries)
```
<xarray.Dataset>
Dimensions:        (feature: 1, time: 524)
[...]

PR #259, opened 2021-05-31
Remove binary reducers #258

m-mohr:
For discussion: Do we actually need the binary reducers and is someone planning to implement them? If not, we may want to clean-up and remove them as otherwise, we'd also need to provide them for e.g. reduce_spatial.


jdries:
Actually we do support the binary variant in our backend, but do not seem to properly declare it:
https://github.com/Open-EO/openeo-python-driver/blob/master/openeo_driver/ProcessGraphDeserializer.py#L842

In fact, it could very well be that we don't properly support the non-binary variant...
Do we actually need them: not if the backend can auto-detect that an optimization is possible, in case the process (graph) is a commutative + associative reducer.

m-mohr:
Good to know, I've only looked into the Hub and there was no back-end declaring support for it.

Back-ends can surely optimize if it doesn't influence the result. I think the main use case for the binary variants were UDFs.

This issue is pretty much to discuss whether we think those processes are still good to have or whether we could remove them (having in mind #226 and a potential new process `reduce_spatial_binary`). So all feedback will be appreciated.

soxofaan:
> Actually we do support the binary variant in our backend, but do not seem to properly declare it:
https://github.com/Open-EO/openeo-python-driver/blob/master/openeo_driver/ProcessGraphDeserializer.py#L842

I don't think we (still) support it actually, as far as I understand, the related code path got lost during some 0.4-1.0  related refactors

m-mohr:
Removing. Has not been implemented and the sole use-case is UDFs, which itself would likely not lead to good performance as binary UDFs are too fine-grained.

Issue #260, opened 2021-05-31
Unpredictable result for first/last in aggregate_spatial / reduce_spatial


m-mohr:
While writing a proposal for reduce_spatial, it came up that I thought about "use cases" which compute a value depending on the order of the values. For example, passing a reducer such as first or last (or median) will not give predictable results depending on how the implementation "collects" the values from the underlying spatial areas/dimensions. 

Has someone experiences from implementations how this could be handled and documented gracefully?

For now, I added text to the spec that documents a "naive" approach:
> The list of values has an undefined order, therefore processes such as ``last()`` and ``first()`` that depend on the order of the values will lead to unpredictable results.

That makes it "documented", but not very useful for cases where you don't run e.g. "sort" before. What's your take on this?

soxofaan:
I don't think this is worth to spend a lot of documentation on to be honest. 
There is no intrinsic  ordering for the combination of two dimensions (even if each dimension has a intrinsic order). I think it's enough to just warn the user to not assume some arbitrary order. Your proposal "naive" text is fine for me

m-mohr:
Has been improved a bit in PR #261

Issue #126, opened 2020-01-16
aggregate_spatial: Statistics for more than one dimension?


m-mohr:
@mkadunc wrote in https://github.com/Open-EO/openeo-processes/pull/123#discussion_r367501563 regarding aggregate_polygon: 

> Why are we limiting this to one remaining dimension - e.g. why not get statistics over space for each (time,band) combination?

m-mohr:
Postponing to 1.0-final if there's no veto...

m-mohr:
Hey @jdries , you were talking about aggregate_polygon today and it seems you are invested most into this process. What do you think about this? Ever stumbled across a need for this with your use cases?

jdries:
Yes, we actually do retain the band dimension as well, so you can get a 'multiband' aggregated time series. This is maybe not compliant with the current spec, strictly speaking?

m-mohr:
Sounds like that. Should we adopt the specification of aggregate_spatial (fka aggregate_polygon)?

clausmichele:
@m-mohr Rising up this issue. Why don't we make available aggregate_spatial over the full spatial dimension instead of a polygon? Defining the polygon is cumbersome compared to get the statistics over the area I've already carefully selected. Like a resampling with the target size as 1 pixel in spatial domain.

m-mohr:
@clausmichele That seems to be a different issue, right? For your usecase you'd usually do a reduce, but that doesn't really work as you have two spatial dimensions instead of one.

clausmichele:
exactly, but since I see that we can't use reduce_dimension with the two spatial dimension at the moment, I thought we could use aggregate_spatial instead.

soxofaan:
<ping>


In the mean time, we actually have a couple of example notebooks, demo's and videos (against the VITO backend) that illustrate this 'multiband' aggregated time series feature, which is strictly speaking violating the spec.

m-mohr:
This will be implemented in 2.0.0. See #382

PR #261, opened 2021-05-31
reduce_spatial #226 and clarifications #260

m-mohr:
- Add new process reduce_spatial #226 ~(reduce_spatial_binary still has to be added, depends on outcomes in #259)~
- Clarified that in aggregate_spatial (and reduce_spatial) the order of values is not predefined, see #260 (happy to improve this, not sure how though)
- Open question: How to handle any vertical dimensions? Should we add a parameter where a user can optionally choose a vertical dimension to include into the reduction operation? Could also be added later as a non-breaking change.


aggregate_spatial.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Aggregates statistics for one or more geometries (e.g. zonal statistics for polygons) over the spatial dimensions.\n\nThis process passes a list of values to the reducer. The list of values has an undefined order, therefore processes such as ``last()`` and ``first()`` that depend on the order of the values will lead to unpredictable results. In contrast, ``aggregate_spatial_binary()`` passes two values, which may be better suited especially for UDFs in case the number of values gets too large to be processed at once. An 'unbounded' aggregation over the full extent of the horizontal spatial dimensions can be computed with the process ``reduce_spatial()``.\n\nThe data cube must have been reduced to only contain two spatial dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series. Otherwise, this process fails with the `TooManyDimensions` exception.\n\nThe number of total and valid pixels is returned together with the calculated values.",
```
soxofaan, line 4: I know this is the original text, but there is room for improvement here:
> The data cube must have been reduced to only contain two spatial dimensions and a third dimension the values are aggregated for, for example the temporal dimension to get a time series

With "must have been reduced" it sounds that the user must do something before using `aggregate_spatial`, while in most use cases (all uses cases as far as I've seen) the input data will be just fine as-is.

Also: in the VITO backend we *do* support `aggregate_spatial` on a 4D cube ("t", "bands", "x", "y") and return a "3D" vector cube ("t", "band", "geometry_id").  So I'm not sure this note on `TooManyDimensions` is actually necessary .

proposals/reduce_spatial.json
```diff
@@ -0,0 +1,79 @@
+{
+    "id": "reduce_spatial",
+    "summary": "Reduce horizontal spatial dimensions",
```
soxofaan, line 3: I understand that "horizontal" means "x" and "y" in GIS-speak, but this might still be confusing to non-experts. Maybe just say "Reduce spatial dimensions 'x' and 'y'"

proposals/reduce_spatial.json
```diff
@@ -0,0 +1,79 @@
...
+    "description": "Applies a unary reducer to a data cube dimension by collapsing all the pixel values along the horizontal spatial dimensions (i.e. axes `x` and `y`) into an output value computed by the reducer. The horizontal spatial dimensions are dropped.\n\nThis process passes a list of values to the reducer. The list of values has an undefined order, therefore processes such as ``last()`` and ``first()`` that depend on the order of the values will lead to unpredictable results.\n\nIn contrast to this process, ``reduce_spatial_binary()`` passes two values, which may be better suited especially for UDFs in case the number of values gets too large to be processed at once. An aggregation over certain spatial areas can be computed with the process ``aggregate_spatial()``.",
```
soxofaan, line 4: > Applies a unary reducer to a data cube dimension by

I think  "dimension" should not be there
(the process tackles two dimensions at same time)

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clarified disallowed characters in subtype `file-path`.
 - Clarified that UDF source code must contain a newline/line-break (affects `run_udf`).
 - `aggregate_spatial`, `aggregate_spatial_binary`: Clarified that Features, Geometries and GeometryCollections are a single entity in computations. Only FeatureCollections are multiple entities. [#252](https://github.com/Open-EO/openeo-processes/issues/252)
+- `aggregate_spatial`: Clarified that the values have no predefined order and reducers such as `first`, `last` and `median` return unpredictable results. [#260](https://github.com/Open-EO/openeo-processes/issues/260)
```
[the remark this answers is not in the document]
m-mohr: True, fixed.

proposals/reduce_spatial.json
m-mohr: Makes sense, changed.

proposals/reduce_spatial.json
m-mohr: Yes, true. Removed.

aggregate_spatial.json
m-mohr: Most collection I've seen was x,y,t,bands, so users would need to reduce before. So I think the language is fine as it is.

Yeah, we were pretty strict at the beginning by only allowing 3D. I'm not sure all back-ends can/want to support 4D? Anyway, this is the issue to discuss it: #126

proposals/reduce_spatial.json
```diff
@@ -0,0 +1,79 @@
+{
+    "id": "reduce_spatial",
+    "summary": "Reduce spatial dimensions 'x' and 'y",
```
soxofaan, line 3: I think the `y` is missing a closing single quote

proposals/reduce_spatial.json
```diff
@@ -0,0 +1,79 @@
...
+    "description": "Applies a unary reducer to a data cube by collapsing all the pixel values along the horizontal spatial dimensions (i.e. axes `x` and `y`) into an output value computed by the reducer. The horizontal spatial dimensions are dropped.\n\nAn aggregation over certain spatial areas can be computed with the process ``aggregate_spatial()``.\n\nThis process passes a list of values to the reducer. The list of values has an undefined order, therefore processes such as ``last()`` and ``first()`` that depend on the order of the values will lead to unpredictable results.",
```
soxofaan, line 4: Just wondering: is "unary reducer" valid terminology? To me "unary" means operating on a single entity, and "reducing" is about converting a sequence of entities to a single entity. They don't seem to fit well together.

I guess we added "unary" to make the distinction with the "binary" variants, but now that we dropped these in #258, we might remove "unary" from these descriptions as well?

(feel free to address in separate issue/PR)

apply.json
```diff
@@ -16,7 +16,7 @@
         },
         {
             "name": "process",
-            "description": "A unary process to be applied on each value, may consist of multiple sub-processes.",
+            "description": "A process to be applied on each individual value, may consist of multiple sub-processes. The process must accept and return a single value, for example ``abs()`` or ``linear_scale_range()``.",
```
soxofaan, line 19: `linear_scale_range`  requires at least 3 arguments, so that would be a wrong example, right? I understand that you can "hardcode" the `inputMin` and `inputMax` in the sub-process-graph, but I think  the current description is pretty confusing for the less-experienced openEO user.

m-mohr: It's exactly there for this purpose, to show that it is about the "data" values (parameter "x"). The others are basically the "options". Hard to describe, but the aim was to avoid that people only look for processes with exactly one parameter. I'm not exactly sure how to improve the documentation right now.

mkadunc (approved):

apply.json
soxofaan: I don't think there is "room" to explain how to use `linear_scale_range` in `apply` (except maybe with an actual example), so I would keep it superficial with indeed something like `ln`, `log`, `sin`, ...

m-mohr: log and linear_scale_range have the same "issue" with requiring an additional parameter. I don't see a difference, to be honest.

soxofaan: ah ok, I was assuming that `log` had default `base=10`, but apparently not, so `log` has indeed same issue.

My original remark is that current description is misleading on several levels:

> The process must accept and return a single value, for example ``abs()`` or ``linear_scale_range()``

Superficially: `linear_scale_range` does not accept a single value, it has three required arguments, so this statement looks wrong.

More precisely:  you can not pass `abs` or `linear_scale_range` directly to `apply`, you have to embed it in a process graph (which as a whole indeed  "must accept and return a single value"), but that statement does not reflect that either.

m-mohr: > Superficially: `linear_scale_range` does not accept a single value, it has three required arguments, so this statement looks wrong.

Yes, but as it is mentioned users will acknowledge that this is allowed. Just not mentioning it will make users wonder whether it's allowed to use such processes or not. So I think it's better to list it than not to list it. I agree that it's not ideal though.

> More precisely: you can not pass `abs` or `linear_scale_range` directly to `apply`, you have to embed it in a process graph (which as a whole indeed "must accept and return a single value"), but that statement does not reflect that either.

The process descriptions are aimed towards users, which will use client libraries. There, process graphs are hidden and you can actually directly pass those functions. To improve the situation a bit:

```suggestion
            "description": "A process that accepts and returns a single value and is applied on each individual value in the data cube. The process may consist of multiple sub-processes and could, for example, consist of the processes ``abs()`` or ``linear_scale_range()``.",
```

m-mohr: done.

PR #265, opened 2021-06-14
aggregate_temporal(_period): handling of empty intervals

### LINKED ISSUE #263: aggregate_temporal(_period): handling of empty intervals
One use case for aggregate_temporal is to create data with fixed time intervals. Meaning if I set the period to 'week', I want exactly one observation every week. In practice, this is often combined with e.g. linear interpolation, to fill nodata values in case a given observation was clouded.

However, with the current spec, it is not clear what needs to happen if a given interval doesn't intersect with data in the input datacube? Does the returned datacube have an observation filled with 'nodata' for that interval, or is there no observation at all?

For my current use case, I would like it to be filled with nodata, but this is not necessarily the default case (not sure there).
Do we need a parameter or clarification to pin down the desired behaviour?

m-mohr:
Fixes #263 

Could potentially be part of 1.1.0.


jdries:
Thanks, maybe to be extra clear, we can also expand this sentence:

> For each interval, all data along the dimension will be passed through the reducer.

to

> For each interval, all data along the dimension will be passed through the reducer, or an empty array if the interval is empty.

soxofaan (approved):

sophieherrmann (approved):

m-mohr:
> For each interval, all data along the dimension will be passed through the reducer.

@jdries I think you looked at the original version, not the new proposal. At least this sentence is not in the proposal.

PR #272, opened 2021-07-28
add 2nd example

dthiex:
Adding an example for the "only 1 numerical value available" case


m-mohr:
Sorry, this PR was made from master, not draft. All changes must be made against draft. So I'll cherry-pick the commit instead.

dthiex:
@m-mohr Sorry, for using the wrong base branch and thanks for fixing it. Will be more careful next time!

PR #289, opened 2021-10-25
Clarify save_result

### LINKED ISSUE #288: "Save result" is misleading
The summary of `save_result` is "Save processed data to storage". This can be misleading to users who start synchronous jobs to "just have a look": They don't actually want to _save_ something, especially not to _storage_ -- they just want to _output_ it.

Considering the above use case of "just a look", it's not obvious that you _always_ need this process to actually get results from any process graph.

Looking back it might've been smarter to call this process something like `output`, but I guess for changing the name (especially of such a central process) it's a bit too late now :D

m-mohr:
fixes #288

### DIFF
--- a/save_result.json
+++ b/save_result.json
@@ -1,7 +1,7 @@
 {
     "id": "save_result",
     "summary": "Save processed data",
-    "description": "Saves the processed data to the corresponding medium that is relevant for the context this processes is applied in:\n\n* For **batch jobs** the data is stored in the server-side user workspace of the authenticated user.\n* For **synchronous processing** the data is directly sent to the client.\n* For secondary web services the behavior depends on the service type. Usually, the data is directly provided to the secondary web service, but calling this process may also be rejected by some service types.\n\nThis process aims to be compatible with GDAL/OGR formats and options. STAC-compatible metadata should be stored with the processed data.",
+    "description": "Makes the processed data available in the given file format to the corresponding medium that is relevant for the context this processes is applied in:\n\n* For **batch jobs** the data is stored in the server-side user workspace of the authenticated user. STAC-compatible metadata are usually made available with the processed data.\n* For **synchronous processing** the data is directly sent to the client.\n* For secondary web services the behavior depends on the service type. Usually, the data is directly provided to the secondary web service, but calling this process may also be rejected by some service types.",
     "categories": [
         "cubes",
         "export"
@@ -9,7 +9,7 @@
     "parameters": [
         {
             "name": "data",
-            "description": "The data to output.",
+            "description": "The data to make available in the given file format.",
             "schema": [
                 {
                     "type": "object",
@@ -41,7 +41,7 @@
         }
     ],
     "returns": {
-        "description": "`false` if saving failed, `true` otherwise.",
+        "description": "Reurns `false` if the process failed to make the data available, `true` otherwise.",
         "schema": {
             "type": "boolean"
         }
--- a/save_result.json
+++ b/save_result.json
@@ -1,7 +1,7 @@
 {
     "id": "save_result",
     "summary": "Save processed data",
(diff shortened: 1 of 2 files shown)


save_result.json
```diff
@@ -1,15 +1,15 @@
 {
     "id": "save_result",
-    "summary": "Save processed data to storage",
-    "description": "Saves processed data to the server-side user workspace of the authenticated user. This process aims to be compatible with GDAL/OGR formats and options. STAC-compatible metadata should be stored with the processed data.\n\nCalling this process may be rejected by back-ends in the context of secondary web services.",
+    "summary": "Save processed data",
+    "description": "Makes the processed data available in the given file format to the corresponding medium that is relevant for the context this processes is applied in:\n\n* For **batch jobs** the data is stored in the server-side user workspace of the authenticated user. STAC-compatible metadata are usually made available with the processed data.\n* For **synchronous processing** the data is directly sent to the client.\n* For secondary web services the behavior depends on the service type. Usually, the data is directly provided to the secondary web service, but calling this process may also be rejected by some service types.",
```
soxofaan, line 4: > For **batch jobs** the data is stored in the server-side user workspace of the authenticated user.

I might be missing something, but the concept "workspace" is only/mainly used for the file storage part of the API.
Batch job results are not necessarily stored in a way that's compatible with that file storage API (e.g. in VITO backend, the batch result are not stored in a user specific "home" folder or something like that)

soxofaan: > For **synchronous processing** the data is directly sent to the client

minor note: the "directly" might be misleading because of additional zipping of the result.

soxofaan: > For secondary web services the behavior depends on the service type. Usually, the data is directly provided to the secondary web service, but calling this process may also be rejected by some service types

I'm left confused about this
- "directly provided": does that mean "skipping save_result"?
- sentence starts with "usually .. directly provided ...",  but what is the "otherwise" part? non-directly providing like a batch jobs?
- what if service type requires save_result, but it is missing?

I guess it's not worth trying to explain all the secondary service aspects here? Maybe the description should be very high short and just refer to some general secondary services docs (e.g at https://openeo.org/documentation/1.0/developers/api/reference.html#tag/Secondary-Services)

save_result.json
```diff
@@ -1,15 +1,15 @@
...
     "categories": [
         "cubes",
         "export"
     ],
     "parameters": [
         {
             "name": "data",
-            "description": "The data to save.",
+            "description": "The data to make available in the given file format.",
```
soxofaan, line 12: just an idea: "deliver" instead of "make available". Works both for synchronous and  batch in my head

save_result.json
```diff
@@ -41,7 +41,7 @@
         }
     ],
     "returns": {
-        "description": "`false` if saving failed, `true` otherwise.",
+        "description": "Reurns `false` if the process failed to make the data available, `true` otherwise.",
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
        "description": "Returns `false` if the process failed to make the data available, `true` otherwise.",
```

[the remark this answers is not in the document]
m-mohr: Thanks, fixed. Interesting that the spell check did not catch that 😒

save_result.json
m-mohr: - batch jobs / user workspace: Yes, it's not necessarily the same. We are also removing that language in #292, so this can be aligned. Simply removed it.
- synchronous: Changed to 
  > For **synchronous processing** the data is sent to the client as a direct response to the request
  
- sec. web services: Hmm, is this better? 
  > **Secondary web services** are provided with the processed data so that it can make use of it (e.g., visualize it). Web service may require the data in a certain format. Please refer to the documentation of the individual service types for details.

save_result.json
m-mohr: Works for me, changed.

save_result.json
soxofaan: yes looks good.

re: secondary services: I guess we don't have enough experience with that at the moment to have fully polished docs, so I'm fine with last proposal

soxofaan (approved):

save_result.json
m-mohr: Agreed.

PR #297, opened 2021-11-10
quantiles: Deprecate q in favor of probabilities #293

### LINKED ISSUE #293: quantiles: Unify probabilities and q
Originated in https://github.com/Open-EO/openeo-processes/issues/278:

The quantiles process is old and has the issue that one of two optional parameters is required, which we try to avoid nowadays in openEO (then only array_element would be left). Also, the naming is a bit confusing to some communities.

The proposal is to merge probabilities and q into a single parameter `q`:
```
{
  "name": "q",
  "description": "Quantiles to calculate. Either a list of probabilities or the number of intervals:\n\n* Provide an array with a list of probabilities to calculate quantiles for. The probabilities must be between 0 and 1 (inclusive).\n*Provide an integer to specify the number of intervals to calculate quantiles for. Calculates q-quantiles with equal-sized intervals.",
  "schema": [
    {
      "title": "List of Probabilities",
      "type": "array",
      "items": {
        "type": "number",
        "minimum": 0,
        "maximum": 1
      }
    },
    {
      "title": "Number of Intervals",
      "type": "integer",
      "minimum": 2
    }
  ]
}
```

By itself, this is a breaking change. If we want to change this earlier without breaking anything, we could simply change the schema for q as proposed above and deprecate the parameter probabilities. @soxofaan I think I'd like that...

m-mohr:
Fixes #293


metzm (approved):
I tend to prefer two different options, but in this case there are good arguments to avoid two different options which would be mutually exclusive, at the same time one of them is required, and the two different meanings (probabilities or number of intervals) are very easy to distinguish.

quantiles.json
```diff
@@ -21,20 +21,29 @@
...
+                    "items": {
+                        "type": "number",
+                        "minimum": 0,
+                        "maximum": 1
+                    }
+                },
+                {
+                    "title": "Number of intervals (q-quantiles)",
+                    "type": "integer",
+                    "minimum": 2
```
m-mohr, line 38:
```suggestion
                    "minimumLength": 2
```

mkadunc (approved):

quantiles.json
```diff
@@ -21,20 +21,29 @@
...
         {
             "name": "probabilities",
-            "description": "A list of probabilities to calculate quantiles for. The probabilities must be between 0 and 1 (inclusive).",
-            "schema": {
-                "type": "array",
-                "items": {
-                    "type": "number",
-                    "minimum": 0,
-                    "maximum": 1
+            "description": "Quantiles to calculate. Either a list of probabilities or the number of intervals:\n\n* Provide an array with a list of probabilities to calculate quantiles for. The probabilities must be between 0 and 1 (inclusive).\n* Provide an integer to specify the number of intervals to calculate quantiles for. Calculates q-quantiles with equal-sized intervals.",
```
soxofaan, line 24: Something to discuss or add I think: what should happen if the list of probabilities is not sorted (e.g. user specifies `[0.5, 0.1, 0.9]`?
- throw error
- back-end automatically sorts probabilites and returns 10%, 50% and 90% quantiles in that order
- back-end returns quantiles in order 50%, 10%, 90%

m-mohr: What does make most sense? I guess it's not the error?

soxofaan: Silently sorting the probabilities back-end side is indeed the worst option.

The error option is probably the simplest to start with (in terms of documenting and implementing it), and it can be relaxed later if there is interest in that.

However: `numpy.quantile` apparently supports returning the quantiles in the same order as the probabilities:
 ```
>>> np.quantile(range(100), [0.5, 0.1, 0.9])
array([49.5,  9.9, 89.1])
```
and I guess it's  fairy straightforward to support that anyway in other implementations

m-mohr: For simplicity and interoperability, I'm happy to go with the error if no one else objects. We can then still open it up later if desired.

dthiex (approved):

quantiles.json
m-mohr: Changed, please review.

soxofaan (approved):

Issue #287, opened 2021-10-22
apply_polygon: callback on pixels inside a polygon


jdries:
We received the following use case to support:

- the user has polygons of parcels (agriculture)
- for each parcel, a pixel-level map is expected, (a raster), so a map of the parcel (easy in openEO)
- Now it comes: the timeseries of rasters (4D cube) for a given polygon needs to be filtered on the correlation between different dates. So dates with a correlation that deviates from average correlation need to be filtered out. 

So question is how we could support this?

## Proposal 1 'apply_spatiotemporal'

We already have processes like 'aggregate_spatial' and 'apply_neighbourhood'. So an 'apply_spatiotemporal' process could receive the input polygons, and simply apply a function to the timeseries of pixels within that polygon, receiving a labeled array as input? The output is then again a labeled array with updated pixel values.

In our case, the callback would probably be a UDF.

JeroenVerstraelen:
The suggested process:

_apply_spatial_temporal(code=None, runtime=None, process=None, version='latest', geometries)_
- code (str) – UDF code or process identifier (optional)
- runtime – UDF runtime to use (optional)
- process – a callback function that creates a process graph, see Processes with child “callbacks”
- version (str) – Version of the UDF runtime to use
- geometries (Union[BaseGeometry, dict, str, Path, Parameter]) – shapely geometry, GeoJSON dictionary or path to GeoJSON file

The UDF function is applied per geometry (= parcel).
Example UDF code:
```
def apply_datacube(cube: XarrayDataCube, context: dict):
  # Cube = tile(s) that intersect the geometry (masked to only include pixels in geometry)
  # Cube is masked, pixel is NaN if not in geometry.
  inarr=cube.get_array()
  dates = ['2021-10-22', '2021-10-23', '2021-10-24', '2021-10-25']
  results = {}
  for date in dates:
    field = inarr.loc[date,:]
    result[date] = some_computation(field)
  correlation_matrix(results)
  
  return cube
```

@m-mohr Is the name of the process in line with the rest of the API, or should it be one of:
- apply_spatiotemporal
- apply_spatio_temporal
- apply_spatialtemporal

m-mohr:
I'm happy to provide a process description for this. Please keep in mind that run_udf and the new process would probably not be joined like in your example. While you can do that in Python the openEO processes would be more generic. I'll try to get a draft PR up today. I do not understand everything regarding this process yet though, so the proposal may be somewhat off, but can get us started better.

[...]

PR #298, opened 2021-11-10
Add apply_polygon #287

m-mohr:
A proposal to discuss for issue #287.

(Not needed in deliverable)


JeroenVerstraelen (approved):

proposals/chunk_polygon.json
```diff
@@ -0,0 +1,80 @@
...
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "chunks",
+            "description": "A GeoJSON object containing at least one polygon. The provided feature types can be one of the following:\n\n* A `Polygon` or `MultiPolygon` geometry,\n* a `Feature` with a `Polygon` or `MultiPolygon` geometry,\n* a `FeatureCollection` containing at least one `Feature` with `Polygon` or `MultiPolygon` geometries, or\n* a `GeometryCollection` containing `Polygon` or `MultiPolygon` geometries. To maximize interoperability, `GeometryCollection` should be avoided in favour of one of the alternatives above.",
+            "schema": {
+                "type": "object",
```
m-mohr, line 22: We may add vector-cube as a type here.

proposals/chunk_polygon.json
```diff
@@ -0,0 +1,80 @@
...
+                "subtype": "geojson"
+            }
+        },
+        {
+            "name": "process",
+            "description": "A process that accepts and returns a single data cube and is applied on each individual chunk. The process may consist of multiple sub-processes.",
```
soxofaan, line 28: > The process may consist of multiple sub-processes.

Isn't this a bit redundant? It's basically the definition of a process.

soxofaan: I think there should also be a note that the process should preserve the original dimensions (so no adding/dropping of dimensions, no type changes, ...). 

I'm not sure it should be allowed or forbidden to also change labels, cardinality, resolution,

m-mohr: 1) Yes for us "pros" that is clear, but we have that in all(?) callback definitions afaik. The reason is that at some point we had people asking whether they could only provide a single process (e.g. "mean") here or more.

2) Good point. I'm not actually sure what the intention here was originally, so forwarding this question to @JeroenVerstraelen and @jdries ...

proposals/chunk_polygon.json
m-mohr: I'll do this separately in one go for all processes.

m-mohr:
Sad to see that there was no communication in the PR to further discuss and add this to this PR during the implementation phase. Now, I can't guarantee that the definition here will be exactly what you expose and we may diverge.

JeroenVerstraelen:
The mask_value parameter was requested by the internal VITO member who asked for this process. 

The issue was that a UDF receives data for a rectangular area, but geometries can have any shape. So any pixels not in the geometry had to be masked out with a value provided by the user.

m-mohr:
This proprietary parameter will likely not become part of the spec though as this usecase is already covered by mask_polygon. I'd suggest using that instead of a proprietary addition.

m-mohr:
I've just added a definition of mask_value to chunk_polygon. Please review.

soxofaan:
It's probably a bit late to raise this, 
but while working on documentation it appeared to me that the name "chunk_polygon" does not align well with comparable processes like `apply`, `apply_dimension`, `apply_neighborhood`, `reduce_dimension`. ...  which contain a verb like "apply" or "reduce" that make it self-documenting.
I think chunk_polygon belongs to the "apply" family, but that is not obvious: usage-wise the process does not really "chunk" your data (in the sense that it would return an array of chunks or something like that);  it "applies" a child process to chunks (internally), but still returns a single data cube.
Is there still opportunity to rename to, e.g. "apply_chunks" , "apply_chunked" or "apply_polygon"?

m-mohr:
@soxofaan I think you need to raise this internally, because I think only VITO itself is implementing it.

soxofaan:
indeed, I should have cc'ed @jdries .  
Still, as it's a naming/allignment thing, it doesn't hurt to discuss this publicly too.

m-mohr:
Sure, but what I meant is whether it's fine for you too change at all. It VITO says you don't want to change for backward compatibility, it doesn't make a lot of sense to discuss it here, too.

jdries:
The name can still change, we know most of the cases and persons that use this, and can always keep an alias for backwards compatibility.
Note that the process is quite popular, it's used by:
- electrical conductivity service
- water reservoir mappoing
- geo informed

m-mohr:
Okay, sounds good.

Any preferences? Some thoughts:
- This process is either a variant of apply or mask.
- `apply_chunked` and similarly generic names don't reflect the spatial nature. How would we call a chunking process over e.g. the temporal dimension?
- `apply_polygon` would be a short variant that hopefully implies chunking
- `apply_chunks_polygon` could be a bit more clear, but it a mouthful...

soxofaan:
I also think  `apply_polygon` is the best candidate. It aligns very well with `apply_neighborhood` in terms of the chunking aspect

m-mohr:
Renamed. Would be really nice to finally merge this so requesting reviews again: @jdries @soxofaan @JeroenVerstraelen

jdries:
Looks good, is there a need to say something about non-spatial dimensions? Do we need to mention explicitly that chunking happens along the x,y dimensions, and all others are passed on fully into the 'chunk'?

jdries:
@JeroenVerstraelen  can you add your review?

m-mohr:
@jdries That's probably a good clarificaton to add.

Another question that is still open is https://github.com/Open-EO/openeo-processes/pull/298#discussion_r822546408 @JeroenVerstraelen

proposals/apply_polygon.json
```diff
@@ -0,0 +1,100 @@
+{
+    "id": "apply_polygon",
+    "summary": "Apply a process to spatial chunks of a data cube",
+    "description": "The given data cube is chunked by the given polygons and applies the given process to each individual chunk.",
```
LukeWeidenwalker, line 4: This is a nitpick, but as a dask user the word "chunk" is overloaded for me. In dask, it would be possible that two of these polygons live within the same chunk. Also dask chunks are rectangular, which geometries are not necessarily. How about something like "The provided process is applied only to the subregions of the data cube described by the provided polygons."?

LukeWeidenwalker (commented):
One nitpick and one general comment: 
Is there anything this process should say about what happens with overlapping geometries, or is handling of this entirely the business of backends? I guess since there's no focal aspect to this process, the results will be the same regardless of whether geometries overlap, and it's thus only a potential performance optimisation?

proposals/apply_polygon.json
soxofaan: Indeed, with the rename to `apply_polygon` we might want to steer away from using "chunk" all together

proposals/chunk_polygon.json
JeroenVerstraelen: @m-mohr Our implementation assumes that no dimensions can be added/dropped and that the (#rows, #cols) of the input is identical to that of the output. But you can add/drop individual dates or bands.

As it was originally requested in https://github.com/Open-EO/openeo-processes/issues/287, the user needed to be able to view a 4D cube per polygon and create a correlation matrix based on some computation for each date.

I believe that adding/dropping dimensions falls outside of the scope of this process but I think that the preservation of the spatial dimensions can be discussed. Taking into account this documentation for apply https://openeo.org/documentation/1.0/datacubes.html#apply the fact that we call the function 'apply_polygon' could mean:
1. Identical to apply_neighbourhood, apply_kernel: apply a function over every pixel in the datacube using a polygon as a kernel (wrong)
2. Identical to apply_dimension: apply a function over every pixel using the polygon as the input (right)

(In our case we provide a list of polygons so apply_polygons might be an even better name.) So we can use the exact same description as we use for the _apply_ process, it is just that we divided it into an apply for every polygon instead of the whole datacube. The return value of the apply process is:
_A data cube with the newly computed values and the same dimensions. The dimension properties (name, type, labels, reference system and resolution) remain unchanged._

Given that description I think our VITO implementation already permits too much by allowing to add/drop dates or bands.

proposals/apply_polygon.json
```diff
@@ -0,0 +1,100 @@
...
+        {
+            "name": "data",
+            "description": "A data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "chunks",
```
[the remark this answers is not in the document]
m-mohr: Good idea, will rename.

proposals/apply_polygon.json
m-mohr: Hmm... happy to optimize the wording, but "Apply a process to each given polygon in the datacube." sounds for me like you apply something on a vector data cube. I'll think about an alternative...

JeroenVerstraelen: Maybe it was also nitpicky from me. Given that we want to emphasize that the polygons are cut out of the input data cube first and then a process is applied to each polygon the longer sentence is probably better.

proposals/apply_polygon.json
```diff
@@ -0,0 +1,100 @@
...
+            "description": "Additional data to be passed to the process.",
+            "schema": {
+                "description": "Any data type."
+            },
+            "optional": true,
+            "default": null
+        }
+    ],
+    "returns": {
+        "description": "A data cube with the newly computed values and the same dimensions. The dimension properties (name, type, labels, reference system and resolution) remain unchanged.",
```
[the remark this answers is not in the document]
m-mohr: Does the general process description that I just updated help to clarify this?

proposals/chunk_polygon.json
m-mohr: So is it fair to update the description of the returned sub data cube to

> The updated sub data cube with the newly computed values and the same dimensions. The dimension properties (name, type, reference system and resolution) must remain unchanged. The labels can change, but the number of labels must remain unchanged.

m-mohr:
I've updated the descriptions of the process according to the comments and also added vector data cube support. Please let me know what you think, I hope we can approve this soon.

JeroenVerstraelen (approved):
Looks good to me!

m-mohr:
@LukeWeidenwalker @clausmichele @dthiex I'd appreciate your final reviews on this PR.

m-mohr:
I'll merge this EOB on Tuesday if no other comments come in.

LukeWeidenwalker:
One question still open from me: What's the expected behaviour for pixels that intersect multiple of the provided geometries?

m-mohr:
@LukeWeidenwalker Hmm... good question. Open to suggestions. As it is hard to actually get reproducible results if you run over an area multiple times, should we throw an error if overlapping geometries are provided?

LukeWeidenwalker:
> @LukeWeidenwalker Hmm... good question. Open to suggestions. As it is hard to actually get reproducible results if you run over an area multiple times, should we throw an error if overlapping geometries are provided?

I think throwing an error is the best option here, doing this multiple times over an area just seems ill-defined to me!

m-mohr:
Okay, I've added an exception.

jdries:
> Okay, I've added an exception.

If polygons overlap, I believe the function is still invoked once per polygon and pixels are masked according to the polygon that is being handled. Not sure if this is ill defined and requires an exception? @JeroenVerstraelen

m-mohr:
I assumed @JeroenVerstraelen thumbs up to Lukas' comment meant agreement.

m-mohr:
Great, thanks. So I'll merge this now. If there's further room for improvement, please open a new issue and/or PR.

Issue #300, opened 2021-11-12
load/save_model for ML models


m-mohr:
Propose two processes load_model and save_model to load and store ML models. Should be similar to load_results with loading from job id and URL. The input could be STAC items described with the (upcoming) ml-model extension: https://github.com/stac-extensions/ml-model

jdries:
I guess this is the metadata spec that we're referring to:
https://github.com/radiantearth/geo-ml-model-catalog
?

m-mohr:
The Geo-ML-Model-Catalog was the first attempt from Radiant to write this specification *not* based on STAC, but this evolved somehow into a work-in-progress STAC extension: https://github.com/stac-extensions/ml-model
I'll check with Jon what the long-term plan is with the Geo-ML-Model-Catalog repo is.

m-mohr:
> The STAC extension one is now the main repo. The GMLMC one will be archived after we [have] transfer[ed] everything.

m-mohr:
@jdries @aljacob Do you have an idea what metadata would be minimally required for our use case?

jdries:
Really minimally, the model just needs a name. But we'll probably also specify the type of model file, so we can determine if it's compatible with our software.

m-mohr:
Thanks and good to know, but I guess I should have asked what you need minimally (so that I can ensure this is at least available) and what other information would be useful in addition? @jdries 

From what you have listed right now, we may not really need a model extension yet.
- The name is either the STAC id and/or title.
- The generating software could be exposed via the processing extension although you'd run into issues if the software can generate multiple types so maybe a model file type would need to be part of a new extension. Looking at standardization: How could we ensure that model types match between implementations? Like is there a common name for each model type that we could refer to? Otherwise, implementations may choose different type names although they refer to the same thing.

PR #304, opened 2021-11-16
Load/Save ML model #300

m-mohr:
Draft for issue #300.


m-mohr:
In the dev telco the idea came up to integrate loading and/or saving into the ML processes. If they are tightly coupled and you can't do more things with ML models, we may actually want to consider this to make the user experience better. In this case these processes would be merged with the ML processes. I don't feel like I have enough experience to decide so would appreciate thoughts.

proposals/load_ml_model.json
```diff
@@ -0,0 +1,44 @@
...
+            "schema": [
+                {
+                    "title": "URL",
+                    "type": "string",
+                    "format": "uri",
+                    "subtype": "uri",
+                    "pattern": "^https?://"
+                },
+                {
+                    "title": "Batch Job ID",
```
m-mohr, line 23: How to specify the model to load (if there are multiple per job)?

m-mohr:
Meeting conclusion: Have load/save specifically as process to seperate concerns and be more flexible in the future. Also, makes the processes a bit simpler.

proposals/load_ml_model.json
m-mohr: > Loading a model by batch job ID is possible only if a single model has been saved by the job. Otherwise, you have to load a specific model from a batch job by URL.

proposals/load_ml_model.json
```diff
@@ -0,0 +1,53 @@
...
+                    "description": "Loading a model by batch job ID is possible only if a single model has been saved by the job. Otherwise, you have to load a specific model from a batch job by URL.",
+                    "type": "string",
+                    "subtype": "job-id",
+                    "pattern": "^[\\w\\-\\.~]+$"
+                },
+                {
+                    "title": "User-uploaded Files",
+                    "type": "string",
+                    "subtype": "file-path",
+                    "pattern": "^[^\r\n\\:'\"]+$"
```
soxofaan, line 33: a path to user uploaded file is always a relative path I guess, so shouldn't the regex specify that it can't start with `/`?

proposals/save_ml_model.json
```diff
@@ -0,0 +1,44 @@
...
+            "schema": {
+                "type": "object",
+                "additionalParameters": false
+            },
+            "default": {},
+            "optional": true
+        }
+    ],
+    "returns": {
+        "description": "Returns `false` if the process failed to store the model, `true` otherwise.",
```
soxofaan, line 31: IMHO this is a bad design pattern: if the storage failed, an error should be thrown, 
it's just too error prone to return a boolean nobody is going to check.

I think it's cleaner to return nothing (`null`) or return the original `data`

soxofaan (commented):
some minor notes

proposals/load_ml_model.json
m-mohr: No, the root is the user workspace. /a.txt is simply a file in the root folder of the user's workspace. That's equivalent to a.txt and ./a.txt as the cwd is also the root.

soxofaan: The API already defines these constraints that it must be relative, and `.` should be avoided:

https://github.com/Open-EO/openeo-api/blob/f303d65a3291d4cd74dacc0e796803bb5d6fa03b/openapi.yaml#L3283-L3288
https://github.com/Open-EO/openeo-api/blob/f303d65a3291d4cd74dacc0e796803bb5d6fa03b/openapi.yaml#L5668-L5680

m-mohr: That's a misunderstanding: In OpenAPI we define the path in the URL, in the processes we define the path from the user-POV, which are slightly different use cases. A path should not start with a / in a URL as then you could end up with `https://example.com/files//a.txt`.

proposals/save_ml_model.json
m-mohr: Yeah, this was copied over from save_result. The same also applies there and the boolean was not the wisest thing, it should indeed throw an error and I assume most implementations will actually do that, right?

m-mohr: Created a separate issue: https://github.com/Open-EO/openeo-processes/issues/334

m-mohr:
Merging, please open a separate issue or PR to discuss further changes.

PR #307, opened 2021-11-23
Add run_process

m-mohr:
Proposal for solving https://github.com/Open-EO/openeo-api/issues/413 in a non-breaking way.


jdries:
See comment on other issue: currently not in favor of this as a solution to avoid allowing parameters for process_id.

dthiex (approved):
Changes look okay to us.

soxofaan:
a devil's advocate argument against this process: I think you lose the ability to statically check/validate  your process graph because the output schema of run_process is only known at run time.

This might be annoying in several places:  visualization of process graph (e.g. in web editor), in clients that use output schemas to help building a process graph , the `/validation` end point,  the dry-run procedure in the geopyspark driver, ...

m-mohr:
> This might be annoying in several places: visualization of process graph (e.g. in web editor), in clients that use output schemas to help building a process graph , the `/validation` end point, the dry-run procedure in the geopyspark driver, ...

Yes, but checking it in the JS tooling shows that this is equally annoying as allowing variable for `process_id`. It is basically the same implementation-wise, just a different encoding.

m-mohr:
Closing for now until we actually want to tackle this again.

PR #309, opened 2021-11-29
Array modify: doc finetuning

soxofaan:


soxofaan:
Related: 
To me,  default `length=1`  seems pretty arbitrary: you replace one element, regardless of the size of `values`.
Also: if you use it to append to the array, you specify `index=$array_size` and `length=1` is then actually invalid because it is an out of bound reference.


I'm not sure how clean it is to change this process at this point (it's still under "proposals" though),
but I would propose default `length=0` so that you just insert values by default, without losing original values.

(Another option would be default `length=null`, with new behavior: automatically use the size of `values`)

soxofaan:
Another thing:
I think there is a contradiction between:

> If the index is greater than the number of elements, the process throws an `ArrayElementNotAvailable` exception.

and the example

> "Add a value at a specific non-existing position after the array, fill missing elements with `null`.",
> array_modify(data = ["a","b"], values = ["e"], index = 4) => ["a","b",null,null,"e"]

m-mohr:
The process has not been implemented yet (according to the Hub) and it's still in proposals. As such I think we can still change it.

The default for length comes from the update/delete use-case, where a common one is to simply replace/delete a single value. I'll think about this once more and check common use cases against the proposed values 0 and null.

Yes, the given example "Add a value at a specific non-existing position ..." is a left-over from an old version where those out-of-bounds indices were still allowed. We should remove that. Can you incorporate it into this PR?

proposals/array_modify.json
```diff
@@ -19,7 +19,7 @@
         },
         {
             "name": "values",
-            "description": "The values to fill the array with.",
+            "description": "The values to insert into the `data` array.",
```
m-mohr, line 22: I've used fill here as insert into sounded a bit too much like "add", but "update" is also allowed. But if you see an update as insert & remove, it works...

proposals/array_modify.json
```diff
@@ -57,7 +57,7 @@
     },
     "exceptions": {
         "ArrayElementNotAvailable": {
-            "message": "The array has no element with the specified index."
+            "message": "An invalid insert index is specified."
```
m-mohr, line 60: I don't think this improves the message. Maybe:

```suggestion
            "message": "The array can't be modified as the given index doesn't exist."
```

proposals/array_modify.json
soxofaan: `fill` sounds like overwriting the whole array, so I wanted to fix that. 

First I also had `.. insert/replace into ..`  but I'm not sure that's proper English

proposals/array_modify.json
soxofaan: First of all some context why I changed it:  if your given array has N elements, the range of valid indexes for `array_modify` is 0 to N (inclusive), so N+1 index values in total.  Index `N` is valid for `array_modify`, while there is no element there. The original message is bit confusing about that.

> I initially had "or negative" appended, but that should already be captured by the schema validation (minimum set to 0).

Yeah, but we want to throw that error `ArrayElementNotAvailable`  also on that schema violation. Or should we throw a generic  `ProcessParameterInvalid` ?

proposals/array_modify.json
```diff
@@ -29,15 +29,15 @@
...
+            "description": "The index in the `data` array of the element to insert the value(s) before. If the index is greater than the number of elements in the `data` array, the process throws an `ArrayElementNotAvailable` exception.\n\nTo insert after the last element, there are two options:\n\n1. Use the simpler processes ``array_append()`` to append a single value or ``array_concat()`` to append multiple values.\n2. Specify the number of elements in the array. You can retrieve the number of elements with the process ``count()``, having the parameter `condition` set to `true`.",
             "schema": {
                 "type": "integer",
                 "minimum": 0
             }
         },
         {
             "name": "length",
-            "description": "The number of elements to replace. This parameter has no effect in case the given `index` does not exist in the array given.",
+            "description": "The number of elements in the `data` array to replace. This parameter has no effect in case the given `index` does not exist in the array given.",
```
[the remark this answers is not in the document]
soxofaan: hmm yeah, that is still confusing, especially the "exceeding elements" thing.

Maybe the whole modify operation should be explained in two steps:
1. first remove `length` elements at index `index` 
2. then insert `values` before index `index`

trying to explain as one atomic operation will probably be confusing

soxofaan:
> About the case where index=$array_size: This also applies if you give a index + length that exceed the bounds. We may need to clarify that the length is "clipped"/limited to the actual bounds.

I was not aware of this auto-clipping feature. I didn't "get" that from the docs

proposals/array_modify.json
m-mohr: I'm not strictly against adding "or negative" again, just found it to be superfluous. So, yes, consistent would be if values that fails due to schema restrictions would return a ProcessParameterInvalid and runtime specific issues would return a specific error (here: ArrayElementNotAvailable).

proposals/array_modify.json
m-mohr: Hmm, I think I came to the conclusion that insert into somewhat works if we think about this generally as remove & insert operations. We may want to make this more explicit in the process overall.

proposals/array_modify.json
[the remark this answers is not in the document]
m-mohr: Yeah, maybe let's revamp the whole process a bit and tell people that "update" can be seen as a "remove & insert". That maybe makes the process overall better to understand and also this easier to describe.

soxofaan:
> What was your expectation of what would happen?

I'm not sure. As a python dev I think I would expect silent clipping. As a C/Java dev I would expect some kind of overflow error.

proposals/array_modify.json
m-mohr: I've changed to the longer error message without "or negative", but was not really sure whether to add it or not. Open for changes.

proposals/array_modify.json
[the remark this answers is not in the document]
m-mohr: I made some changes that should better describe the process as "remove & insert" for update and hopefully makes it more clean. Thoughts?

m-mohr:
> I'm not sure. In general (regardless of docs), as a python dev I think I would expect silent clipping. As a C/Java dev I would expect some kind of overflow error.

I updated the descriptions again and follow the clipping approach, although I'm not set on it and also happy to specify that an error should be thrown.

> However, the docs currently say about `length`
> 
> > The number of elements to replace.
> 
> With this definition I would throw an error if that number of elements is not available.

It is better with the new descriptions? @soxofaan

proposals/array_modify.json
```diff
@@ -181,6 +161,22 @@
                 "b",
                 "c"
             ]
+        },
+        {
+            "description": "Remove multiple values from the end of the array and ignore that the given length is exceeding the size of the array.",
```
m-mohr, line 166: Not sure whether this is a good example. It's more a unit test than a valuable example to users, so we can also remove it if desired.

soxofaan: I'm fine with this example it illustrates removing with inserting and the clipping feature

proposals/array_modify.json
```diff
@@ -29,15 +29,15 @@
...
+            "description": "The index in the `data` array of the element to insert the value(s) before. If the index is greater than the number of elements in the `data` array, the process throws an `ArrayElementNotAvailable` exception.\n\nTo insert after the last element, there are two options:\n\n1. Use the simpler processes ``array_append()`` to append a single value or ``array_concat()`` to append multiple values.\n2. Specify the number of elements in the array. You can retrieve the number of elements with the process ``count()``, having the parameter `condition` set to `true`.",
             "schema": {
                 "type": "integer",
                 "minimum": 0
             }
         },
         {
             "name": "length",
-            "description": "The number of elements to replace. This parameter has no effect in case the given `index` does not exist in the array given.",
+            "description": "The number of elements in the `data` array to remove (or replace) at the given index. If there are not enough elements to remove, the process simply removes all elements after the given index.",
```
soxofaan, line 40: "not enough elements to remove" sounds a bit strange to me

proposal:

> If the array contains less elements starting from the given index, the process simply removes up to the end.

proposals/array_modify.json
```diff
@@ -29,15 +29,15 @@
...
             "optional": true,
             "default": 1,
```
soxofaan, line 42: As mentioned above (but probably lost in the discussion), I think a default `length=1`  is unnecessarily opinionated.

I would propose to use 0 by default, or not having a default at all

proposals/array_modify.json
```diff
@@ -1,15 +1,15 @@
 {
     "id": "array_modify",
     "summary": "Change the content of an array (insert, remove, update)",
-    "description": "Allows to insert into, remove from or update an array.\n\nAll labels get discarded and the array indices are always a sequence of numbers with the step size of 1 and starting at 0.",
+    "description": "Allows to insert into, remove from or update an array. This process describes updating an array as two separate operations that first removes and then inserts elements.\n\nAll labels get discarded and the array indices are always a sequence of numbers with the step size of 1 and starting at 0.",
```
soxofaan, line 4: > "summary": "Change the content of an array (insert, remove, update)",
>  "description": "Allows to insert into, remove from or update an array. This process describes updating an array as two separate operations that first removes and then inserts elements.

I think there is some room for finetuning here (related to defining update as remove+insert), e g.

> "summary": "Change the content of an array (remove, insert, update)",
> "description": "Modify an array by removing, inserting or updating items. Updating can be seen as removing items followed by inserting new items (not necessarily the same number).

proposals/array_modify.json
m-mohr: Isn't "less" wrong here? At least it sounds weird. I've mixed things up and came up with:

> The number of elements in the `data` array to remove (or replace) starting from the given index. If the array contains too few elements to remove, the process simply removes all elements up to the end

proposals/array_modify.json
m-mohr: I had a look but forgot to reply, sorry. I couldn't find a less opinionated alternative yet assuming that the main usecases are update and remove with having array_append and array_concat. I don't feel like there are a lot of common use cases for inserting in-between elements, but length = 0 would mostly cater for them. Length = null (i.e. length of values) would be okay, I think, but seems to make descriptions more complex. So if replacing length = 1 with length = null seems important enough for a change, then that would be my proposal.

proposals/array_modify.json
m-mohr: Updated, with the minor change that I used "elements" instead of "items" to be consistent.

m-mohr (approved):
The main goal here was to fine-tune docs, which I think has been achieved. For the change to the default value for length, we should probably open a new issue?!

proposals/array_modify.json
soxofaan: I mean "less than `length`"

My main issue is just that "too few" or "not enough" sound like there is something wrong with the input (like there are wrongly fewer elements than expected)

soxofaan:
It started with the idea to "just finetune" the docs yes :stuck_out_tongue:

proposals/array_modify.json
m-mohr: It reads better if I include the sentence before, indeed. So changed to ~less~ fewer:

>The number of elements in the `data` array to remove (or replace) starting from the given index. If the array contains fewer elements, the process simply removes all elements up to the end.

proposals/array_modify.json
soxofaan: I don't think that the existence of `array_append` and `array_concat` is a reason to arbitrary pick default `length=1`. 

One could argue that default `length=0` is also opinionated, but at least it is the least destructive option (any length>0 removes data)

Another defense for in favor for the `length=0`  default is that the most obvious default array  is an empty array (length 0).

m-mohr: Opened a new issue: #312

Issue #264, opened 2021-06-14
Buffer process


m-mohr:
Request from VITO: A buffer process.

Something like `buffer(data:vector-cube, distance:number, unit:string="meters")`

mkadunc:
IMO it would make more sense to have a more atomic operation:
- buffer for single geometry: `buffer(geojson:object, distance:number, unit:string="meters")` to be then mapped over an array, or
- buffer for array of geometries: `buffer(array[geojson:object], distance:number, unit:string="meters")`

Unless we've settled on `vector-cube` as the way to represent (arrays of) geometries in openEO, in which case 💯👍 .

m-mohr:
> * buffer for single geometry: `buffer(geojson:object, distance:number, unit:string="meters")` to be then mapped over an array, 

What would this actually return so that it's useful? GeoJSON again? A raster cube (i.e. a mask)?

> Unless we've settled on `vector-cube` as the way to represent (arrays of) geometries in openEO, in which case 💯👍 .

I think we should finally settle on something. The next step is probably to write a PSC proposal for vector cubes and then start defining processes based on it?!

jdries:
The return type should be the same as the input vector type.

As for the input type: if geojson implies that it can only be in EPSG:4326, then maybe it's better to define a proper, more abstract, vector-cube.
For me, a vector cube should probably be something like a feature collection at this moment, and not define too much properties yet, to allow for some evolution.

m-mohr:
> The return type should be the same as the input vector type.
> As for the input type: if geojson implies that it can only be in EPSG:4326, then maybe it's better to define a proper, more abstract, vector-cube.

So that would be GeoJSON then, which is EPSG:4326 only, yes. 
https://github.com/opengeospatial/ogc-feat-geo-json (which is adding CRS to GeoJSON) is just staring and thus not an option in the near future. Also, GeoJSON doesn't really scale well for very large collections.

edzer:
> So it seems reasonable to write up something about vector cubes...

[...]

PR #314, opened 2021-12-17
Add vector_buffer

m-mohr:
Adds a new process "vector_buffer", see #264 

If I understand https://openeo.vito.be/openeo/1.0/processes/backend/vector_buffer correctly (some ambiguities are present), it should be possible to merge both in a way that there's no conflict.

Implementations looked at:
- https://r-spatial.github.io/sf/reference/geos_unary.html
- https://openeo.vito.be/openeo/1.0/processes/backend/vector_buffer
- https://postgis.net/docs/ST_Buffer.html
- https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.buffer.html
- ( https://developer.ibm.com/learningpaths/data-analysis-using-python/working-with-geospatial-vector-data-in-python/#buffer )
- https://developers.google.com/earth-engine/apidocs/ee-feature-buffer

Question:
- [x] How to handle the number of segments per quadrant? -> leave up to implementation, be tolerant in test suite
- [x] Should we allow negative distances? -> seems useful


mattia6690 (approved):
I think that a negative buffer is very useful for many researchers to either exclude or specifically research areas at the boundaries of Polygons or in the vicinity of points and lines.
Otherwise the current structure is fine

jdries:
We certainly use negative buffer distances.
Not sure about segments per quadrant, seems like a parameter that you can leave to the implementation?
Seems fine otherwise, hope we don't need to mathematically specify what a rounded buffer is.

m-mohr:
> Not sure about segments per quadrant, seems like a parameter that you can leave to the implementation?

@jdries I assume that would be an issue in the process validation suite if the implementations differ in the number of segments.

jdries:
> I assume that would be an issue in the process validation suite if the implementations differ in the number of segments.

Yes, when doing vector operations on the sphere, quite a few things can go wrong and need to be specified if you want absolute accuracy. So you would need to start by defining your error tolerances. Or make it dynamic, because if vector data is combined with raster, then there's no need to go far beyond the resolution of your raster cubes.
Another example is vector-pixel intersection tests, which you use in aggregate_spatial.

So my experience is that the validation suite will need to have some built-in leniency as to what the error can be for a specific process.

m-mohr:
Okay, happy to keep it simple (except in the test suite). Then we can (until requested by users) define the process without such a parameter. Negative distances seem useful so then this is ready for review.

clausmichele (approved):

proposals/vector_buffer.json
```diff
@@ -0,0 +1,59 @@
+{
+    "id": "vector_buffer",
+    "summary": "Compute a buffer around geometries",
+    "description": "Calculates a round buffer around each geometry given. Buffers can be applied to points, lines and polygons, but the results are always polygons. Multi-part types (e.g. `MultiPoint`) are also allowed.",
```
soxofaan, line 4: Having a non-GIS background, I found the term "buffer" a bit strange at first. I think it would be good if the description contained other keywords too like "morphological operation", "dilation"(="outward buffer"), "erosion" ("inward buffer")

proposals/vector_buffer.json
```diff
@@ -0,0 +1,59 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "distance",
+            "description": "The radius of the buffer in the unit given in the `unit` parameter. A negative number can be provided to subtract from geometries.",
```
soxofaan, line 26: > in the unit given in the `unit` parameter.

"given by the ..."?

soxofaan: > A negative number can be provided to subtract from geometries.

A positive distance results in outward buffering (dilation), a negative distance results in inward buffering (erosion).

soxofaan: > The radius of the buffer

Maybe better to keep using "distance" instead of introducing "radius"?

proposals/vector_buffer.json
m-mohr: Sure, that would be appreciated, I think. Would you mind providing a proposal?

proposals/vector_buffer.json
m-mohr: Sounds all good. Fixed.

I also agree with distance after reading https://stackoverflow.com/a/70151787/9709414

proposals/vector_buffer.json
m-mohr: Okay, I've added something. What do you think?

proposals/vector_buffer.json
```diff
@@ -0,0 +1,59 @@
+{
+    "id": "vector_buffer",
+    "summary": "Buffer geometries by distance",
+    "description": "Buffers each input geometry by a given distance, which can either expand (dilate) or a shrink (erose) the geometry. Buffers can be applied to points, lines and polygons, but the results are always polygons. Multi-part types (e.g. `MultiPoint`) are also allowed.",
```
mattia6690, line 4: erose -> erode

proposals/vector_buffer.json
```diff
@@ -0,0 +1,59 @@
...
+            "default": null,
+            "schema": [
+                {
+                    "title": "Unit of the SRS",
+                    "type": "null"
+                },
+                {
+                    "title": "Meters",
+                    "type": "string",
+                    "const": "m"
```
mattia6690, line 47: Correct me if am wrong but isn't it better to always use the units of the current CRS there? Otherwise I assume that you'd need to do a reprojection or some other kind of conversion on the fly, right? Is that planned?

soxofaan: I guess the reasoning is that GeoJSON only strictly supports epsg:4326, while you usually want to express the buffer in meters?

clausmichele: Exactly, geoJSON officially supports only epsg:4326. Having also meters would imply to convert it to degrees, but we should inform the user that it's not a perfect conversion.

proposals/vector_buffer.json
m-mohr:
```suggestion
    "description": "Buffers each input geometry by a given distance, which can either expand (dilate) or a shrink (erode) the geometry. Buffers can be applied to points, lines and polygons, but the results are always polygons. Multi-part types (e.g. `MultiPoint`) are also allowed.",
```

proposals/vector_buffer.json
m-mohr: Well, this also supports vector-cube, which are not restricted to GeoJSON and EPSG4326. So @mattia6690 is raising a valid point here. Until now, we have two different approaches in processes:
1. in resample_spatial it is based on the unit of the CRS: "Specified in the units of the target projection."
2. in apply_neighborhood you can either specify "m" or "px" as the unit for the size.

What is preferable here? The easiest solution would probably be 1.

soxofaan: We've been doing geometry buffering in a couple of VITO use cases already, and I think the most natural way for the user to specify this is in meter  (e.g. inward buffering to eliminate edge issues). Right, @jdries ?
So I think option 2 makes most sense.

Option 1 assumes that your have a target data cube and that its CRS is sensible to express the buffer distance in, which is not necessarily try

jdries: I'm afraid it's even more difficult.
A vector cube normally has a projection system associated with it, which can be geodetic (lat lon) or specified in meters.
The meters of a projected coordinated system however, do not necessarily correspond to meters on earth.
IMO the most logical buffering operation indeed uses meters, but on the earth. This is however not the easiest variant.

You could also consider supporting the other variants, but then we indeed need the necessary properties to differentiate.

m-mohr: Hmm, so I guess the "easiest" way is to just redirect these issues to the underlying libraries? It seems most rely on the "unit of the spatial reference system, so we can probably just start with that? Updated the process accordingly.

soxofaan: > just redirect these issues to the underlying libraries? It seems most rely on the "unit of the spatial reference system, so we can probably just start with that? 

I don't think that is very practical. E.g. if you start from GeoJSON, which is strictly lat-lon, you have to specify the buffer in degrees, which no user probably wants to do in a real use case. We also don't have a vector reprojection process, so you're stuck.

m-mohr: Right now every process does something differently, some use meters, some degrees, some "unit of the srs", etc. We need to commit to a consistent behavior, I think.

m-mohr:
Merging as this seems to be finished, except for the general unit discussion in #330.

Issue #313, opened 2021-12-13
Sampling


m-mohr:
For several use cases such as ML training, one or more sampling processes could be useful. 

This came up in the discussions around #295 

Related work: 
- https://eo-learn.readthedocs.io/en/latest/eolearn.geometry.sampling.html
- https://developers.google.com/earth-engine/apidocs/ee-image-sample?hl=en

Alternative: 
- polygon_to_point?
- #155

clausmichele:
The sampling process could also allow to "extract" all the valid points setting a parameter. This would cover the case where we select some polygons (filter_spatial) and we want to keep all the pixels inside of them for training.

Some questions that come up in my mind:
1. Is the output a vector-cube?
2. Does it keep x,y information about the points or do we discard them?

clausmichele:
The sampling will be performed using a new process called polygon_to_points:
- polygon_to_points takes as input a geoJSON (or other supported formats, TBD) and returns the same structure with point coordinates
- possible parameters are:
1. Sampling strategy: random, fibonacci, ...
2. Max number of points to extract
3. Other?

m-mohr:
* We still need to verify which name to choose, they should be somewhat aligned between vector processes.
* Input should probably be GeoJSON or vector cube, as usual.
* Would it be useful to also pass lines and get points for them or should this be purely polygon focused?
* Sampling strategy could be a callback (so random and Fibonacci would be separate processes) or a given set of methods.

clausmichele:
Sorry for being not so clear, with same structure I was referring to the geoJSON or vector-cube structure/data format, not the content itself.

m-mohr:
Thanks for the clarification, @clausmichele.

I thought about basing percentages around the pixel centers. You should have a known number of pixels and could create a list of points for the pixel centers, right?

clausmichele:
From the last meeting I understood that the process should take as input geometries (geoJSON) and output also geometries to be used in aggregate_spatial, that's why I firstly called it "polygon_to_points". The output would be a list of points which can be used in aggregate_spatial to create a vector-cube for training the ML model.

@mattia6690 @jdries can you comment on this? Do I remember correctly?

m-mohr:
Ah, okay! Then I misunderstood or remembered incorrectly. I can change that. That makes the process simpler by only creating points from a shape without directly combining them with the values from the raster cube. That's also fine for me, just didn't fully understood the idea then. Stay tuned for an updated version... :-)

[...]

Issue #270, opened 2021-07-09
Keep GeoJSON (vector) properties


m-mohr:
Re 3: In general, it would be very useful to keep the GeoJSON properties so that people can actually identify the Features they've provided in the results. I'm not sure if the processes describe this well enough, but we can probably clarify this in general for vector file formats that handle properties.

_Originally posted by @m-mohr in https://github.com/Open-EO/openeo-geopyspark-driver/issues/84#issuecomment-876430614_

clausmichele:
After some discussion, we agreed that this feature is crucial in the following scenario:
1. We need to train a ML model
2. We create a geoJSON containing the polygons with associated classes (for classification) or target values (for regression)
2.1 (Optional) We use the (still to be defined) process polygon_to_points process to extract several points from the geometries.
3. We use aggregate_spatial to create the input vector-cube for the ML model training process.

The ML model training process will need to distinguish which one is the class/target from the output of aggregate_spatial (back-end specific?).

For other scenarios where we do not define the labels/classes/target by ourselves inside the geoJSON but use instead two different rasters the problem does not persist.

m-mohr:
I'm wondering whether it would be the easiest solution to add a new parameter `keep_properties` (default: false) that can be enabled to add this behavior? Or do we assume that this should be possible by default and just "clarify" that keeping the properties is the expected (but baldy documented) behavior of the existing process?

jdries:
I'm considering to start keeping them by default in our processes, mainly to avoid that users would suddenly have to specify this extra argument. Keeping those properties should also not break existing code.
Backends that do not keep them yet, but also do not support training ML models, would not be hugely impacted.

jdries:
Dev telco: avoid adding a parameter, try to implement in the backends.
Process documentation would need to be updated.

m-mohr:
Would a simple addition such as "Vector properties are preserved for vector data cubes and all GeoJSON Features." be good enough?

cc @soxofaan

soxofaan:
yes, I would not overthink it for now. We can finetune it if actual implementation details become more clear

m-mohr:
Done, see commits above (and potentially below).

PR #315, opened 2021-12-20
Add vector_to_points #313

m-mohr:
Adds a new process for the requirements given in #313.

Some questions are still open, see the comments in #313. This is therefore more a basis for further discussions rather than a final proposal.


mattia6690 (approved):
I think that a negative buffer is very useful for many researchers to either exclude or specifically research areas at the boundaries of Polygons or in the vicinity of points and lines.
Otherwise the current structure is fine

m-mohr:
@mattia6690 I assume you want to copy your comment over to `vector_buffer` in #314 and remove the comment here.

proposals/raster_to_points.json
```diff
@@ -0,0 +1,81 @@
...
+            "name": "data",
+            "description": "The raster data cube to extract a sample from.",
+            "schema": {
+                "type": "object",
+                "subtype": "raster-cube"
+            }
+        },
+        {
+            "name": "geometries",
+            "description": "Geometries to extract samples for. If not given or `null`, the bounding box of the raster data cube is used as the Polygon to sample for.\n\nTo maximize interoperability, a nested `GeometryCollection` should be avoided. Furthermore, a `GeometryCollection` composed of a single type of geometries should be avoided in favour of the corresponding multi-part type (e.g. `MultiPolygon`).",
```
mattia6690, line 21: This sounds a bit strange imo: Maybe rather "Input Geometries for sample extraction". Same for the next sentence I would substitute "as the Polygon to sample" with "as input polygon" or something similar

mattia6690 (changes requested):

proposals/raster_to_points.json
mattia6690: the rest is as we've discussed and should be fine for most of the applications

m-mohr: Alright, fixed.

m-mohr:
After the comment from @clausmichele in https://github.com/Open-EO/openeo-processes/issues/313#issuecomment-998683330, I've added another process `vector_to_points` to support the use case you have described. So depending on what we want, we can choose one of the processes in this PR and discard the other one. Or if both processes are potentially useful, merge both.

@clausmichele Is the new process okay for you?

jdries:
I also understood/remember that we would do vector_to_points, and then use aggregate_spatial to sample over a raster if desired.
raster_to_points seems like a shortcut for the same thing, but don't feel the need to have a process that is also covered by a simple combination of other processes. Also doesn't feel like it would be more efficient.

m-mohr:
Alright, I removed raster_to_points.

proposals/vector_to_points.json
```diff
@@ -0,0 +1,68 @@
+{
+    "id": "vector_to_points",
+    "summary": "Sample random points from geometries",
+    "description": "For each geometry samples a number of random points based on one of the sampling method.",
```
soxofaan, line 4: > based on one of the sampling method.

typo, I guess must be

> based on a sampling method

proposals/vector_to_points.json
```diff
@@ -0,0 +1,68 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "count",
+            "description": "The number of points to compute. For each point given the maximum number of computed points is always 1.",
```
soxofaan, line 27: Bit unclear if this parameter is total number  or number per geometry.

I assume it must be the former (total number). But other parts of description seem to suggest the latter:

> For each geometry samples a number of random points

> For each point given ... number is always 1

soxofaan: > For each point given the maximum number of computed points is always 1.

This sentence was very confusing to me. I guess it describes the behavior when your input geometries contain points. I think that discussing that edge case here makes things more confusing than it should be.

proposals/vector_to_points.json
```diff
@@ -0,0 +1,68 @@
...
+            "optional": true,
+            "default": 1,
+            "schema": {
+                "type": "integer",
+                "minimum": 1
+            }
+        },
+        {
+            "name": "method",
+            "description": "The method to distribute the samples:\n\n* `random` (default): Random sampling\n* `fibonacci`: Fibonacci spiral sampling",
```
soxofaan, line 37: I'm not an expert on this, but when a user want's Fibonacci spiral sampling, they might care about CRS of the calculation of the Fibonacci spiral?

Also on the backend side: I think it is far from easy to do Fibonacci sampling in a non-trivial geometry so that you get exactly the requested number of samples.

proposals/vector_to_points.json
m-mohr, line 4: Yeah, we could either append an "s" or use your suggestion

```suggestion
    "description": "For each geometry samples a number of random points based on a sampling method.",
```

m-mohr: Yeah, we could either append an "s" or use your suggestion

```suggestion
    "description": "For each geometry samples a number of random points based on a sampling method.",
```

proposals/vector_to_points.json
m-mohr: Hmm, I guess I'd check what implementations do to not make the implementation overly complicated. 

```suggestion
            "description": "The number of points to compute. Does not apply for point geometries.",
```

proposals/vector_to_points.json
m-mohr: Me neither, so I would appreciate thoughts from others, e.g. @mattia6690 

We could also just leave out the method for now.

proposals/vector_to_points.json
soxofaan: To be honest, "sample based on a sampling method"  sounds quite redundant to me.

If you want the description different enough from the summary, something like this:

> Generate a vector cube of points by sampling random points from input geometries.

m-mohr: Works for me except that "random" would sound a bit like just the "random" method is available although we (at least currently) still have "fibonacci" available, too.

proposals/vector_to_points.json
soxofaan: I think part of my confusion (also see my first comment in this side-thread) is that I first assumed the input data is a geojson object. I somehow missed the part that the input data can also be a vector cube. 

So `count` is the number of samples per "feature" (in geojson sense)?

m-mohr: Yes, I think that was the intention, but as said, let's check implementations, too.

soxofaan: > Does not apply for point geometries for which the count is always `1`.

And what about MultiPoints or geometry collections with multiple points? You could cover that generically with something like:

> Note that points in the input geometries can be selected at most once by the sampling.

m-mohr: That sounds pretty good, indeed! 👍

```suggestion
            "description": "The number of points to compute per geometry. Note that points in the input geometries can be selected at most once by the sampling.",
```

proposals/vector_to_points.json
```diff
@@ -0,0 +1,68 @@
...
+            "schema": {
+                "type": [
+                    "integer",
+                    "null"
+                ]
+            }
+        }
+    ],
+    "returns": {
+        "description": "Returns a vector data cube with the sampled points as `MultiPoint` per geometry given.",
```
soxofaan, line 62: Are only the points returned as geometries, or also the original properties of the input features?

m-mohr: Good question, which is closely related to how we go forward with #270. The behavior should (usually) be consistent across the processes so we'd need to discuss that soon.

proposals/vector_to_points.json
soxofaan: yeah, the random can be dropped

m-mohr:
```suggestion
    "description": "Generate a vector cube of points by sampling points from input geometries.",
```

I hope this is long enough...

proposals/vector_to_points.json
clausmichele: I would add something like "desired" in between: "The desired number of points to compute. ...". It is not obvious that the process can extract exactly that specific amount of points.
An edge case: we provide 5 points and ask for 10: not possible as we said that for each point we can extract only one.

Or maybe I'm just thinking too much and it's not necessary.

proposals/vector_to_points.json
m-mohr: Okay, I can remove Fibonacci. Should we include some kind of a "regular" option instead?

proposals/vector_to_points.json
mattia6690: maybe we can consider also to just name it "n" (generally used as placeholder) or "size" (used e.g. by Edzer in his sf package). "Count" is more likely used in statistics to count certain occurrences

proposals/vector_to_points.json
jdries: I think it makes sense to keep original properties if there is an unambiguous relation between a single input and output feature, like in this case.
Otherwise, it would be very hard to get the properties back.

proposals/vector_to_points.json
```diff
@@ -0,0 +1,68 @@
...
+                    "subtype": "geojson"
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "count",
```
m-mohr, line 26: As proposed in https://github.com/Open-EO/openeo-processes/pull/315#discussion_r782271205:

```suggestion
            "name": "size",
```

proposals/vector_to_points.json
m-mohr: Sure: https://github.com/Open-EO/openeo-processes/pull/315/files#r782296286

proposals/vector_to_points.json
soxofaan: > important to have a regular (e.g. one point every x meters) ... sampling 

@mattia6690  Does that mean that the spatial density of sample points  is more important than their  number? If so, the `count` (or `size`) argument might be annoying in practice  for Fibonacci or other regular sampling methods .

proposals/vector_to_points.json
m-mohr: So I think the way forward is that:
1. Input is a Geometry: You receive a new MultiPoint geometry
2. Input is a Feature: You receive the same Feature with properties preserved, but the geometry is now a MultiPoint geometry.
3. Vector Data Cubes: Properties are preserved (basically as above).

I think this is captured reasonably well with the addition "Vector properties are preserved for vector data cubes and all GeoJSON Features." in the `data` parameter. Happy to incorporate improvements, of course.

clausmichele:
I'm currently trying to implement this process. However, I was thinking that maybe the "count" parameter should be more a maximum number of points to extract in total. Othrwise there could be uge polygons where we extract 100 points a tiny ones where we do the same. What do you think?
I have implemented a basic python code that performs this operation, you can have a look at it ere https://github.com/clausmichele/openeo_vector_to_points/blob/main/vector_to_points_test.ipynb

cc @soxofaan let me know if you have suggestions on how to improve the code!

soxofaan:
> However, I was thinking that maybe the "count" parameter should be more a maximum number of points to extract in total. Otherwise there could be luge polygons where we extract 100 points a tiny ones where we do the same. What do you think?

yes, I'm also wondering whether some kind of target density wouldn't be more practical than a bare count. (Still having optional minimum and maximum count might still be useful.) See https://github.com/Open-EO/openeo-processes/pull/315#discussion_r783222625

proposals/vector_to_points.json
mattia6690: > @mattia6690 Does that mean that the spatial density of sample points is more important than their number? If so, the `count` (or `size`) argument might be practically unusable for Fibonacci or other regular sampling methods .

It depends... you can have a regular grid also based on a size argument. The smaller the size the bigger the gaps between the points. The same is true for the spiral fibonacci sampling. You have the same sampling structure but with a less dense cloud of points with a smaller `count` (or `size`) argument. Even though I would really remove Fibonacci for the moment...

mattia6690:
> I'm currently trying to implement this process. However, I was thinking that maybe the "count" parameter should be more a maximum number of points to extract in total. Otherwise there could be luge polygons where we extract 100 points a tiny ones where we do the same. What do you think? 

I think that this should be in the hand of the user tbh. If you want to have only some samples within a polygon (e.g. for a validation action on a cube) you'd associate a small `size` argument to a big polygon whereas you want to have a high density of points if you want to later extract all the values from an image within the polygon.
I think it might be useful, however, to implement a maximum amount of points based on the polygon extent in order to prevent a computationally intensive but yet unneccessary amount of points.

m-mohr:
So something like three parameters? min, max, and target? Those are still pure "counts", right? I'm not sure how I'd specify a density...

mattia6690:
> So something like three parameters? min, max, and target? Those are still pure "counts", right? I'm not sure how I'd specify a density...

No I would leave the `size` or `count` parameters for the number of points generated within a vector. 
In the implementation of the process, however, it might be useful to set a maximum number of points to be generated. Let's say you have a rather small 10m2 polygons and the user has the `size` parameter set to 1.000.000 there would be points separated by a few cm even though this is not very ueful and computationally intensive. Shouldn't we try to catch this by let's say setting a maximum density of 1 point per m2 or based on a raster-cube pixel size in order to prevent requests like this? I know it is unlikely but it could happen

clausmichele:
@mattia6690 the problem is that this process is unaware of the datacube resolution where you will want to sample the point and it does not necessarily need to be used together with a datacube, i.e. one could use it just to extract points and then use the resulting geoJSON (or vector data) somewhere else.

mattia6690:
> the problem is that this process is unaware of the datacube resolution where you will want to sample the point and it does not necessarily need to be used together with a datacube, i.e. one could use it just to extract points and then use the resulting geoJSON (or vector data) somewhere else.

Thanks for the clarification @clausmichele ! I am wondering whether it could be an option to sample the points based on the pixels (e.g. the centroids of the Pixels within a vector ). This option is often used to extract the pixel values of a raster image / raster cube. But maybe this leads too far within this process...

clausmichele:
> 
> yes, I'm also wondering whether some kind of target density wouldn't be more practical than a bare count. (Still having optional minimum and maximum count might still be useful.) See [#315 (comment)](https://github.com/Open-EO/openeo-processes/pull/315#discussion_r783222625)

@soxofaan  Do you mean min/max each geometry/polygon or in total?
In my opinion, if we define minimum points each feature/geometry always = 1 and maximum/target number of total points as a parameter we can increase the density increasing the max/target number of pixels required. So no additional parameters required and the implementation follows the code I've posted before.

> 
> Thanks for the clarification @clausmichele ! I am wondering whether it could be an option to sample the points based on the pixels (e.g. the centroids of the Pixels within a vector ). This option is often used to extract the pixel values of a raster image / raster cube. But maybe this leads too far within this process...

It would be a different process, since this one does not have any information about "pixels", let's focus on creating a good draft of this and we will be able to think about a new one soon if needed.

soxofaan:
> @soxofaan Do you mean min/max each geometry/polygon or in total?

there I was talking about a per-geometry min/max


> maximum/target number of total points as a parameter we can increase the density increasing the max/target number of pixels required. 

I'm not sure I understand what you mean here:
- you still have a user defined density in some way, and the max setting is just a upper limit of samples points per geometry?
- There is no user defined density, but the pixel density is used as implicit density?
-  there is no user defined density and you just get target number of samples per geometry, meaning you'll get different sample density for  geometries with different area?

clausmichele:
In my previous comment, where I've posted the code https://github.com/Open-EO/openeo-processes/pull/315#issuecomment-1017630399, I was trying to say that in my opinion makes more sense to have a "count","total","max number" (or whatever we want to call it) single parameter defining the total number of points we expect as output, not for each geometry.

The density for each geometry will be based on this total number of points we request and the user does not need to define it.

m-mohr:
@clausmichele @mattia6690 Could someone recap what has been decided on (or discussed) during the dev telco, please? I couldn't fully follow due to the connectivity issues, sorry about that.

mattia6690:
> Could someone recap what has been decided on (or discussed) during the dev telco, please? I couldn't fully follow due to the connectivity issues, sorry about that.

Sure! Here a few remarks:

There should/ could be two possibilities of sampling that are both used for different purposes:
- based on the density. This allows a sampling depending on the size of the polygons. More samples for a bigger and less samples for a smaller polygon
- based on a number (count/size). There you have the same amount of points for each Polygon independent from the area

In addition it could make sense to introduce a distance measure. Like this the user is able to have points with a certain distance to each other. This makes sense when raster values will be extracted later on and you want to avoid duplicate samples or multiple samples at nearly the same place. 

However, we should include either maximum sample number or minimum distance between points (based on the polygon dimension) to avoid oversampling or computationally intensive operations

An inclusion of raster cubes is useful but not within this process since it would increase the complexity too much

m-mohr:
Thank you! So there was no final decision on an approach?

mattia6690:
> Thank you! So there was no final decision on an approach

I think that the two sampling measures (density / fixed value) are quite clear. The other points with the minimum unit and distance are not clearly decided upon yet

mkadunc:
I think a good reference to look at is also the [`spsample` function](https://www.rdocumentation.org/packages/sp/versions/1.4-6/topics/spsample) in the `sp` R package - this seems to have:

- n: approximate sample size (equivalent to the "number/count/size" option)

- cellSize: used for all sampling methods except "random" — this to me is equivalent to the distance option, which is approximately the same as the "density" option

The questions we need to answer for the `vector_to_points` are:

- when a user requests `n` sample points, will the process guarantee exactly `n` points or is the `n` approximate, as in the `spsample` (I suggest we go with approximate, which is much easier to implement for different sampling methods and allows for reasonable behaviour in corner-cases)

- if the user provides both `cellSize` and `n`, do we try to handle it, or just decide that only one of the two should be allowed?

---

I like the addition to support avoiding oversampling or computationally intensive operations with cellSize/distance that is smaller than pixel size. This is only relevant when the user specifies the number of samples with `n` and the cell size has to be determined automatically (when the user specifies the cellSize, we can trust them to choose appropriate size for the dataset).

This leads to an an easy solution, where both `n` and `cellSize` are used: `n` is used to determine `user-requested-cellSize`; if this is smaller than `cellSize`, `cellSize` is used for computing the samples. This means that `cellSize` description can be adjusted as follows:

- **cellSize**: minimal cell size, used for all sampling methods; average density of sampled points will be (proportional to) `density = 1 / (cellSize * cellSize)`. If `n` is provided, the actual `cellSize` used may be larger than specified;

clausmichele:
Continuing the discussion, here is my proposal with some sample visualizations.
Input geometries: (one big and two small polygons, two points)
image

**Case 1**: Random sampling. User wants to sample random points from the provided geometries. Requires that we sample max 1000 points in total and max 100 points each provided geometry.
Parameters:
sampling_strategy = 'random'
max_points_total = 1000
max_points_geometry = 100

Result:
image

Discussion:
The provided max_points_total is used to determine the max_points_geometry value if not provided. If provided, we use that one instead.

**Case 2**: Random sampling. User wants to sample random points from the provided geometries. Requires that we sample max 1000 points in total and does not care about the number of points each geometry.
Parameters:
sampling_strategy = 'random'
max_points_total = 1000
max_points_geometry = null

Result:
image

clausmichele:
@m-mohr @soxofaan @mattia6690 @mkadunc 

Continuing the discussion, here is my proposal with some sample visualizations.
Input geometries: (one big and two small polygons, two points)
image

**Case 1**: Random sampling. User wants to sample random points from the provided geometries. Requires that we sample max 1000 points in total and max 100 points each provided geometry.
Parameters:
sampling_strategy = 'random'
max_points_total = 1000
max_points_geometry = 100

Result:
image

Discussion:
The provided max_points_total is used to determine the max_points_geometry value if not provided. If provided, we use that one instead.

**Case 2**: Random sampling. User wants to sample random points from the provided geometries. Requires that we sample max 1000 points in total and does not care about the number of points each geometry.
Parameters:
sampling_strategy = 'random'
max_points_total = 1000
max_points_geometry = null

Result:
image


**Case 3**: Random sampling. User wants to sample random points from the provided geometries. Requires that we sample max 10 points each geometry.
Parameters:
sampling_strategy = 'random'
max_points_total = null
max_points_geometry = 10

Result:
image


**Case 4**: Regular sampling. User wants to sample points from a regular grid. User provides the cell size / min distance in degrees.
Parameters:
sampling_strategy = 'regular'
call_size = 0.001

Result:
image

m-mohr:
It seems that cases 1-3 and 4 are pretty distinct, so I'm wondering whether it would make sense to have them as separate processes instead of flooding a single process with parameters that depend on each other?

m-mohr:
Thanks @clausmichele - as discussed I've split the processes based on your proposal and tried to fix several minor issues. Please check whether the new definitions make sense. Especially the default values of the count parameters were not clear and setting both values to null was also undefined.

proposals/vector_to_regular_points.json
```diff
@@ -0,0 +1,41 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "cell_size",
+            "description": "The size in degrees of the cell of the regular grid. If the cell size is bigger than the geometry, a single point is returned located in the centroid of the geometry. If the centroid is not enclosed in the geometry area, no point is returned.",
```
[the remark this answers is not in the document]
clausmichele: I've just checked and the answer is yes, this is fine for example:
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": null
    }]
}
```

source: https://datatracker.ietf.org/doc/html/rfc7946 (look for the Feature keyword)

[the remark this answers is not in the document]
clausmichele: Should we add that we would need to return a Feature with empty geometry in that particular case?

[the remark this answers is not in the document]
m-mohr:
```suggestion
            "description": "The size in degrees of the cell of the regular grid. If the cell size is bigger than the geometry, a single point is returned located in the centroid of the geometry. If the centroid is not enclosed in the geometry area, no point is returned (i.e. a feature without geometry).",
```
Like that?

Another question: How does this description work for Point and LineString Geometries as `data`? Or does this process only really work for  (Multi)Polygons?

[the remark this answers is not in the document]
clausmichele: Good question. I would return points as they are (same as in the random sampling), since if someone included them in the input data could mean that they want to have them in output as well. Or should we have a boolean for that? Like "return single Point even if not regularly spaced?".
For lines I think it would be possible to sample points with a spacing depending on the cell. If the spacing is greater than the line length we can again either return the center or nothing.

There are so many combinations to consider actually...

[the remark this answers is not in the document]
m-mohr: Discussed with @clausmichele: We propose the following per geometry:
1. For Polygons: As described above.
2. For Lines: Start with a point at the beginning on the line and then put a new point every `cell_size` degrees (by the way, should we rename cell_size to distance or so?)
3. For points simply return the point.

proposals/vector_to_random_points.json
```diff
@@ -0,0 +1,85 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "geometry_count",
+            "description": "The maximum number of points to compute per geometry. Defaults to a maximum of one point per geometry.\n\nPoints in the input geometries can be selected only once by the sampling.",
```
clausmichele, line 27: I would remove "Defaults to a maximum of one point per geometry" since this excludes the case where we set total_count but not geometry_count, i.e. automatically distribute the requested total_count points among the geometries.

m-mohr: Why does it exclude this? The default value is 1, so if you set it to null specifically you should get the desired behavior, no?

proposals/vector_to_regular_points.json
```diff
@@ -0,0 +1,41 @@
...
+            "schema": {
+                "type": "integer",
```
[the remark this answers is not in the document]
m-mohr: Good point!

```suggestion
                "type": "number",
```

proposals/vector_to_regular_points.json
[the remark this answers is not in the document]
m-mohr: add smth like: distance: minimum distance to other points per geometry

proposals/vector_to_regular_points.json
```diff
@@ -23,16 +23,16 @@
             ]
         },
         {
-            "name": "cell_size",
-            "description": "The size in degrees of the cell of the regular grid. If the cell size is bigger than the geometry, a single point is returned located in the centroid of the geometry. If the centroid is not enclosed in the geometry area, no point is returned.",
+            "name": "distance",
+            "description": "Defines the minimum distance in degrees required between two samples generated for a single geometry.\n\n- For **polygons**, the distance defines the cell sizes of a regular grid that starts at the upper-left bound of each polygon. The centroid of each cell is then a sample point. If the centroid is not enclosed in the polygon, no point is sampled, which results in an empty geometry.\n- For **lines** (line strings), the sampling starts with a point at the first coordinate of the line and then walks along the line and samples a new point each time the distance to the previous point has been reached again.\n- For **points**, the point is returned as given.",
```
clausmichele, line 27: Basically we don't allow to sample points from geometries with centroid which is not within the geometry? This could be bad if we are using a layer like Corine where all the geometries are pretty complex.

m-mohr: This is coming from your proposal above ("If the centroid is not enclosed in the geometry area, no point is returned."). So why have you added it? ;-) We can also specify a different behavior...

clausmichele:
For the vector_to_random_points: should we specify a limit of random points drawn for each polygon/geometry? If a geometry is complex, there could be cases where it's difficult to find random points that are within that geometry and it could lead to long processing time, where the algorithms tries to find a new random point but it's always outside the geometry.

proposals/vector_to_regular_points.json
clausmichele: My proposal was to return the centroid if we can't sample points from the geometry due to its limited size (distance greater than the geometry "size" in lat/lon). If in this case the centroid is not contained within the geometry return an empty one.
From the current description the behavior I understand is: firstly check if the centroid is within the geometry and if yes do the rest, otherwise return an empty feature.

m-mohr: Ah! Yes, there's an ambiguity in the description, indeed. I've tries to fix that in the last commit: bcefb4ba
That should work, right?

m-mohr: So to clarify, what I assume we are trying to express here: 

image

clausmichele:
@edzer Matthias found your implementation of the sampling processes here https://r-spatial.github.io/sf/reference/st_sample.html

We have noticed that you explicitly say that the points are sampled from the bbox of the polygon and therefore it might not return exactly the number of required points for each geometry. What if all the points you draw are not inside the polygon? Do you return an empty geometry or do you discard it?
Me and Matthias were discussing about this. In my opinion it would be better to always return at least one point from each geometry. This is the strategy for `vector_to_random_points` I've thought:

1. Draw `geometry_count number` random points from the geometry bbox
2. Return the ones that are contained in the geometry
3. If no point is within the geometry, compute the centroid of it, if it is contained return it
4. If the centroid is not in the geometry, return the first geometry coordinate

The same procedure is also valid for `vector_to_regular_points`, where at step 1. we sample regularly using the distance provided.

This would be particularly useful when performing a land cover classification task. The Corine layer can have tiny complex geometries and if we do not sample anything from them we would lose a lot of information.

Additionally, since we plan to implement the `vector_buffer` process, returning the first geometry coordinate at step 4 it would not be an issue, since we could firstly apply a negative buffer to the data and therefore all the geometry points will be for sure contained in the original geometries. I hope that you get what I mean!

@jdries I tag you here, if you'd like to give us your opinion about this.

proposals/vector_to_regular_points.json
```diff
@@ -0,0 +1,41 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "distance",
+            "description": "Defines the minimum distance in degrees that is required between two samples generated *inside* a single geometry.\n\n- For **polygons**, the distance defines the cell sizes of a regular grid that starts at the upper-left bound of each polygon. The centroid of each cell is then a sample point. If the centroid is not enclosed in the polygon, no point is sampled. If no point can be sampled for the geometry at all, the first coordinate of the geometry is returned as point.\n- For **lines** (line strings), the sampling starts with a point at the first coordinate of the line and then walks along the line and samples a new point each time the distance to the previous point has been reached again.\n- For **points**, the point is returned as given.",
```
soxofaan, line 27: > minimum distance in degrees

What kind of degrees is that? longitude/latitude? E.g. at the latitude of Belgium, a latitude degree is almost 2 times the distance on the ground of a longitude degree.

> the cell sizes of a regular grid

I guess a rectangular/square grid is meant (and not a triangular/hexagonal). I think it is important to mention that.

> regular grid that starts at the upper-left bound of each polygon.

Is there a particular reason to state that? I think it is computationally more efficient, predictable and stable to use the same grid for all geometries (e.g. aligned with the coordinate ref system)

proposals/vector_to_random_points.json
```diff
@@ -0,0 +1,85 @@
...
+            "description": "Input geometries for sample extraction. Vector properties are preserved for vector data cubes and all GeoJSON Features.\n\nTo maximize interoperability, a nested `GeometryCollection` should be avoided. Furthermore, a `GeometryCollection` composed of a single type of geometries should be avoided in favour of the corresponding multi-part type (e.g. `MultiPolygon`).",
+            "schema": [
+                {
+                    "type": "object",
+                    "subtype": "geojson"
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
```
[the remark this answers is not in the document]
m-mohr: Yes. but that's not backward-compatible.

proposals/vector_to_random_points.json
```diff
@@ -0,0 +1,85 @@
...
+                },
+                {
+                    "title": "Unrestricted",
+                    "type": "null"
+                }
+            ]
+        },
+        {
+            "name": "total_count",
+            "description": "The maximum number of points to compute overall.\n\nThrows a `CountMismatch` exception if the specified value is less than the provided number of geometries.",
```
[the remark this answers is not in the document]
m-mohr: Well, implicitly there's a minimum of one due to the geometry_count being set to "minimum": 1 - should we make this more explicit?

proposals/vector_to_regular_points.json
m-mohr: 1) Goes back to the question about units, we need to decide on a common way here. Pops up now in mostly every vector process PR, will open a separate issue.

2) Yes, will clarify.

3) Is it? I guess, we could replace "each" with polygon "all polygons combined". Originated here, I think https://github.com/Open-EO/openeo-processes/pull/315#discussion_r813034467

proposals/vector_to_random_points.json
```diff
@@ -0,0 +1,85 @@
+{
+    "id": "vector_to_random_points",
+    "summary": "Sample random points from geometries",
+    "description": "Generate a vector data cube of points by sampling random points from input geometries.\n\nThrows a `CountsMissing` exception if `geometry_count` and `total_count` are both unrestricted (i.e. set to `null`).",
```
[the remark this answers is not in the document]
m-mohr: How would you express that schematically while not restricting current functionality?

proposals/vector_to_regular_points.json
clausmichele: > What kind of degrees is that? longitude/latitude? E.g. at the latitude of Belgium, a latitude degree is almost 2 times the distance on the ground of a longitude degree.

Yes, but since we are limited by geoJSON which supports WGS84 only that's the only way to define it at the moment. When we'll have vector-cubes with different CRSs we could use meters instead. Should we consider this already?

> Is there a particular reason to state that? I think it is computationally more efficient, predictable and stable to use the same grid for all geometries (e.g. aligned with the coordinate ref system)

Good idea. Basically we would need to compute a single regular grid for the whole area of interest and then check which points fall in which geometry.

m-mohr: We are not limited by GeoJSON. The schema includes vector-cubes. We should consider more, yes.

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
soxofaan: > implicitly there's a minimum of one due to the geometry_count being set to "minimum": 1

That minimum only applies when geometry_count is set to an integer. When it's `null`, there is no per-geometry upper or lower bound. So yes, I think it makes sense to make that constraint explicit. I would even do that at the level of the process description (so not only at the level of the parameter descriptions).

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
soxofaan: I don't completely understand what my proposal would restrict.

now:
- default:  `geom_count=1` and `total_count=null`  -> 1 sample per geom
- `geom_count=null`   and `total_count=null`   ->  invalid

my proposal would be to change that invalid case to
- `geom_count=null`   and `total_count=null`   ->  total_count=number of geoms so basically: 1 sample per geom


advantage is that you then can make `geom_count=null` by default and the user doesn't have to explicitly set geom_count to null if they just want to set total_count

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
soxofaan: > but that's not backward-compatible

yeah, we can not simplify the existing processes, but I was mainly referring to this geojson copy-pasta in all these new vector processes under proposal

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
clausmichele: so we would have the following scenarios covered with @soxofaan proposal:

- `geom_count=null` and `total_count=null` -> return 1 point each geometry
- `geom_count=null` and `total_count=100` -> return 100 points, distributed between the available geometries
- `geom_count=10` and `total_count=null` -> return 10 points each geometry
- `geom_count=10` and `total_count=1`-> return CountMismatch Exception or return 10 points and ignore total_count?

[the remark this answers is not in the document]
soxofaan: yes indeed, with defaults `geom_count=null` and `total_count=null` , which give same result as current defaults, but usage is simpler

To iterate on your examples:
- `vector_to_points()`  -> 1 point per geometry
- `vector_to_points(total_count=100)` -> 100 points, distributed between the available geometries   or `CountMismatch` if there are more than 100 geometries
- `vector_to_points(geom_count=10)` -> 10 points per geometry
- `vector_to_points(geom_count=10, total_count=100)` -> at most 10 points per geometry and at most 100 points in total:
    - 10 geometries or less:  -> 10 points per geometry
    - between 10 and 100 geometries: -> 100 samples distributed between these geometries (with at least one per geometry
    - more than 100 geometries: `CountMismatch`
- `vector_to_points(geom_count=10, total_count=1)` : a bit of strange edge case, but it's still possible to honor the maximum constraints:
    - 1 geometry ->   1 point per geometry
    - 2 or more geometries:  `CountMismatch`

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: Not sure. It feels like this is a nice shortcut from a user POV and I'd personally enjoy still having that possibility.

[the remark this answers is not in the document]
soxofaan: Indeed, but I would move the responsibility of accepting and adapting GeoJSON to the client side, and keep the API lean and orthogonal

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: Sorry, but I think there's an issue here: So if total_count = null means number of geometries, then `vector_to_points(geom_count=10)` for e.g. 10 geometries would return 1 point per geometry as implicitly the total_count is 10.

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: So the client would implicitly also allow geojson always for those types and add a new process load_geometry (or however it's called) to the process graph if present? That seems rather unintuitive and an explicit approach seems better IMHO.

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: Added to the process description:

> At least one point is computed per input geometry.

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: I assume that total_count always trumps geom_count (which we should probably document).

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
soxofaan: Yes that's indeed the idea. What we for example try to do in the Python client is that the user can use Python specific objects and classes (e.g. shapely geometries or numpy/xarray arrays) and the client takes care to convert these things appropriately to openEO process graph JSON format. 
To me that is more intuitive (or at least user friendly) than requiring the user to do all kinds of conversions themselves.

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
soxofaan: > Sorry, but I think there's an issue here: ... So you'd always need to specify a total_count now which is number of geom * geom_count. 

Indeed, good point. Explaining the my intended behavior  through a "dynamic" default value of  `total_count` is probably not the best solution.  


Taking a step back, I think there are two concerns I want to tackle simultaneously:
- giving  `geom_count=null` and `total_count=null` currently causes `CountsMissing` error. I think we should provide a non-error behavior (e.g. 1 sample per geometry)
- if the user just want to specify `total_count=100`, they should not forget to also explicitly set `geom_count` to `null` otherwise it won't work and they just get 1 sample per geometry (because that is the default for `geom_count`). I don't like this UI/UX-wise.

My proposal to tackle both is:
-  give `geom_count` default value `null`
- allow `geom_count=null` + `total_count=null` : 1 sample per geometry

Achieving the latter can be done (instead of the "dynamic" `total_count` default value trick I suggested earlier) by replacing the note about `CountsMissing` with something like:

> If `geometry_count` and `total_count` are both unrestricted (i.e. set to `null`), one sample per geometry is used.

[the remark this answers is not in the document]
soxofaan: > I assume that total_count always trumps geom_count (which we should probably document).

I don't understand what you mean here. They are both currently described as "maximum", which implies that the most tight constraint trumps the other by definition

clausmichele:
@m-mohr @soxofaan important question: do we consider a MultiPolygon as a single or multiple 'geometries' ? Just to understand how to apply geometry_count in this case

m-mohr:
@clausmichele For me, it's a single geometry.

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: Having geojson as an additional data type doesn't restrict you from implementing the additional behavior in a client though. On the other hand, I'll run into a bit of a problem in the JS / Web Editor with this implicit behavior (and I feel it's inconsistent with existing processes).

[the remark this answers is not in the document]
soxofaan: Yes, in client libs we can provide this convenience functionality anyway.

I'm just worried about about the duplication of all the GeoJSON details in all these vector processes:
- what GeoJSON types are supported?
- what GeoJSON types are recommended or to be avoided?
- how are the different GeoJSON types converted to a vector cube: e.g. MultiPolygon becomes single "feature", a GeometryCollection too? Are there any auto-generated feature ids?


That said, if we have some kind of `geojson_to_vectorcube` process, we can also say "see `geojson_to_vectorcube` for conversion rules.

> I'll run into a bit of a problem in the JS / Web Editor with this implicit behavior

I'm not sure what you mean here. A process graph generated by the python client with this implicit conversion will still be a valid process graph, so the web editor will not notice that there was any implicit conversion in play. Or do you mean on how to support implicit conversion in JS/web editor?

> I feel it's inconsistent with existing processes

yes, but that's because in the existing processes there was no alternative to supporting GeoJSON, as vector cubes were practically undefined

soxofaan:
Indeed, I even remember that we also agreed to handle a GeometryCollection as a single "feature" or entity

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: > I'm just worried about about the duplication of all the GeoJSON details in all these vector processes:
> 
> * what GeoJSON types are supported?

Well, if we add vector cubes it will change to "what geometry types are supported" and be more general. We will still need to make the type distinctions for vector cubes as often you have different behavior based on the type.

> * what GeoJSON types are recommended or to be avoided?
> * how are the different GeoJSON types converted to a vector cube: e.g. MultiPolygon becomes single "feature", a GeometryCollection too? Are there any auto-generated feature ids?

We can basically just say: We allow for GeoJSON what is supported for vector cubes and link to the corresponding docs.

> That said, if we have some kind of `geojson_to_vectorcube` process, we can also say "see `geojson_to_vectorcube` for conversion rules.

We can still have it and link to it. We can basically just say: We allow for GeoJSON what is supported for vector cubes.

> Or do you mean on how to support implicit conversion in JS/web editor?

Yes.

soxofaan:
> 3. If no point is within the geometry, compute the centroid of it, if it is contained return it
> 4. If the centroid is not in the geometry, return the first geometry coordinate

FYI, shapely/geopandas has this method [representative_point](https://shapely.readthedocs.io/en/latest/manual.html#object.representative_point):
> Returns a cheaply computed point that is guaranteed to be within the geometric object.

I'm not completely sure how it works, but I think it's explained here: https://gis.stackexchange.com/questions/414260/how-does-geopandas-representative-point-work

clausmichele:
> > 3. If no point is within the geometry, compute the centroid of it, if it is contained return it
> > 4. If the centroid is not in the geometry, return the first geometry coordinate
> 
> FYI, shapely/geopandas has this method [representative_point](https://shapely.readthedocs.io/en/latest/manual.html#object.representative_point):
> 
> > Returns a cheaply computed point that is guaranteed to be within the geometric object.
> 
> I'm not completely sure how it works, but I think it's explained here: https://gis.stackexchange.com/questions/414260/how-does-geopandas-representative-point-work

Thanks, this is indeed interesting

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
clausmichele: Another question: how would you distribute the points among the geometries if `total_count = 100` and `geom_count = null` ?

My idea was to compute an estimate of the total area of all the geometries, compute the density total_area / total_count and then compute how many samples each geometry based on their area.
Too complex?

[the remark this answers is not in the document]
soxofaan: That question is even more general: 
how would you distribute the points among the geometries if `geom_count > total_count /  number_of_geometries` or `geom_count=null` ?

clausmichele:
@m-mohr I've just noticed that:
- We defined the vector_to_*_points to return the points sampled from each geometry as MultiPoint.
- However, if we want to use this in combination with aggregate_spatial, it aggregates all the points within a MultiPoint feature and returns a single value. This is an issue since we would like to extract and use the values from all the points.

How do we solve this? Should we modify the definition and return just points instead of MultiPoint?

soxofaan:
Good point

and related: if you make each point a separate feature, you probably need a reference back to the original geometry in some way (because you can not just do that based on index anymore)

m-mohr:
Would a new process such as split_multi_geometries work? But overall the question is what is generally useful and required? Do you need a reference back or group the original geometry? Or is the underlying issue aggregate_spatial and we should not (mis-)use it then?

clausmichele:
It's getting every day more complicated!
From a back-end perspective I would rather work only on single geometries and not with their grouped version, it would be easier to define and handle.
From a user perspective I see that it could be useful to define MultiPolygon for a single class and also having a MultiPoint from the same class, avoiding to repeat the properties or target variable multiple times.

Possible solutions:
1. Return single points instead of multi point
2. Define another process to split multiple geometries
3. Modify aggregate_spatial (the less viable solution)
4. 
`split_multi_geometries` would work imho, but we would need to define another process instead of modifying a draft of the current ones.

m-mohr:
It seems like the actual issue here is aggregate_spatial though, right? We are somewhat misusing it to combine points with values, right? If that's the actual issue, then we need to check how we can accomplish that in a clean way.

clausmichele:
> 

Yes, in my opinion using single points in aggregate_spatial it's an edge_case that can be handled as we are doing.

m-mohr:
But then we have no solution. I don't really understand what you are heading at right now.
Another option is probably to let a user choose in vector to points processes whether they want the points to be grouped (returns MultiPoint for each geometry given) or independent (returns points without reference back to original geometries). Would that work?

clausmichele:
This would be a good solution indeed! For me it would be totally fine to let the user choose based on a parameter.

m-mohr:
Does the following work? https://github.com/Open-EO/openeo-processes/pull/315/commits/41b37ade70e2495114e3c00d4a7b3c7d0246cb7b

Also, we still need to update distance to not work in degree any longer, but instead adopt what has been discussed in #330

m-mohr:
@soxofaan Likely because you commented on a specific commit...

m-mohr:
I'm completely overwhelmed by this PR and can't really follow the status any longer. Thus, I think I'd merge this soon and then work on the remaining issues separately in issues and PRs.

proposals/vector_to_regular_points.json
```diff
@@ -0,0 +1,50 @@
...
+                },
+                {
+                    "type": "object",
+                    "subtype": "vector-cube"
+                }
+            ]
+        },
+        {
+            "name": "distance",
+            "description": "Defines the minimum distance in the unit of the referecne system that is required between two samples generated *inside* a single geometry.\n\n- For **polygons**, the distance defines the cell sizes of a regular grid that starts at the upper-left bound of each polygon. The centroid of each cell is then a sample point. If the centroid is not enclosed in the polygon, no point is sampled. If no point can be sampled for the geometry at all, the first coordinate of the geometry is returned as point.\n- For **lines** (line strings), the sampling starts with a point at the first coordinate of the line and then walks along the line and samples a new point each time the distance to the previous point has been reached again.\n- For **points**, the point is returned as given.",
```
soxofaan, line 27: typo

> referecne

soxofaan:
> I'm completely overwhelmed by this PR and can't really follow the status any longer. Thus, I think I'd merge this soon and then work on the remaining issues separately in issues and PRs.

I even think we should do that more often: merge something under `proposals` knowingly it's incomplete. At least there is something implementers can start to play with and find additional issues and unclarities

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: Added a separate issue: https://github.com/Open-EO/openeo-processes/issues/343

proposals/vector_to_random_points.json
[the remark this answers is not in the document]
m-mohr: Opened issues https://github.com/Open-EO/openeo-processes/issues/344 and https://github.com/Open-EO/openeo-processes/issues/345 to keep track of these questions.

proposals/vector_to_regular_points.json
m-mohr: Fixed

m-mohr:
Okay, I think I've opened issues for all remaining open discussions: # 344, #345, #347, #343, #330
If I've missed something, please open a new issue.
Merging now.

Issue #308, opened 2021-11-26
Flatten / Reshape data cube dimensions


m-mohr:
There are sometimes use cases that need to "flatten" (or stack: [xarray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.stack.html#xarray-dataarray-stack), [pands](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.stack.html)) data cube dimensions. Right now, VITO uses "apply_dimension" + target_bands as a workaround, but that may not fully be covered through the specification.

We need to check whether we really want to use that approach long-term, it is a bit weird to use a const operation as callback.
The better approach could be to actually define a new process.

This is already required by multiple use cases: SRR2 UC3, SRR3 UC8
It has already been discussed as part of two other issues at least:
- https://github.com/Open-EO/openeo-processes/issues/233#issuecomment-862356870
- https://github.com/openEOPlatform/architecture-docs/issues/84#issuecomment-929006383

clausmichele:
So, the input could be:
1. x, y, bands, time
2. x, y, bands
3. x, y, time
4. other?

in my opinion the result should have a shape of MxN, resulting from a combination of the available ones in the datacube.
Hence, the process should allow to recombine depending on the user's input.

Practical example:
- I train a random forest regressor with NDVI values as target and [B04,B08] as predictors/features.
- For training, if the data comes out an aggregate_spatial process has already a MxN shape. (M=number of polygons, N=number of bands). This is possible since aggreate_spatial removes the spatial bands (x,y).
- For predicting, the data must have again the MxN shape. No problems if we use again aggregate_spatial. However, if we want to apply the prediction over a raster-cube, we need to flatten the data first.
- If the input data to prediction is then x,y, bands (with bands= [B0,B08] for this particular example), we need to flatten it to MxN, with M= x * y and N = 2 (the number of bands).

If the input data has also the time dimensions, we need to allow a result like:
- M = x * y * time N = bands
- M = x * y N = bands * time (time series regression)

Anyway, we will lose the necessary information for reshaping the output of the machine learning algorithm, so maybe we will also need another process to reshape the output or having a more general reshape process allowing to flatten the data but also reconstruct it following a sample datacube (the data before flattening for instance.)

references:
numpy flatten
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flatten.html
numpy reshape
https://numpy.org/doc/stable/reference/generated/numpy.reshape.html

[...]

PR #316, opened 2021-12-20
Combine dimensions and vice-versa #308

m-mohr:
This PR drafts two processes that can be used to combine/stack/flatten/implode dimensions into a single one and vice-versa (split/unstack/unflatten/explode) - not sure what the right terminology is. As you can see I don't have a clue how to name this intuitively and so I've used the rather "provoking" terminology from the good old PHP days. 💣😅

This is more a basis for further discussions rather than a final proposal.

I'm not exactly sure whether this would make sense for vector cubes as "stringifying" and parsing the vector geometries seems pretty clunky and dirty.


soxofaan:
I didn't went deep into the details yet and I don't have a better naming suggestion at this moment. Just a quick note:

I can imagine that "implode" would be handy in some use cases, but is there a real use case for "explode" at the moment? 
If not, maybe focus first on implode (which is the easy and most useful one) and leave explode (more complex and less useful) for another PR?

jdries:
implode_dimensions already looks fairly good as to what we need. The explode_dimensions process is probably not something we'd implement/need right away.
For naming, combine_dimensions or merge_dimensions would also be pretty descriptive?

proposals/implode_dimensions.json
```diff
@@ -0,0 +1,57 @@
...
+    ],
+    "returns": {
+        "description": "A data cube with the new shape. The dimension properties (name, type, labels, reference system and resolution) for all other dimensions remain unchanged.",
+        "schema": {
+            "type": "object",
+            "subtype": "raster-cube"
+        }
+    },
+    "exceptions": {
+        "DimensionNotAvailable": {
```
m-mohr, line 53: Add an error for when the label separator is invalid (e.g. invalid character (are there any?), leads to conflict, exists in one of the labels).

m-mohr:
We can surely split this PR into two parts before merging.

merge_dimensions <-> split_dimensions works for me, too.

soxofaan:
I searched a bit around and found "flatten" to be the most common verb to describe the operation (or some variation of it): [Python numpy](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flatten.html), [PyTorch](https://pytorch.org/docs/stable/generated/torch.flatten.html?highlight=flatten#torch.flatten), [Analytica](https://wiki.analytica.com/index.php?title=Flatten), [Javascript Array.flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat), R has also a couple of variations on that theme.

"reshape" is also often used: [Python numpy](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html), [Matlab](https://www.mathworks.com/help/matlab/ref/reshape.html). However I think it applies badly for our use case because the implementations require you to know the dimension sizes of you data array (which does not fit our way of working).

Another one I wouldn't have thought of is "unnest", apparently used in some SQL dialects.

m-mohr:
Let's try this novel and fancy approach 😅

- Vote 🚀 (rocket) for flatten/unflatten
- Vote  🎉 (hooray) for merge/split
- Vote 👀 (eyes) for implode/explode
- Vote 👎 (thumbs down) if you don't like any of the options (please comment below)

m-mohr:
According to the vote: Renamed to `flatten_dimensions` and `unflatten_dimension`

clausmichele (approved):

clausmichele:
> According to the vote: Renamed to `flatten_dimensions` and `unflatten_dimension`

@m-mohr I think you have missed an s in `unflatten_dimension`

m-mohr:
Maybe? It was intentional, but I see why this can be confusing.

flatten_dimensionS  because you flatten two dimensionS into one.
unflatten_dimension because you flatten one dimension into two.

Preferences?

dthiex:
> Preferences? :-)

I like `unflatten_dimension` because it already hints that you are starting with 1 single dimension.

clausmichele:
> Depends?! It was intentional, but I see why this can be confusing.
> 
> flatten_dimensionS because you flatten two dimensionS into one (see also the parameter "dimensionS" that accepts an array) unflatten_dimension because you flatten one dimension into two (see also the parameter "dimension" that accepts a single string)
> 
> Preferences? :-)
> 
> Otherwise, this seems to be ready to go.

I see, it makes sense. Sorry I've just thought that it was a type :)

m-mohr:
Define how the resulting labels are sorted. Should be based on the order of the dimensions parameter and the labels. I need to add language.

Used in UC8.

m-mohr:
Just started to implement a test for it in the process test engine, lol. Would have been found and fixed it in the next minutes, too.

soxofaan:
Some other issues I found (I'm adding the corresponding methods in python client):

https://github.com/Open-EO/openeo-processes/blob/bdbd6f6a9aafb3a9c2cbdc3ac42ef67665a27171/proposals/unflatten_dimension.json#L19-L22
> The order of the array defines the order in which the dimension labels and values are split.

It's not clear what "the array" means, because the `dimension` parameter is a string. I think you refer to another parameter `target_dimensions` here?

About "labels and values are split": that's bit confusing: I understand that you split labels, but in what sense are you splitting values?

https://github.com/Open-EO/openeo-processes/blob/bdbd6f6a9aafb3a9c2cbdc3ac42ef67665a27171/proposals/unflatten_dimension.json#L26-L31

> each with a single dimension label to replace.

I don't understand what this is about

> Non-existing dimensions will be created  ...

but what happens with "existing" dimensions: that should raise an `DimensionExists` error, no?

soxofaan:
https://github.com/Open-EO/openeo-processes/blob/bdbd6f6a9aafb3a9c2cbdc3ac42ef67665a27171/proposals/unflatten_dimension.json#L36-L37

> Each label will be split at the first occurrence of the given string only.

does this mean than it is only possible to unflatten to 2 dimensions, not 3 or more?

soxofaan:
(sorry for the post-merge review notes)

soxofaan:
FYI: I moved my post-merge notes to #336

soxofaan:
FYI @clausmichele : I already added `DataCube.flatten_dimensions()` and `DataCube.unflatten_dimension` in the python client  https://github.com/Open-EO/openeo-python-client/commit/e20b6f7859c97f2a4b3a5b320403aa54f5b8bbb1

m-mohr:
> It's not clear what "the array" means, because the `dimension` parameter is a string. I think you refer to another parameter `target_dimensions` here? [...] About "labels and values are split": that's bit confusing: I understand that you split labels, but in what sense are you splitting values?

Yes, fixed.

> I don't understand what this is about [...] but what happens with "existing" dimensions: that should raise an `DimensionExists` error, no?

I completely reworded this, removed the "existing dimension" behavior and added a new error, indeed. (Also rephrased it for aggregate_spatial similarly)

> does this mean than it is only possible to unflatten to 2 dimensions, not 3 or more?

Yes, but actually it doesn't make sense to restrict this => removed.

PR #317, opened 2021-12-21
Improve support for labeled arrays in experimental array processes

m-mohr:
Added better support for labeled arrays in experimental array processes that previously simply discarded the labels. Affected processes:
- `array_append`
- `array_concat`
- `array_modify`

The idea is that this will help overall to improve support for labeled arrays and get a better solution for #233, e.g. through a modified version of `apply_dimension` (as discussed with @clausmichele).

### DIFF
--- a/proposals/array_append.json
+++ b/proposals/array_append.json
@@ -1,7 +1,7 @@
 {
     "id": "array_append",
     "summary": "Append a value to an array",
-    "description": "Appends a value to the end of the array. Array labels get discarded from the array.",
+    "description": "Appends a new value to the end of the array, which may also include a new label for labeled arrays.",
     "categories": [
         "arrays"
     ],
@@ -23,6 +23,18 @@
             "schema": {
                 "description": "Any data type is allowed."
             }
+        },
+        {
+            "name": "label",
+            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as string is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
+            "optional": true,
+            "default": null,
+            "schema": {
+                "type": [
+                    "string",
+                    "null"
+                ]
+            }
         }
     ],
     "returns": {
@@ -32,6 +44,11 @@
             "items": {
                 "description": "Any data type is allowed."
             }
+        }
+    },
+    "exceptions": {
+        "LabelExists": {
+            "message": "A label with the specified name exists."
         }
     },
     "examples": [
@@ -49,21 +66,5 @@
                 3
             ]
         }
-    ],
-    "process_graph": {
-        "append": {
-            "process_id": "array_concat",
-            "arguments": {
-                "array1": {
-                    "from_parameter": "data"
-                },
-                "array2": [
-                    {
-                        "from_parameter": "value"
-                    }
-                ]
-            },
-            "result": true
-        }
-    }
+    ]
 }
--- a/proposals/array_append.json
+++ b/proposals/array_append.json
@@ -48,7 +48,7 @@
     },
     "exceptions": {
         "LabelExists": {
-            "message": "A label with the specified name exists."
+            "message": "An array element with the specified label already exists."
         }
     },
     "examples": [
--- a/proposals/array_concat.json
+++ b/proposals/array_concat.json
@@ -1,7 +1,7 @@
 {
     "id": "array_concat",
     "summary": "Merge two arrays",
-    "description": "Concatenates two arrays into a single array by appending the second array to the first array. Array labels get discarded from both arrays before merging.",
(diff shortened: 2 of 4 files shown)


proposals/array_append.json
```diff
@@ -23,6 +23,18 @@
             "schema": {
                 "description": "Any data type is allowed."
             }
+        },
+        {
+            "name": "label",
+            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as string is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
```
soxofaan, line 29: > a new label for the new value should be given.

the parameter is optional, so I would replace "should" with "can"

proposals/array_append.json
```diff
@@ -34,6 +46,11 @@
             }
         }
     },
+    "exceptions": {
+        "LabelExists": {
+            "message": "A label with the specified name exists."
```
soxofaan, line 51: suggestion

> An array item with specified label already exists

proposals/array_concat.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "array_concat",
     "summary": "Merge two arrays",
-    "description": "Concatenates two arrays into a single array by appending the second array to the first array. Array labels get discarded from both arrays before merging.",
+    "description": "Concatenates two arrays into a single array by appending the second array to the first array. Array labels are kept only if both given arrays are labeled. Otherwise, the labels get discarded from both arrays.",
```
soxofaan, line 4: what about common labels?
- throw error?
- change conflicting labels silently?
- discard all labels?

add a parameter for this?

proposals/array_modify.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "array_modify",
     "summary": "Change the content of an array (remove, insert, update)",
-    "description": "Modify an array by removing, inserting or updating elements. Updating can be seen as removing elements followed by inserting new elements (not necessarily the same number).\n\nAll labels get discarded and the array indices are always a sequence of numbers with the step size of 1 and starting at 0.",
+    "description": "Modify an array by removing, inserting or updating elements. Updating can be seen as removing elements followed by inserting new elements (not necessarily the same number).\n\nArray labels are kept only if both arrays given in `data` and `values` are labeled (or empty). Otherwise, all labels get discarded and the array indices are a sequence of numbers with the step size of 1 and starting at 0.",
```
soxofaan, line 4: same question as with array_concat: what about common labels?

proposals/array_append.json
m-mohr: Should != must. Here should is meant to be understood as a recommendation so that you don't end up with weird string indices.

proposals/array_append.json
m-mohr: We don't use "item" anywhere else, instead we use value or element. So I'd propose:

```suggestion
            "message": "An array element with the specified label already exists."
```

proposals/array_concat.json
m-mohr: Good question. I guess throwing an error would be the easiest solution...
If that's not good, we could discard them.

Changing them silently (to what?) sounds like the worst desirable solution.

soxofaan: Throwing an error would also be my first choice, but it might not be the most user friendly option.

Discarding all labels might be even more frustrating for end user compared to just changing the conflicting ones.

 > Changing them silently (to what?) 

You could take the behavior from array_append "the array index as string is used as the label.". 
Another option is appending a number to get a unique name, like file managers sometimes do when you copy a file to folder and the filename is already taken.

m-mohr: > You could take the behavior from array_append "the array index as string is used as the label.".

Which could - in theory - also exist and then you need yet another behavior for it.

> Another option is appending a number to get a unique name, like file managers sometimes do when you copy a file to folder and the filename is already taken.

Same as above and you this also seems to be pretty difficult to explain so that users can reliably predict the labels.

So for example, if you want to merge the following labels simply by appending integers (starting at 1), what would you get?
["1", "2", "11", "12"] and ["1"]
What should/would the result be for the label in the second array? "111", "13", "3"?

I think I'd stick with an error until people come up with a need for the renaming behavior or back-end want to implement it. I think we could also add such behavior later in a non-breaking way as adding functionality for something that would usually error doesn't seem very breaking to me. 😅

soxofaan: > Same as above and you this also seems to be pretty difficult to explain so that users can reliably predict the labels.

Well I mean if you want to add `file.txt` to a folder, and it already exists, and so does `file 2.txt`, you keep incrementing than number till you have a file name that does not yet exist, be it `file 99.txt` or file 1235.txt`, or ...
The idea is that users might prefer a non-predicable descriptive name over a predictable non-descriptive name, or no name at all.

That being said, I agree that we just could start with throwing an error and improve on that once it becomes an issue

m-mohr: I've added the corresponding description and exception to both processes.

soxofaan (approved):

PR #319, opened 2022-01-26
Allow loading vector cubes in load_... processes

m-mohr:
Allow loading vector data cubes via load_collection, load_result and load_uploaded_files as discussed in the dev telco @jdries .
Currently, only adds the schema. May need some additional descriptive text or so, thus still in draft.

Open question: Bands are pretty much raster-based. We need to check what makes sense there.


m-mohr:
Closing in favor of #382

PR #327, opened 2022-02-22
`round`: Clarify that the rounding for ties applies not only for integers. #326

### LINKED ISSUE #326: round: confusing definition of rounding
**Process ID:** round

**Describe the issue:**
Documentation states:

> If the **fractional part of x** is halfway between **two integers**, one of which is even and the other odd, then the even number is returned. This behavior follows [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229).

However, based on the definition in the standard, I'm not sure this should hold for integers _only_. I would expect this would be used at arbitrary point, e.g. `round(x=0.25, p=1) => 0.2` and `round(p=0.35, p=1) => 0.4`.

**Proposed solution:**
Clarify that the number should be rounded to the one with even least significant digit.

m-mohr:
Tries to fix #326. I found it relatively hard to describe the behavior in "simple" English, it may need further fine-tuning. Thoughts?

### DIFF
--- a/round.json
+++ b/round.json
@@ -1,7 +1,7 @@
 {
     "id": "round",
     "summary": "Round to a specified precision",
-    "description": "Rounds a real number `x` to specified precision `p`.\n\nIf the fractional part of `x` is halfway between two integers, one of which is even and the other odd, then the even number is returned.\nThis behavior follows [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229). This kind of rounding is also called \"round to nearest (even)\" or \"banker's rounding\". It minimizes rounding errors that result from consistently rounding a midpoint value in a single direction.\n\nThe no-data value `null` is passed through and therefore gets propagated.",
+    "description": "Rounds a real number `x` to specified precision `p`.\n\nIf the reamining part of `x` that is relevant for rounding under consideration of `p` is halfway between two numbers, one of which is even and the other odd, then the even number gets rounded to.\nThis behavior follows [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) and is often called \"round to nearest (even)\" or \"banker's rounding\". It minimizes rounding errors that result from consistently rounding a midpoint value in a single direction.\n\nThe no-data value `null` is passed through and therefore gets propagated.",
     "categories": [
         "math > rounding"
     ],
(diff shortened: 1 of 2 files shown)


round.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Rounds a real number `x` to specified precision `p`.\n\nIf the reamining part of `x` that is relevant for rounding under consideration of `p` is halfway between two numbers, one of which is even and the other odd, then the even number gets rounded to.\nThis behavior follows [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) and is often called \"round to nearest (even)\" or \"banker's rounding\". It minimizes rounding errors that result from consistently rounding a midpoint value in a single direction.\n\nThe no-data value `null` is passed through and therefore gets propagated.",
```
Ardweaden, line 4: There is a typo
> reamining 

--------------------

I think (also in the original text)

>  one of which is even and the other odd

is redundant.

----------------------

I would write:

> Rounds a real number `x` to specified precision `p`. If `x` is halfway between closest numbers of precision `p`, it is rounded to the one with even least significant digit. This behavior follows ...

--------------------

Feel free to choose either way, I am fine with the current solution. Only the typo needs to be fixed.

Ardweaden (changes requested):

round.json
m-mohr: "even least significant digit" -> Just from reading it I would not understand what it means. Might be a language issue on my side, but I think that should be easier to understand. Anyway, here's a new variant:

```suggestion
    "description": "Rounds a real number `x` to specified precision `p`.\n\nIf `x` is halfway between closest numbers of precision `p`, it is rounded to the closest even number.\nThis behavior follows [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) and is often called \"round to nearest (even)\" or \"banker's rounding\". It minimizes rounding errors that result from consistently rounding a midpoint value in a single direction.\n\nThe no-data value `null` is passed through and therefore gets propagated.",
```

Ardweaden: Well, it certainly isn't as precise. I believe parity is only really defined for integers, talking about odd or even floats is a matter of debate. It's not hard to imagine `it is rounded to the closest even number` could be understood by someone as rounding to the closest event _integer_ because of that. And what a "least significant digit" is should be pretty clear imo... 

It's similar wording IEEE uses:
>  if the two nearest floating-point numbers bracketing an unrepresentable infinitely
precise result are equally near, the one with an even least significant digit shall be delivered;

Anyway, it is generally understandable if you get the gist of it. I think the examples tend to be the ones that convey the behaviour most clearly anyway.

Ardweaden (approved):

m-mohr:
I added another "of precision `p`", which hopefully helps that people don't think about integers. Anyway, thanks for the proposals. Merged.

PR #332, opened 2022-03-04
Run EOEPCA externally

m-mohr:


jdries:
I guess you'll also want to define the interface that your EOEPCA process needs to support to be callable by openEO? EOEPCA processes are fully generic AFAIK, so I guess they need to be more specific if we'd want to call them from openEO.

Note that we have done the other way around: calling openEO from EOEPCA processes is super easy and allows for some nice use cases where the EOEPCA process can do multiple calls to openEO, or advanced pre/post processing.

m-mohr:
Hmm... I thought OGC API - Processes is the interface? We should probably mention that the other way around might be better defined and easier to do.

Context: We were asked by Guenther Landgraf how it could work to call EOEPCA from openEO, so this is more a demo/conceptish thing rather than a final proposal.

proposals/run_eoepca_externally.json
```diff
@@ -0,0 +1,55 @@
...
+        "cubes",
+        "import",
+        "export",
+        "udf"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "The data to be passed to the external processing platform.",
```
m-mohr, line 15: This implicitly does a save_result to a file format the image can read.

proposals/run_eoepca_externally.json
```diff
@@ -0,0 +1,55 @@
...
+            },
+            "default": {},
+            "optional": true
+        }
+    ],
+    "returns": {
+        "description": "The data processed by the external processing platform.",
+        "schema": {
+            "type": "object",
+            "subtype": "raster-cube"
```
m-mohr, line 45: Here the application package would implictly need to save a result in a way that openEO can read it. The allowed file formats are available through GET /file_formats, but there might be issues in the details (e.g. how are dimensions named in netCDF, etc.)

proposals/run_eoepca_externally.json
```diff
@@ -0,0 +1,55 @@
+{
+    "id": "run_eoepca_externally",
+    "summary": "Run processing externally on EOEPCA",
+    "description": "Runs processing tasks on an [EOEPCA](https://eoepca.org/) compliant processing platform, which essentially runs a hosted docker container.",
```
m-mohr, line 4: We should probably add information about the API interface. Is this always OGC API - Processes?

jdries: It needs to be much more specific than just saying OGC API - Processes.
EOEPCA should have a kind of profile with process signatures that an application package can support to be callable by openEO.

jdries:
Note that from a user perspective, calling an external service may not be necessary. In fact, if there's an application package that satisfies a few conditions, an openEO backend can use it in a way that's very similar to our UDF setup.
This would also avoid having to solve the problem of authentication between these services.

Issue #347, opened 2022-03-15
vector_to_regular_points: Drop or duplicate vector properties for independent points?


m-mohr:
We recently added a "group" parameter to `vector_to_regular_points` so that either MultiPoints or individual points are generated.

For MultiPoints it's easy to keep vector properties.
For Points you'd need to duplicate the vector properties. Is that intended? Right now it's documented to drop the properties. The program would also need to generate new IDs then.

Origin: https://github.com/Open-EO/openeo-processes/pull/315#commitcomment-68730976 and https://github.com/Open-EO/openeo-processes/pull/315#commitcomment-68731148

soxofaan:
What a user probably wants is to keep properties like landcover class, scene classification, etc
What an implementer probably wants to drop/replace is a property that acts as unique identifier.

What might help is to extend the vector cube definition to have a dedicated identifier property (a unique identifier associated with each geometry).  At load time it can be populated from a user specified property, or, if that is missing, autogenerated (e.g. autoincrement int)

For the "ungrouped" version of "vector_to_regular_points" you can then specify:
- the id property is dropped and fully replaced with new ids
- all the other properties are kept/duplicated

m-mohr:
This is actually a much broader discussion that we need to have, regardless of this process. Essentially, what do you do with the metadata that is not part of the data cube? This also applies for raster cubes, it's also basically undefined there.

m-mohr:
I've changed to "vector properties are generally preserved", but for MultiPoint the ID is also preserved while for Points new IDs must be assigned.

clausmichele:
> 

I do not see why the ID must be different from other properties, in my opinion we should just duplicate all of them. Or we need to define it as a mandatory field of a vector-cube.
The back-end can still have another indexing column which can be kept internally for working with the vector-cubes.

I've added a comment here https://github.com/Open-EO/openeo-processes/pull/352/files

m-mohr:
Can't see the comment in the PR, but the ID must change for points because otherwise, it's not unique anymore. I assume IDs should be always unique?

clausmichele:
But the question is, does a feature/geometry always need to have an unique ID? If yes, I've missed this.

The geoJSON samples that I've sent around some time ago didn't have any ID field but only geometries plus some properties.

m-mohr:
No, indeed, an ID is likely not required.

PR #352, opened 2022-03-21
Clarifications for vector_to_*_points

### LINKED ISSUE #344: vector_to_random_points: Change default value of geometry_count to null?
For full context see: https://github.com/Open-EO/openeo-processes/pull/315#discussion_r814782147

In the end, @soxofaan wrote:

> [...] I think there are two concerns I want to tackle simultaneously:
> 
> * giving  `geom_count=null` and `total_count=null` currently causes `CountsMissing` error. I think we should provide a non-error behavior (e.g. 1 sample per geometry)
> * if the user just want to specify `total_count=100`, they should not forget to also explicitly set `geom_count` to `null` otherwise it won't work and they just get 1 sample per geometry (because that is the default for `geom_count`). I don't like this UI/UX-wise.
> 
> My proposal to tackle both is:
> 
> * give `geom_count` default value `null`
> * allow `geom_count=null` + `total_count=null` : 1 sample per geometry
> 
> Achieving the latter can be done (instead of the "dynamic" `total_count` default value trick I suggested earlier) by replacing the note about `CountsMissing` with something like:
> 
> > If `geometry_count` and `total_count` are both unrestricted (i.e. set to `null`), one sample per geometry is used.

m-mohr:
Fixes #344 and #347

### DIFF
--- a/proposals/vector_to_random_points.json
+++ b/proposals/vector_to_random_points.json
@@ -1,7 +1,7 @@
 {
     "id": "vector_to_random_points",
     "summary": "Sample random points from geometries",
-    "description": "Generate a vector data cube of points by sampling random points from input geometries. At least one point is sampled per input geometry.\n\nThrows a `CountsMissing` exception if `geometry_count` and `total_count` are both unrestricted (i.e. set to `null`).",
+    "description": "Generate a vector data cube of points by sampling random points from input geometries. At least one point is sampled per input geometry. Vector properties are preserved.\n\nIf `geometry_count` and `total_count` are both unrestricted (i.e. set to `null`, which is the default), one sample per geometry is used.",
     "categories": [
         "cubes",
         "vector"
@@ -24,9 +24,9 @@
         },
         {
             "name": "geometry_count",
-            "description": "The maximum number of points to compute per geometry. Defaults to a maximum of one point per geometry.\n\nPoints in the input geometries can be selected only once by the sampling.",
+            "description": "The maximum number of points to compute per geometry.\n\nPoints in the input geometries can be selected only once by the sampling.",
             "optional": true,
-            "default": 1,
+            "default": null,
             "schema": [
                 {
                     "type": "integer",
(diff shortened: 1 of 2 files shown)


proposals/vector_to_random_points.json
```diff
@@ -56,7 +56,7 @@
         },
         {
             "name": "group",
-            "description": "Specifies whether the sampled points should be grouped by input geometry (default) or be generated as independent points.\n\n* If the sampled points are grouped, the process generates a `MultiPoint` per geometry given. Vector properties are preserved.\n* Otherwise, each sampled point is generated as a distinct `Point` geometry. Vector properties are *not* preserved.",
+            "description": "Specifies whether the sampled points should be grouped by input geometry (default) or be generated as independent points.\n\n* If the sampled points are grouped, the process generates a `MultiPoint` per geometry given which keeps the original identifier.\n* Otherwise, each sampled point is generated as a distinct `Point` geometry with a new identifier.",
```
clausmichele, line 59: What do you mean with identifier? I don't think it's a mandatory field to have.

m-mohr:
```suggestion
            "description": "Specifies whether the sampled points should be grouped by input geometry (default) or be generated as independent points.\n\n* If the sampled points are grouped, the process generates a `MultiPoint` per geometry given which keeps the original identifier.\n* Otherwise, each sampled point is generated as a distinct `Point` geometry without identifier.",
```

Would that be better?

proposals/vector_to_regular_points.json
```diff
@@ -32,7 +32,7 @@
         },
         {
             "name": "group",
-            "description": "Specifies whether the sampled points should be grouped by input geometry (default) or be generated as independent points.\n\n* If the sampled points are grouped, the process generates a `MultiPoint` per geometry given. Vector properties are preserved.\n* Otherwise, each sampled point is generated as a distinct `Point` geometry. Vector properties are *not* preserved.",
+            "description": "Specifies whether the sampled points should be grouped by input geometry (default) or be generated as independent points.\n\n* If the sampled points are grouped, the process generates a `MultiPoint` per geometry given which keeps the original identifier.\n* Otherwise, each sampled point is generated as a distinct `Point` geometry with a new identifier.",
```
m-mohr, line 35:
```suggestion
            "description": "Specifies whether the sampled points should be grouped by input geometry (default) or be generated as independent points.\n\n* If the sampled points are grouped, the process generates a `MultiPoint` per geometry given which keeps the original identifier.\n* Otherwise, each sampled point is generated as a distinct `Point` geometry without identifier.",
```

proposals/vector_to_random_points.json
clausmichele: Yes better for both proposals!

PR #362, opened 2022-04-04
is_nodata: Clarify NaN #361

### LINKED ISSUE #361: is_nodata(NaN)  when NaN is used as no-data value  ?
In our geotrellis implementation IEEE754 NaN is used as no-data value for float raster cubes.

This creates a conflict in the definition of `is_nodata(NaN)`: https://github.com/Open-EO/openeo-processes/blob/2cfd022fdc4826af89166d88880a873cbea20e90/is_nodata.json#L4

> data is missing data, i.e. equals to null or any of the no-data values

=>  `is_nodata(NaN) = true`

> `The special numerical value NaN as defined by IEEE Standard 754 is not considered no-data and must return false.

=>  `is_nodata(NaN) = false`

m-mohr:
`is_nodata`: Clarified that `NaN` can be considered as no-data value is explicitly specified. 

Fixes #361

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `apply_neighborhood`: Parameter `overlap` was optional but had no default value and no schena for the default value defined.
 - `array_contains` and `array_find`: Clarify that giving `null` as `value` always returns `false` or `null` respectively, also fixed the incorrect examples. [#348](https://github.com/Open-EO/openeo-processes/issues/348)
 - `array_interpolate_linear`: Return value was incorrectly specified as `number` or `null`. It must return an array instead. [#333](https://github.com/Open-EO/openeo-processes/issues/333)
+- `is_nodata`: Clarified that `NaN` can be considered as no-data value is explicitly specified. [#361](https://github.com/Open-EO/openeo-processes/issues/361)
 - `rename_labels`: Clarified that the `LabelsNotEnumerated` exception is thrown if `source` is empty instead of if `target` is empty. [#321](https://github.com/Open-EO/openeo-processes/issues/321)
 - `round`: Clarify that the rounding for ties applies not only for integers. [#326](https://github.com/Open-EO/openeo-processes/issues/326)
 
--- a/is_nodata.json
+++ b/is_nodata.json
@@ -1,7 +1,7 @@
 {
     "id": "is_nodata",
     "summary": "Value is a no-data value",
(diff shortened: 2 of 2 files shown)


soxofaan:
> is only considered as no-data value if specified as no-data value in the metdata.

doesn't that make things a bit too complicated?  `is_nodata(x)` looks like a simple check,  but apparently  it is context dependent.  But what if there is no context/metadata or you're working on a `merge_cube`'d datacube that has multiple metadata sources?

Isn't a regular user always going to expect that `NaN` means "pixel has no data"

m-mohr:
> But what if there is no context/metadata

Then there's no no-data value defined and it is not handled specifically as the computational engine doesn't know about it. Metadata doesn't necessarily mean STAC, by the way. If you read a GeoTiff with e.g. 0 as a no-data value (stored in the GeoTiff) then the no-data value is actually available via the GeoTiff metadata.

>  or you're working on a `merge_cube`'d datacube that has multiple metadata sources?

Good point. We did not define in merge_cubes what happens if you have two data cubes with two different no-data values. We should document that, but I'm not sure right now what the behavior should be? Anyway, either there's still a no-data value define somewhere internally or not.

> Isn't a regular user always going to expect that `NaN` means "pixel has no data"

No, for example for R users there is a difference. In a R (UDF) env you have distinct NA (no-data) and NaN values.

soxofaan (approved):

soxofaan:
Approved this PR already because it resolves #361

the rest of discussion is maybe for another ticket

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `apply_neighborhood`: Parameter `overlap` was optional but had no default value and no schena for the default value defined.
 - `array_contains` and `array_find`: Clarify that giving `null` as `value` always returns `false` or `null` respectively, also fixed the incorrect examples. [#348](https://github.com/Open-EO/openeo-processes/issues/348)
 - `array_interpolate_linear`: Return value was incorrectly specified as `number` or `null`. It must return an array instead. [#333](https://github.com/Open-EO/openeo-processes/issues/333)
+- `is_nodata`: Clarified that `NaN` can be considered as no-data value is explicitly specified. [#361](https://github.com/Open-EO/openeo-processes/issues/361)
```
m-mohr, line 41:
```suggestion
- `is_nodata`: Clarified that `NaN` can be considered as a no-data value only if it is explicitly specified as no-data value. [#361](https://github.com/Open-EO/openeo-processes/issues/361)
```

m-mohr:
Yes, let's open new issues for open issues in this PR.

For merge_cubes I remembered my thinking again: Internally the no-data value from the source get's converted to whatever is used internally in the system for no-data (e.g. NA in R, null in JS, ...) and then there's no difference in the no-data value while merging. The no-data value may get set again to something specific for each output format, see also #364.

PR #363, opened 2022-04-04
`is_nan`: Fixed issue with return value and clarification #360

### LINKED ISSUE #360: is_nan(3.14) = True?
https://github.com/Open-EO/openeo-processes/blob/d0ce91fcd347360b907ea2d9589d7564a2c1e1e3/is_nan.json#L4

>    Returns `true` for numeric values (integers and floating-point numbers), except for the special value `NaN` as defined by ...

Must be "Returns `false`" I guess?

> All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",

The "including arrays with NaN values" is a redundant statement I think. I think the intention was to add "including arrays that contain normal numerical values".

m-mohr:
Fixes #360


is_nan.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "is_nan",
     "summary": "Value is not a number",
-    "description": "Checks whether the specified value `x` is not a number. Returns `true` for numeric values (integers and floating-point numbers), except for the special value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935). All non-numeric data types MUST also return `true`, including arrays that contain `NaN` values.",
+    "description": "Checks whether the specified value `x` is not a number (`NaN`). Returns `false` for numeric values (integers and floating-point numbers), except for the special value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935) and all non-numeric data types.",
```
soxofaan, line 4: >  false for numeric, except for NaN and all non-numeric data types

could be read as

> false for numeric, except for ( NaN and all non-numeric data types)

instead of intended

> false for (numeric, except NaN) and all non-numeric data types

soxofaan: Another thing: is `boolean`  numeric or not?  (FYI: booleans in Python are "subclass" of int)

soxofaan: And another thing about first sentence:

> ... specified value `x` is not a number (`NaN`)

Adding this `(NaN)`  can be confusing: you seem to be equating the narrow IEEE754 float NaN definition   with  everything that is not a number in a broader sense (arrays, strings, ...)

m-mohr: 1) Wouldn't that require a comma after the NaN to understand the latter? Not sure how to clarify. Shall I write:
> false for numeric, except for (1) NaN and  (2) all non-numeric data types

2) boolean is non-numeric, the API generally follows the JSON data types.

3) Indeed, that was an oversight from another draft. I'll remove that again.

soxofaan: > false for numeric, except for (1) NaN and (2) all non-numeric data types

Ok apparently I was doubly confused because you also changed the "return true" to "return false". 
That whole "return false for ... except ... and ..."  construct is pretty confusing language-wise :smile: . 
I would propose to split it in two sentences again (a bit like it was originally) and start with a "return true" case. For example:

> Returns `true` for all non-numeric data types. Returns `false` for numeric values (integers and floating-point numbers), except for the special value `NaN` as defined by ...

m-mohr: As the description pretty much replicates the return value description, I'm removing the actual return value description from the process description and just define what valid numbers are. I feel like this is easy to understand and I'm pretty happy with this iteration.

PR #370, opened 2022-05-18
Added exceptions for empty data cube handling

m-mohr:
As discussed in the dev telco:

- Added a more concrete `DataCubeEmpty` exception to `save_result`.
- Added a `NoDataAvailable` exception to `load_collection`.

### DIFF
--- a/save_result.json
+++ b/save_result.json
@@ -23,7 +23,7 @@
         },
         {
             "name": "format",
-            "description": "The file format to use. It must be one of the values that the server reports as supported output file formats, which usually correspond to the short GDAL/OGR codes. If the format is not suitable for storing the underlying data structure, a `FormatUnsuitable` exception will be thrown. This parameter is *case insensitive*.",
+            "description": "The file format to use. It must be one of the values that the server reports as supported output file formats, which usually correspond to the short GDAL/OGR codes. This parameter is *case insensitive*.\n\n* If the data cue is empty and the file format can't store empty data cubes, a `DataCubeEmpty` exception is thrown.\n* If the file format is otherwise not suitable for storing the underlying data structure, a `FormatUnsuitable` exception is thrown.",
             "schema": {
                 "type": "string",
                 "subtype": "output-format"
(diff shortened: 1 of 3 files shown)


dthiex (approved):
Looks good and makes sense from our side.

soxofaan (approved):
Looks good

soxofaan:
one minor concern: in VITO backend we have the feature to combine the spatial and temporal extents from `load_collection`, `filter_bbox` and `filter_temporal` nodes in the process graph, and use the combined extents directly in the `load_collection` call. This could practically mean that a NoDataAvailable error is thrown from `load_collection`, while it is actually caused by e.g. a `filter_bbox` node. 

Should  NoDataAvailable  also be mentioned in all the `filter_*`  processes as well?

clausmichele:
In my opinion yes, we should have this error message for the filter processes as well.

m-mohr:
@soxofaan Strictly speaking from the process descriptions, the error message doesn't make a lot of sense there. filtering removes dimension labels and you could get to an empty data cube. The question is whether you'd also want to throw an error if in arbitrary filter processes the result is a dimension with no labels (i.e. an empty data cube). This would also apply, if you for example load data from 2022 and then filter for data from May 2021 by accident. If that should result in an error, we could add a separate exception to the filter processes. 
In this case, I'd propose to use a generic DataCubeEmpty exception for load_collection and filters, which says something like:
> The process resulted in an empty data cube, please verify whether the given constraints are correct

save_result.json
```diff
@@ -23,7 +23,7 @@
         },
         {
             "name": "format",
-            "description": "The file format to use. It must be one of the values that the server reports as supported output file formats, which usually correspond to the short GDAL/OGR codes. If the format is not suitable for storing the underlying data structure, a `FormatUnsuitable` exception will be thrown. This parameter is *case insensitive*.",
+            "description": "The file format to use. It must be one of the values that the server reports as supported output file formats, which usually correspond to the short GDAL/OGR codes. This parameter is *case insensitive*.\n\n* If the data cue is empty and the file format can't store empty data cubes, a `DataCubeEmpty` exception is thrown.\n* If the file format is otherwise not suitable for storing the underlying data structure, a `FormatUnsuitable` exception is thrown.",
```
soxofaan, line 26: typo "data cue"

m-mohr: Typo fixed

Issue #346, opened 2022-03-15
process to convert inline GeoJSON to a vector cube


soxofaan:
Closely related to #343 (or maybe even a duplicate):

(do) we need a process that converts inline GeoJSON to a vector-cube

soxofaan:
Still, regardless of #343, it could be useful to have an explicit process that can be more powerful compared to doing implicit conversions with best effort defaults.

e.g. possible features/options to think of or include:
- handle Polygon/MultiPolygon by implicitly handling them as single-feature feature collections
- automatically add a unique (e.g. auto-increment int) "id" property
- drop provided properties, or enforce missing properties with a default

m-mohr:
How's the relation to load_files? In principle, it doesn't matter where the GeoJSON is coming from, but you have the same issues or parameters to influence loading/converting the data. Thus, it's a bit strange to have a dedicated process for inline geojson, but then a completely different process for loading all kinds of files in load_files.

soxofaan:
Good point, there would indeed be quite some overlap.

An alternative is defining a process like `load_vectorcube` , which would be the vector-equivalent of what `load_collection` is for raster cubes. It could accept several source types:
1. (inline) JSON object, to be interpreted as GeoJSON/JSON-FG
2. relative path (string) for loading vector date from user workspace
3. URL string(s) for loading vector data from URL

or we add 1. to the current `load_files` proposal of #322, and find a more generic name than "file" so that it isn't weird to load inline GeoJSON that way

m-mohr:
load_collection could also load a vector-cube, if you expose vector data from the back-end side at /collections. Also, the second word in load_* processes always refers to the origin, not to what is returned.

The issue on the other hand with load_files is that it's weird to mix in inline stuff. This doesn't necessarily only need to be GeoJSON, you could also inline CovJSON at some point. And then it gets messy. 

The underlying issue here is that the processes don't really know about files and formats. This is all handled through /file_formats right now, also for raster. And in that sense GeoJSON is not different, but seems to have some kind of special handling right now due to being easily includable in process graphs. So what we may need to start at the openEO level at some point is to document best practices around file formats, similar to how we do it in Platform already a bit: https://docs.openeo.cloud/federation/backends/fileformats.html

This is challenging though as for some formats the conversion is not really clear (GeoTiff), while for GeoJSON to vector-cubes it might be a bit clearer.

PR #382, opened 2022-09-07
Migrate to general data cube definition

m-mohr:
This goes through the existing processes and makes changes so that, if applicable, they work with vector cubes.

Tasks:
- [x] Replace raster-cube subtype with datacube
- [x] Define datacube dimension minimum requirements
- [x] Update tests
- [x] Update descriptions (e.g. check mentioning of "pixel values" or raster data cubes)
- [x] Check geojson parameters (and also allow vector-cubes)
- [x] ~Add guidelines how geojson should be converted into a vector-cube (as we have geojson as input data type it might be the only file format for which we describe the actual import procedure)~ -> I'll keep this as a separate PR, tracked by #346
- [x] Update dimension types, e.g. in add_dimension
- [x] Update subtype-schemas.json
- Process specific:
  - [x] Update aggregate_spatial with all its flaws and proposed changes
  - [x] load_collection/load_results: Update descriptions to reflect vector use cases
  - [x] filter_bbox/spatial: Update description to reflect vector use cases (only include values that are fully included in the given extent?)
  - [ ] What dimension types are required in ard_normalized_radar_backscatter and sar_backscatter? spatial + bands?
  - [x] merge_cubes needs new/more examples
  - [ ] Does it make sense to extend climatological_normal from raster-cubes to general datacubes?


merge_cubes.json
```diff
@@ -1,7 +1,7 @@
...
...e the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all values. The merged data cube has the same dimensions and labels as the original data cubes, but all values have been processed by the overlap resolver.\n4. A data cube with dimensions (`x`, `y`, `t` / `bands`) or (`x`, `y`, `t`, `bands`) and another data cube with dimensions (`x`, `y`) have the same dimension labels in `x` and `y`. Merging them will join dimensions `x` and `y`, so the lower dimension cube is merged with each time step and band available in the higher dimensional cube. This can for instance be used to apply a digital elevation model to a spatio-temporal data cube. An overlap resolver is *required* to resolve the overlap for all pixels.\n\nAfter the merge, the dimensions with a natural/inherent label order (with a reference system this is each spatial and temporal dimensions) still have all dimension labels sorted. For other dimensions where there is no inherent order, including bands, the dimension labels keep the order in which they are present in the original data cubes and the dimension labels of `cube2` are appended to the dimension labels of `cube1`.",
```
LukeWeidenwalker, line 4: What happens if I try to merge a raster cube and a vector cube? Is this an operation that is even defined? Sorry if this has an obvious answer, but purely from this description I wouldn't know this!

resample_spatial.json
```diff
@@ -12,7 +12,16 @@
             "description": "A raster data cube.",
             "schema": {
                 "type": "object",
-                "subtype": "raster-cube"
+                "subtype": "datacube",
+                "dimensions": [
+                    {
+                        "type": "spatial",
```
LukeWeidenwalker, line 18: Couldn't this also apply to vector cubes? Not for resampling ofc, but for reprojection!
Apologies if this would be obvious to someone with more EO background than I have!

tests/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@openeo/processes-validator",
```
LukeWeidenwalker, line 2: Is this package on npm? Can't find it: https://www.npmjs.com/search?q=%40openeo%2F

merge_cubes.json
m-mohr: I haven't really finalized merge_cubes yet, this is still WIP. What would you expect?

resample_spatial.json
m-mohr: That is true and makes things complicated, unfortunately. Maybe this should be a separate process though as the behavior is really quite a bit different with one working just on the vector labels?!

merge_cubes.json
LukeWeidenwalker: Hmm. My first naive thought was that you could either rasterise the vector data or vectorise the raster data, but both of that gets weird really quickly and even weirder when considering how overlap resolvers would work. 

Also at least for EODC I don't see how we'd implement such a merge_cubes process with our current xarray/geopandas stack (maybe something helpful will come out of [this thread](https://discourse.pangeo.io/t/vector-data-cubes/2904) on the pangeo discourse). So would be in favour of this being undefined behaviour until it becomes clearer what that behaviour should even be and we have a usecase that requires it!

m-mohr: Ah! I think merging a vector and a raster data cube is not allowed in merge_cubes because they're not "compatible". I should mention this explicitly.

Another thought I had today is how merging two vector cubes would work. Issues could arise for an equality check based on the geometry (floating point imprecision) and/or the id (e.g. the same id in two datasets or applying a buffer on a vector cube without changing the ID).

Hmm...

LukeWeidenwalker: > Ah! I think merging a vector and a raster data cube is not allowed in merge_cubes because they're not "compatible". I should mention this explicitly.

Ah, I'm glad, that simplifies things :D

> There are three options I assume:
> 
> geometry: floating point imprecision might arise
> id: vector cube A with unique IDs -> apply buffer (and results in vector cube B), IDs don't change (I assume?) -> merge them, which would lead to a conflict / unpredictable behavior. So either we need to change the IDs when the geometries change or ...?
> Rely on whatever the software being used considers to be equal (I assume that's aligned with SF). I assume that's our best bet, but may lead to different results across backends?
> Thoughts?

- geometry: What about using an `almost_equal` scheme (akin to `GeoSeries.geom_almost_equals`) with an openeo-wide precision setting? I can't think of a case where that wouldn't be appropriate off the top of my head?
- ID: Where were you thinking to get these from in the first place? In shapely 2.0 geometries will become hashable, based on their WKB representation - could that be an option? 

These we'd probably just decide on a convention for, but two other factors that could influence equality:
- row/column order 
- CRS

resample_spatial.json
LukeWeidenwalker: Hmm - is the behaviour really different enough to require entirely separate processes? The `resolution` parameter (which wouldn't apply to vector cubes) is already optional and the resampling methods would also apply to both, right? In the spirit of minimalism (or as a user) I think I favour for this to remain one process that does both.

I also just realised that I'd like to extend my comment to `resample_cube_spatial` - I think there it could be quite useful to allow all datacubes - e.g. to reproject a rastercube to the CRS of a vectorcube.

m-mohr: I don't think the `method` and `align` parameters apply to vector. I could see a use case for having both supported in `resample_cube_spatial` as you may want to pass a raster and vector cube for alignment, but for resample_spatial the vector variant would be much simpler and also usually be a "different code path". What do others think?
A new process `resample_vector` could be defined as follows:
```json
{
    "id": "resample_vector",
    "summary": "Reprojects the vector dimension",
    "description": "Converts the geometries stored in the vector dimension to a different projection.\n\nTo spatially align two data cubes with each other, better use the process ``resample_cube_spatial()``.",
    "categories": [
        "cubes",
        "aggregate & resample",
        "vector"
    ],
    "parameters": [
        {
            "name": "data",
            "description": "A vector data cube.",
            "schema": {
                "type": "object",
                "subtype": "datacube",
                "dimensions": [
                    {
                        "type": "vector"
                    }
                ]
            }
        },
        {
            "name": "projection",
            "description": "Reprojects the geometries stored in the vector dimension to the given projection, specified as as [EPSG code](http://www.epsg-registry.org/), [WKT2 (ISO 19162) string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html), [PROJ definition (deprecated)](https://proj.org/usage/quickstart.html).",
            "schema": [
                {
                    "title": "EPSG Code",
                    "type": "integer",
                    "subtype": "epsg-code",
                    "minimum": 1000,
                    "examples": [
                        3857
                    ]
                },
                {
                    "title": "WKT2",
                    "type": "string",
                    "subtype": "wkt2-definition"
                }
            ]
        }
    ],
    "returns": {
        "description": "A vector data cube with geometries converted into the new projection. The reference system of the vector dimension changes, all other dimensions and properties remain unchanged.",
        "schema": {
            "type": "object",
            "subtype": "datacube",
            "dimensions": [
                {
                    "type": "vector"
                }
            ]
        }
    },
    "links": [
        {
            "href": "https://openeo.org/documentation/1.0/datacubes.html#resample",
            "rel": "about",
            "title": "Resampling explained in the openEO documentation"
        },
        {
            "rel": "about",
            "href": "https://proj.org/usage/projections.html",
            "title": "PROJ parameters for cartographic projections"
        },
        {
            "rel": "about",
            "href": "http://www.epsg-registry.org",
            "title": "Official EPSG code registry"
        },
        {
            "rel": "about",
            "href": "http://www.epsg.io",
            "title": "Unofficial EPSG code database"
        }
    ]
}

```

m-mohr:
This is ready for review! 🥳

CHANGELOG.md
```diff
@@ -4,11 +4,22 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Changes for vector cubes
+
+- Update the processes based on `raster-cubes` to work with `datacube` instead
+- Renamed `create_raster_cube` to `create_data_cube`
+- `add_dimension`: Added new dimension type `vector`
+- New definition for `aggregate_spatial`:
+    - Allows more than 3 input dimensions
+    - Allow to not export statistics by changing the parameter `target_dimenaion`
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
    - Allow to not export statistics by changing the parameter `target_dimension`
```

CHANGELOG.md
```diff
@@ -4,11 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Changes for vector cubes
+
+- Update the processes based on `raster-cubes` to work with `datacube` instead
+- Renamed `create_raster_cube` to `create_data_cube`
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
- Rename `create_raster_cube` to `create_data_cube`
```

CHANGELOG.md
```diff
@@ -4,11 +4,22 @@ All notable changes to this project will be documented in this file.
...
+- `add_dimension`: Added new dimension type `vector`
+- New definition for `aggregate_spatial`:
+    - Allows more than 3 input dimensions
+    - Allow to not export statistics by changing the parameter `target_dimenaion`
+    - Clarified how the resulting vector cube looks like
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
    - Clarify how the resulting vector cube looks like
```

apply_dimension.json
```diff
@@ -1,7 +1,7 @@
...
+    "summary": "Apply a process to all values along a dimension",
+    "description": "Applies a process to all values along a dimension of a data cube. For example, if the temporal dimension is specified the process will work on the values of a time series.\n\nThe process ``reduce_dimension()`` also applies a process to values along a dimension, but drops the dimension afterwards. The process ``apply()`` applies a process to each value in the data cube.\n\nThe target dimension is the source dimension if not specified otherwise in the `target_dimension` parameter. The values in the target dimension get replaced by the computed values. The name, type and reference system are preserved.\n\nThe dimension labels are preserved when the target dimension is the source dimension and the number of values in the source dimension is equal to the number of values computed by the process. Otherwise, the dimension labels will be incrementing integers starting from zero, which can be changed using ``rename_labels()`` afterwards. The number of labels will equal to the number of values computed by the process.",
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
    "description": "Applies a process to all values along a dimension of a data cube. For example, if the temporal dimension is specified the process will work on the values of a time series.\n\nThe process ``reduce_dimension()`` also applies a process to values along a dimension, but drops the dimension afterwards. The process ``apply()`` applies a process to each value in the data cube.\n\nThe target dimension is the source dimension if not specified otherwise in the `target_dimension` parameter. The values in the target dimension get replaced by the computed values. The name, type and reference system are preserved.\n\nThe dimension labels are preserved when the target dimension is the source dimension and the number of values in the source dimension is equal to the number of values computed by the process. Otherwise, the dimension labels will be incrementing integers starting from zero, which can be changed using ``rename_labels()`` afterwards. The number of labels will be equal to the number of values computed by the process.",
```

filter_bbox.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "filter_bbox",
     "summary": "Spatial filter using a bounding box",
-    "description": "Limits the data cube to the specified bounding box.\n\nThe filter retains a pixel in the data cube if the point at the pixel center intersects with the bounding box (as defined in the Simple Features standard by the OGC).",
+    "description": "Limits the data cube to the specified bounding box.\n\n* For raster data cubes, the filter retains a pixel in the data cube if the point at the pixel center intersects with the bounding box (as defined in the Simple Features standard by the OGC).\n* For vector data cubes, the filter retains the geometry into the data cube of the geometry is fully within the bounding box (as defined in the Simple Features standard by the OGC).",
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
    "description": "Limits the data cube to the specified bounding box.\n\n* For raster data cubes, the filter retains a pixel in the data cube if the point at the pixel center intersects with the bounding box (as defined in the Simple Features standard by the OGC).\n* For vector data cubes, the filter retains the geometry in the data cube if the geometry is fully within the bounding box (as defined in the Simple Features standard by the OGC).",
```

CHANGELOG.md
```diff
@@ -4,11 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Changes for vector cubes
+
+- Update the processes based on `raster-cubes` to work with `datacube` instead
```
m-mohr, line 9:
```suggestion
- Update the processes based on `raster-cube` or `vector-cube` to work with `datacube` instead```

File: filter_bbox.json [hunk 10]
```diff
@@ -1,7 +1,7 @@
 {
     "id": "filter_bbox",
     "summary": "Spatial filter using a bounding box",
-    "description": "Limits the data cube to the specified bounding box.\n\nThe filter retains a pixel in the data cube if the point at the pixel center intersects with the bounding box (as defined in the Simple Features standard by the OGC).",
+    "description": "Limits the data cube to the specified bounding box.\n\n* For raster data cubes, the filter retains a pixel in the data cube if the point at the pixel center intersects with the bounding box (as defined in the Simple Features standard by the OGC).\n* For vector data cubes, the filter retains the geometry in the data cube if the geometry is fully within the bounding box (as defined in the Simple Features standard by the OGC).",
```
m-mohr, line 4: Does this make sense? Should it be fully within or something else?

m-mohr: Add a reference to filter_vector

load_collection.json
```diff
@@ -18,7 +18,7 @@
         },
         {
             "name": "spatial_extent",
-            "description": "Limits the data to load from the collection to the specified bounding box or polygons.\n\nThe process puts a pixel into the data cube if the point at the pixel center intersects with the bounding box or any of the polygons (as defined in the Simple Features standard by the OGC).\n\nThe GeoJSON can be one of the following feature types:\n\n* A `Polygon` or `MultiPolygon` geometry,\n* a `Feature` with a `Polygon` or `MultiPolygon` geometry,\n* a `FeatureCollection` containing at least one `Feature` with `Polygon` or `MultiPolygon` geometries, or\n* a `GeometryCollection` containing `Polygon` or `MultiPolygon` geometries. To maximize interoperability, `GeometryCollection` should be avoided in favour of one of the alternatives above.\n\nSet this parameter to `null` to set no limit for the spatial extent. Be careful with this when loading large datasets! It is recommended to use this parameter instead of using ``filter_bbox()`` or ``filter_spatial()`` directly after loading unbounded data.",
```
m-mohr, line 21: We need to check whether it is defined how it works for multiple geometries.

proposals/ard_normalized_radar_backscatter.json
```diff
@@ -13,8 +13,20 @@
             "name": "data",
             "description": "The source data cube containing SAR input.",
             "schema": {
-                "subtype": "raster-cube",
-                "type": "object"
+                "type": "object",
+                "subtype": "datacube",
+                "dimensions": [
```
m-mohr, line 18: Which dimensions are required? Is temporal required?

-> Alex + Michele tend towards no

proposals/ard_surface_reflectance.json
```diff
@@ -13,8 +13,20 @@
             "description": "The source data cube containing multi-spectral optical top of the atmosphere (TOA) reflectances. There must be a single dimension of type `bands` available.",
             "name": "data",
             "schema": {
-                "subtype": "raster-cube",
-                "type": "object"
+                "type": "object",
+                "subtype": "datacube",
+                "dimensions": [
+                    {
+                        "type": "spatial",
```
m-mohr, line 20: x,y,b seems to be correct

proposals/filter_labels.json
```diff
@@ -74,7 +74,7 @@
         "description": "A data cube with the same dimensions. The dimension properties (name, type, labels, reference system and resolution) remain unchanged, except that the given dimension has less (or the same) dimension labels.",
```
m-mohr, line 74: Add a remark what the label is for vector cubes (i.e. user-defined)

proposals/filter_vector.json
```diff
@@ -0,0 +1,92 @@
...
+                "dimensions": [
+                    {
+                        "type": "vector"
+                    }
+                ]
+            }
+        },
+        {
+            "name": "geometries",
+            "description": "One or more base geometries used for filtering, given as GeoJSON or vector data cube.\n\nTo maximize interoperability, `GeometryCollection` and multi geometries (e.g. `MultiPolygon`) should be be avoided.",
```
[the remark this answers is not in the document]
m-mohr: True, this should be clarified.

m-mohr:
For another issue: Should we change the vector_ prefix for the vector-related processes to something else?

climatological_normal.json
```diff
@@ -12,7 +12,12 @@
             "description": "A data cube with exactly one temporal dimension. The data cube must span at least the temporal interval specified in the parameter `climatology-period`.\n\nSeasonal periods may span two consecutive years, e.g. temporal winter that includes months December, January and February. If the required months before the actual climate period are available, the season is taken into account. If not available, the first season is not taken into account and the seasonal mean is based on one year less than the other seasonal normals. The incomplete season at the end of the last year is never taken into account.",
             "schema": {
                 "type": "object",
-                "subtype": "raster-cube"
+                "subtype": "datacube",
+                "dimensions": [
+                    {
+                        "type": "temporal"
```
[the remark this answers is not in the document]
m-mohr: Should in theory be possible for vector data.

proposals/filter_vector.json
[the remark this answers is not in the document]
m-mohr:
```suggestion
            "description": "One or more base geometries used for filtering, given as GeoJSON or vector data cube. If multiple base geometries are provided, the union of them is used.\n\nTo maximize interoperability, `GeometryCollection` should be be avoided.",
```

filter_bbox.json
m-mohr:
```suggestion
    "description": "Limits the data cube to the specified bounding box.\n\n* For raster data cubes, the filter retains a pixel in the data cube if the point at the pixel center intersects with the bounding box (as defined in the Simple Features standard by the OGC). Alternatively, ``filter_spatial()`` can be used to filter by geometry.\n* For vector data cubes, the filter retains the geometry in the data cube if the geometry is fully within the bounding box (as defined in the Simple Features standard by the OGC). Alternatively, ``filter_vector()`` can be used to filter by geometry.",
```

soxofaan:
> For another issue: Should we change the vector_ prefix for the vector-related processes to something else?

Also: let's move that question to another issue, to  unblock #382

m-mohr:
This is also ready for final review. I hope that I have fixed all open issues sufficiently. If you see minor things that we can fix quickly, please mention them here. If you see larger issues to be discussed, maybe open a separate issue so that we can merge this and then work in smaller PRs that are easier to handle.

Additionally, the schema for vector data cubes has changed due to the changes in the other PRs (API, STAC etc.):
```
                {
                    "type": "object",
                    "subtype": "datacube",
                    "dimensions": [
                        {
                            "type": "geometries"
                        }
                    ]
                }
```
So the type is now `geometries` instead of `vector`.

@soxofaan @mkadunc @aljacob @clausmichele @jdries @dthiex @LukeWeidenwalker @pierocampa

m-mohr:
> Also: let's move that question to another issue, to unblock #382

See #397

pierocampa (approved):

aljacob (approved):
looks all good to me. My only concern would also be about the plural of geometries and bands, it would be cleaner to have both in singular, but it is nothing that is a big issue after all, at least these two are consistant then, as I guess it would be quite the effort for everyone to go from bands to band as well.

m-mohr:
@aljacob The naming is discussed in another issue: https://github.com/Open-EO/openeo-api/issues/479

Issue #387, opened 2022-09-30
apply_neighborhood: Data type (datacube/array) in the callback


m-mohr:
**Process ID:** `apply_neighborhood`

**Describe the issue:**
apply_neighborhood is incompatible with the array processes that we typically use in callbacks. So you can currently only use datacube-based processes such as `run_udf`, `apply` and `apply_dimension` in it, but not for example `mean`, `first` or `array_element`.

**Proposed solution:**
Potential solutions for it:
1. an explicit datacube to 1D array function (e.g. `datacube_to_array(data: datacube) -> labeled-array`)
2. Add `labeled-array` as a data type to the callback schema (see process description below)
3. use `apply_dimension` on the remaining dimension (a bit cumbersome)

It doesn't really work if backends implicitly flatten a datacube into a 1D array when needed, because that doesn't fit into the strictly types process descriptions. User would not expect that. Instead, solution 2 would be required.

 I have no strong preference yet.

**Note:** apply_neighborhood is unfortunately not in the draft state and as such a breaking change requires to wait for v2 (which is planned anyway for vector data cubes). 

**Additional context:**
Origin: https://discuss.eodc.eu/t/moving-average-with-low-pass-filter/428/12

Proposal for 2 (only the `process` parameter):
```
{
  "name": "process",
  "description": "Process to be applied on all neighborhoods.",
  "schema": {
    "type": "object",
    "subtype": "process-graph",
    "parameters": [
      {
        "name": "data",
        "description": "The input data, which is a subset of the data cube as specified in `size` and `overlap`. If the given size and overlap result in a one-dimensional data cube it is converted to an array.",
        "schema": [
          {
            "title": "Multi-dimensional data",
            "type": "object",
            "subtype": "raster-cube"
          },
          {
            "title": "One-dimensional data",
            "type": "array",
            "subtype": "labeled-array"
          }
        ]
      },
      {
        "name": "context",
        "description": "Additional data passed by the user.",
        "schema": {
          "description": "Any data type."
        },
        "optional": true,
        "default": null
      }
    ],
    "returns": {
      "description": "The data cube with the newly computed values. The data type and dimensionality must correspond to the input data. For data cubes this means it must have the same dimensions and the dimension properties (name, type, labels, reference system and resolution) must remain unchanged. Otherwise, a `DataCubePropertiesImmutable` exception will be thrown.",
      "schema": [
        {
          "title": "Multi-dimensional data",
          "type": "object",
[...]

Issue #386, opened 2022-09-30
apply_neighborhood: What to return when an overlap is used?


m-mohr:
**Process ID:** `apply_neighborhood`

**Describe the issue:**
What I’m not clear about when working with the process is: If you specify overlap, what is expected to be returned from the callback? the data with or without overlap?

**Proposed solution:**
Actually, how easy is it to detect what the overlap and what the actual data is? In the end, it's probably easier to return the data with overlap, but the back-end discards any changes to the overlap and only replaces the values for the "actual data" without overlap.

**Additional context:**
https://discuss.eodc.eu/t/moving-average-with-low-pass-filter/428/11

m-mohr:
Dev Teclo: Return with overlap, back-ends handles the rest...

Issue #385, opened 2022-09-23
apply_neighborhood : Descriptions are in conflict


m-mohr:
**Process ID:** apply_neighborhood 

**Describe the issue:**
I see some contrary definitions. On the one hand, the description says:

> The process must not add new dimensions, or remove entire dimensions, but the result can have different dimension labels.

On the other hand, in other places it says:

> The dimension properties (name, type, labels, reference system and resolution) must remain unchanged, otherwise a `DataCubePropertiesImmutable` exception will be thrown.

and 

> A data cube with the newly computed values and the same dimensions. The dimension properties (name, type, labels, reference system and resolution) remain unchanged.

**Proposed solution:**

I assume the following part needs to be removed/changed:

> The process must not add new dimensions, or remove entire dimensions, but the result can have different dimension labels.

**Additional context:**
https://discuss.eodc.eu/t/moving-average-with-low-pass-filter/428/10

m-mohr:
What are your thoughts, @jdries ?

PR #390, opened 2022-10-09
apply_neighborhood improvements

m-mohr:
Proposal for issues #385, #386 and #387.


CHANGELOG.md
```diff
@@ -42,10 +44,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 - `aggregate_spatial`:
-  - Clarified that vector properties are preserved for vector data cubes and all GeoJSON Features. [#270](https://github.com/Open-EO/openeo-processes/issues/270)
-  - Clarified that a `TargetDimensionExists` exception is thrown if the target dimension exists.
+    - Clarified that vector properties are preserved for vector data cubes and all GeoJSON Features. [#270](https://github.com/Open-EO/openeo-processes/issues/270)
+    - Clarified that a `TargetDimensionExists` exception is thrown if the target dimension exists.
 - `apply` and `array_apply`: Fixed broken references to the `absolute` process
-- `apply_neighborhood`: Parameter `overlap` was optional but had no default value and no schena for the default value defined.
+- `apply_neighborhood`:
+    - Parameter `overlap` was optional but had no default value and no schena for the default value  defined.
```
LukeWeidenwalker, line 51:
```suggestion
    - Parameter `overlap` was optional but had no default value and no schema for the default value defined.
```

apply_neighborhood.json
```diff
@@ -1,7 +1,7 @@
...
...on a 'neighborhood' of pixels. The neighborhood can extend into multiple dimensions, this extent is specified by the `size` argument. It is not only (part of) the size of the input window, but also the size of the output for a given position of the sliding window. The sliding window moves with multiples of `size`.\n\nAn overlap can be specified so that neighborhoods can have overlapping boundaries. This allows for continuity of the output. The values included in the data cube or array as overlap must also be included in the data cube or array returned from `process` but shouldn't be modified as any changes to the overlap will be discarded. The missing overlap at the borders of the original data cube are made available as no-data (`null`) in the sub data cubes.\n\nThe neighborhood size should be kept small enough, to avoid running beyond computational resources, but a too small size will result in a larger number of process invocations, which may slow down processing. Window sizes for spatial dimensions typically are in the range of 64 to 512 pixels, while overlaps of 8 to 32 pixels are common.\n\nFor the special case of 2D convolution, it is recommended to use ``apply_kernel()``.",
```
soxofaan, line 4: > The values included in the data cube or array as overlap must also be included in the data cube or array returned from `process` but shouldn't be modified as any changes to the overlap will be discarded.

bit simpler:
> The overlap region should be included in the output (data cube or array) returned by `process`, but note that any changed values will be ignored.

CHANGELOG.md
```diff
@@ -42,10 +44,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - `aggregate_spatial`:
-  - Clarified that vector properties are preserved for vector data cubes and all GeoJSON Features. [#270](https://github.com/Open-EO/openeo-processes/issues/270)
-  - Clarified that a `TargetDimensionExists` exception is thrown if the target dimension exists.
+    - Clarified that vector properties are preserved for vector data cubes and all GeoJSON Features. [#270](https://github.com/Open-EO/openeo-processes/issues/270)
+    - Clarified that a `TargetDimensionExists` exception is thrown if the target dimension exists.
 - `apply` and `array_apply`: Fixed broken references to the `absolute` process
-- `apply_neighborhood`: Parameter `overlap` was optional but had no default value and no schena for the default value defined.
+- `apply_neighborhood`:
+    - Parameter `overlap` was optional but had no default value and no schema for the default value defined.
+    - Clarified that the overlap must be included in the returned data cube and can't be modified. [#386](https://github.com/Open-EO/openeo-processes/issues/386)
```
soxofaan, line 52:
```suggestion
    - Clarified that the overlap must be included in the returned data cube but value changes are ignored. [#386](https://github.com/Open-EO/openeo-processes/issues/386)
```

apply_neighborhood.json
m-mohr:
```suggestion
    "description": "Applies a focal process to a data cube.\n\nA focal process is a process that works on a 'neighborhood' of pixels. The neighborhood can extend into multiple dimensions, this extent is specified by the `size` argument. It is not only (part of) the size of the input window, but also the size of the output for a given position of the sliding window. The sliding window moves with multiples of `size`.\n\nAn overlap can be specified so that neighborhoods can have overlapping boundaries. This allows for continuity of the output. The overlap region must be included in the data cube or array returned by `process`, but any changed values will be ignored. The missing overlap at the borders of the original data cube is made available as no-data (`null`) in the sub-data cubes.\n\nThe neighborhood size should be kept small enough, to avoid running beyond computational resources, but a too-small size will result in a larger number of process invocations, which may slow down processing. Window sizes for spatial dimensions typically range from 64 to 512 pixels, while overlaps of 8 to 32 pixels are common.\n\nFor the special case of 2D convolution, it is recommended to use ``apply_kernel()``.",
```

m-mohr: Fixed and also made some other minor simplifications.

soxofaan (approved):

PR #394, opened 2022-11-28
Clarify temporal intervals #331

### LINKED ISSUE #331: load_collection: Clarification on temporal_extent
**Process ID:** load_collection

**Describe the issue:**
Documentation states:
> The second element is the end of the temporal interval. The specified instance in time is excluded from the interval.

Would e.g `["2017-01-01", "2017-01-01"]` or `["2017-01-01T00:00:00Z", "2017-01-01T00:00:00Z"]` be valid or invalid `temporal_extent`s? If it's valid, how would `["2017-01-01", "2017-01-01"]` be converted to a closed datetime interval?

**Proposed solution:**
Add examples to clarify behaviour when values in `temporal_extent` are the same.

**Additional context:**
We noticed VITO's backend considers aforementioned `temporal_extent`s valid.

m-mohr:
It seems the implementations tend more towards the approach that the two elements in the intervals should not be the same, so I created a PR that proposes this breaking change for v2.0.0.

Fixes #331


clausmichele (commented):
**this process allows the value '24' for the hour** of an end time in order to make it possible that left-closed time intervals can fully cover the day.

What's the difference in writing an interval like ["2017-01-01T00:00:00Z", "2017-01-02T00:00:00Z"] to define a full 24h day?

Including this particular case that uses T24:00:00Z seems too complicated and not clearer in my opinion.

m-mohr:
@clausmichele Good question. I had to think quite a bit about it and I can't remember the exact reasons we had back then, but it clearly originates from ISO8601, but it seems they are also struggling with it:

> An amendment was published in October 2022 featuring minor technical clarifications and attempts to remove ambiguities in definitions. The most significant change, however, was the reintroduction of the `24:00:00` format to refer to the instant at the end of a calendar day.

So it was there in the beginning, got removed and then added back again. I assume their reasons for allowing `24` also apply here.

Without knowing their reasoning (ISO standards must be bought 🤮), I found some edge cases where it may matter and surprise users:

- You want to exactly specify the 28th of February as a day. If you specify it as `2020-02-28T00:00:00Z` - `2020-03-01T00:00:00Z` you'd actually get two days as the 29th sneaks in due to the leap year.
- Similarly, it is a bit inconsistent to define 30 days intervals for each month where sometimes you may need to use the first of the next month and sometimes you need to use the 31st (or 28th). It is a bit easier to just use the 30th (ignoring February for now).

LukeWeidenwalker (dismissed):
LGTM!

Ardweaden (dismissed):

clausmichele:
@m-mohr were you able to access the ISO documents?

m-mohr:
Yes! 

## 24
`T24:00:00` is valid and behaves the same as the next day's `T00:00:00` in ISO8601. 

But: I just realized that JSON Schema for the format `date-time` explicitly mentiones RFC3339 compliance, which disallows `24`. So some schema validators may fail on `T24:00:00`. So we'd either need to remove the format from the schema (and loose automatic validation) or comply and disallowe 24.

## Incl./excl. upper boundary
- The definition of intervals is including the upper boundary, but other parts of the standard allow excluding the upper boundary, so here the standard is not very stringent and helpful. To avoid breaking everyones implementations, we should probably keep excluding it. Otherwise, users could get different results as now there's an additional day included.

## Missing parts of a date-time (e.g. a date)
In ISO8601 it seems that omitted components should simply be replaced by zero (`2022-12-14` is basically `2022-12-14T00:00:00`).

## Conclusion
Well, it's hard to come to a conclusion here as there's so many different aspects to consider, but we likely need to "break" someones implementation as VITO seems to be contraty to all other implementations. So to come up with a precise definition, someone needs to bite the bullet anyway.

I'd propose to:
- Keep Inclusive lower boundary and exclusive upper boundary
- Clarify that omitting a component is equivalent to setting it to 0
- Disallow 24 as the hour
- Give more examples and help users to understand what they need to do
- Throw an error if the values for the lower and upper boundary are the same (for backward compatibility backends can keep the previous behavior instead of throwing an error, but that should likely result in a deprecation warning in the logs).
- We may also help in client implementations by just allowing to pass a single string (e.g. "2020-01-01") or a corresponding "Date object", which gets transformed internally into a ["2020-01-01T00:00:00Z",""2020-01-02T00:00:00Z""]

clausmichele:
For me it would also be fine to change the API and make the upper boundary included, since it caused many headaches in the past and it would be easier to understand. However, we need to clarify it in the best way.

If we change the definition I see mostly two cases:
Case 1: I provide a complete range like `["2020-01-01T00:00:00Z","2020-01-02T00:00:00Z""]` no surprises, I should know what I am asking for and I will also get data at `2020-01-02T00:00:00Z` if there's some. -> all good
Case 2. I provide just the date `["2020-01-01","2020-01-02"]`: I would expect to get data for the 2nd of January as well if we say that the upper boundary is included, but instead the range gets converted with zeros and therefore the data on the 2nd is excluded. -> we have to clarify this to the users

m-mohr:
> For me it would also be fine to change the API and make the upper boundary included, since it caused many headaches in the past and it would be easier to understand. However, we need to clarify it in the best way.

As said above, I don't think this is a good idea. Changing this would make data not reproducible as the now included day would intefere. Results would change without users noticing it.

clausmichele:
Well, all the users using the VITO back-end wouldn't notice any difference actually, since they already return the upper bound as well. Maybe it's something to discuss in the next dev telco?

mkadunc:
>     * Clarify that omitting a component is equivalent to setting it to 0

Is it possible to define a union type for the `date-time`, `date` and `year`, where we could then describe the relationships? Or would we add this clarification to all the process parameters where this union type is used?
 
>     * We may also help in client implementations by just allowing to pass a single string (e.g. "2020-01-01") or a corresponding "Date object", which gets transformed internally into a `["2020-01-01T00:00:00Z","2020-01-02T00:00:00Z""]`.

I agree — either overload the process to allow a single parameter, or provide a helper function such as `single_day_to_interval("2020-01-01")`.
 
>     * In `aggregate_temporal` we allow just times (without dates) and there it seems impossible to specify the full range as 23:59:59 would be excluded and 24:00:00 is not allowed. We should point users at `null`, which allows `["00:00:00", null]` to cover the full day.

Another option would be to speficy "the full range" with two identical instants - `[00:00:00, 00:00:00]` would be 24 hours midnight-to-midnight and `[06:00:00, 06:00:00]` would be 24 hours 6am to 6am. The second example wouldn't be fixed by allowing `24` as the hour. But this is assuming that we want to allowing the time intervals to span more than one calendar day (and I don't know enough about aggregate_temporal use-cases using only `time`, so I don't know whether this would be needed).

m-mohr:
@clausmichele But there's not only the VITO implementation. We can change a lot here, but I think the exclusive upper bounds are somewhat set in stone for now due to the reproducibility issue...

> Is it possible to define a union type for the date-time, date and year, where we could then describe the relationships? Or would we add this clarification to all the process parameters where this union type is used?

m-mohr:
@clausmichele But there's not only the VITO implementation. We can change a lot here, but I think the exclusive upper bounds are somewhat set in stone for now due to the reproducibility issue...

> Is it possible to define a union type for the date-time, date and year, where we could then describe the relationships? Or would we add this clarification to all the process parameters where this union type is used?

@mkadunc We could, but it would change the schemas quite a bit and make them more complex so I'd try to avoid that. But we have the temporal-interval subtype which we can use and then it's "only" the comparison operators which need a clarifying word in the parameter description, I think...

> Another option would be to speficy "the full range" with two identical instants - [00:00:00, 00:00:00]

Hmm... something to look into, the aggregate_temporal spec is quite old and I think there's no actual user (and implementation?) for the time part right now.

mkadunc:
> Although, should eq("2018", "2018-01-01T00:00:00Z") return true or false?

If we want to stick to JSON as the only true context for openEO types, then we have to go with the fact that these are all JSON strings and therefore return `false`.

If we want, we could define date and time types for openEO in more detail, in which case we could either go with "date- or time-like JSON strings represent intervals", like this:
- "year" - represents the temporal interval covering a full year in UTC
- "date" - represents the temporal interval covering a full day in UTC
- "date-time"
  - a) represents the temporal interval covering a single time unit (the unit is usually second, but could be shorter, e.g. ds, cs, ms etc. in case fractional-second digits are present)
  - b) represents a time instant (the instant at the start of the indicated second, or at the start of a shorter unit in case fractional-seconds are present)
  
... or we could specify that "all date- or time-like JSON strings represent single instants" (in line with ISO 8601 interpretation), in which case we have:
- "year": represents the instant at the beginning of the specified year, in UTC
- "date": represents the instant at the beginning of the specified day, in UTC
- "date-time": represents a time instant (the instant at the start of the indicated second, or at the start of a shorter unit in case fractional-seconds are present)

Intuitively, the first option would be more correct, but it does add some complexity to the whole system (all date- and time-like parameters would be intervals rather than single scalars)

meta/subtype-schemas.json
```diff
@@ -296,7 +296,8 @@
...
+            "description": "Left-closed temporal interval, represented as two-element array with the following elements:\n\n1. The first element is the start of the temporal interval. The specified instance in time is **included** in the interval.\n2. The second element is the end of the temporal interval. The specified instance in time is **excluded** from the interval.\n\nThe specified temporal strings follow [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html). `null` can be used to specify open intervals.",
```
mkadunc, line 299: I suggest replacing all uses of `instance in time` with `instant in time` or `time instant` or just `instant` (meaning a single zero-duration point on the time axis)

mkadunc (dismissed):

proposals/load_result.json
```diff
@@ -119,11 +119,12 @@
...
+            "description": "Limits the data to load from the batch job result to the specified left-closed temporal interval. Applies to all temporal dimensions. The interval has to be specified as an array with exactly two elements:\n\n1. The first element is the start of the temporal interval. The specified instance in time is **included** in the interval.\n2. The second element is the end of the temporal interval. The specified instance in time is **excluded** from the interval.\n\nThe second element must always be greater/later than the first element. Otherwise, a `TemporalExtentEmpty` exception is thrown.\n\nThe specified temporal strings follow [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html). Also supports open intervals by setting one of the boundaries to `null`, but never both.\n\nSet this parameter to `null` to set no limit for the temporal extent. Be careful with this when loading large datasets! It is recommended to use this parameter instead of using ``filter_temporal()`` directly after loading unbounded data.",
```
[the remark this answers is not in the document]
m-mohr: Thanks, done.

meta/subtype-schemas.json
m-mohr: Thanks, done.

m-mohr:
There's a proposal up to simplify the comparison processes in PR #399 so that we (hopefully) don't need to bother about them in this issue/PR any longer.

m-mohr:
Ready for review!

CHANGELOG.md
```diff
@@ -37,9 +38,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - Allow `null` as default value for units.
     - Input and Output for the `process` can either be data cubes or arrays (if one-dimensional). [#387](https://github.com/Open-EO/openeo-processes/issues/387)
 - `run_udf`: Allow all data types instead of just objects in the `context` parameter. [#376](https://github.com/Open-EO/openeo-processes/issues/376)
-- `load_collection` and `load_result`:
-    - Require at least one band if not set to `null`. [#372](https://github.com/Open-EO/openeo-processes/issues/372)
+- `load_collection` and `load_result`/`load_stac`:
+    - Require at least one band "if not set to `null`. [#372](https://github.com/Open-EO/openeo-processes/issues/372)
```
soxofaan, line 42:
```suggestion
    - Require at least one band if not set to `null`. [#372](https://github.com/Open-EO/openeo-processes/issues/372)
```
typo?

aggregate_temporal.json
```diff
@@ -22,39 +22,36 @@
...
+            "description": "Left-closed temporal intervals, which are allowed to overlap. Each temporal interval in the array has exactly two elements:\n\n1. The first element is the start of the temporal interval. The specified time instant is **included** in the interval.\n2. The second element is the end of the temporal interval. The specified time instant is **excluded** from the interval.\n\nThe second element must always be greater/later than the first element, except for times. Otherwise, a `TemporalExtentEmpty` exception is thrown.",
```
soxofaan, line 25:
```suggestion
            "description": "Left-closed temporal intervals, which are allowed to overlap. Each temporal interval in the array has exactly two elements:\n\n1. The first element is the start of the temporal interval. The specified time instant is **included** in the interval.\n2. The second element is the end of the temporal interval. The specified time instant is **excluded** from the interval.\n\nThe second element must always be greater/later than the first element, except when using time without date. Otherwise, a `TemporalExtentEmpty` exception is thrown.",
```

aggregate_temporal.json
```diff
@@ -22,39 +22,36 @@
...
                     "minItems": 2,
                     "maxItems": 2,
                     "items": {
                         "anyOf": [
                             {
                                 "type": "string",
                                 "format": "date-time",
-                                "subtype": "date-time"
+                                "subtype": "date-time",
+                                "description": "Date and time with a time zone."
```
soxofaan, line 42: time zone is optional, right? (and assumed to be UTC when omitted)?

aggregate_temporal.json
```diff
@@ -235,13 +232,21 @@
         },
         "DistinctDimensionLabelsRequired": {
             "message": "The dimension labels have duplicate values. Distinct labels must be specified."
+        },
+        "TemporalExtentEmpty": {
+            "message": "At least one of the intervals is empty. The second instances in time must always be greater/later than the first instances."
```
soxofaan, line 237: > The second instances in time must always be greater/later than the first instances.

I don't understand this. Do you mean instants instead of instances?

filter_temporal.json
```diff
@@ -33,19 +33,14 @@
                         {
                             "type": "string",
                             "format": "date-time",
-                            "subtype": "date-time"
+                            "subtype": "date-time",
+                            "description": "Date and time with a time zone."
```
soxofaan, line 37: as mentioned above: time zone is optional I think and UTC by default?

filter_temporal.json
```diff
@@ -92,13 +87,21 @@
     "exceptions": {
         "DimensionNotAvailable": {
             "message": "A dimension with the specified name does not exist."
+        },
+        "TemporalExtentEmpty": {
+            "message": "The temporal extent is empty. The second instance must always be greater/later than the first instance."
```
soxofaan, line 92: instant/instance typos here?

load_collection.json
```diff
@@ -112,31 +112,27 @@
...
+            "description": "Limits the data to load from the collection to the specified left-closed temporal interval. Applies to all temporal dimensions. The interval has to be specified as an array with exactly two elements:\n\n1. The first element is the start of the temporal interval. The specified time instant is **included** in the interval.\n2. The second element is the end of the temporal interval. The specified time instant is **excluded** from the interval.\n\nThe second element must always be greater/later than the first element. Otherwise, a `TemporalExtentEmpty` exception is thrown.\n\nAlso supports unbounded intervals by setting one of the boundaries to `null`, but never both.\n\nSet this parameter to `null` to set no limit for the temporal extent. Be careful with this when loading large datasets! It is recommended to use this parameter instead of using ``filter_temporal()`` directly after loading unbounded data.",
```
soxofaan, line 115: > supports unbounded intervals by setting one of the boundaries to `null`, but never both

Why forbidding to set both ends to null?

load_collection.json
```diff
@@ -231,6 +227,9 @@
     "exceptions": {
         "NoDataAvailable": {
             "message": "There is no data available for the given extents."
+        },
+        "TemporalExtentEmpty": {
+            "message": "The temporal extent is empty. The second instance must always be greater/later than the first instance."
```
soxofaan, line 232: instant/instance type here?

meta/subtype-schemas.json
```diff
@@ -355,7 +356,7 @@
                     ],
                     [
                         "12:00:00Z",
-                        "24:00:00Z"
+                        null
```
soxofaan, line 359: Maybe also add an example crossing midnight  (and without explicit timezone)?
e.g. `["23:00:00", "02:00:00"]`

proposals/date_between.json
```diff
@@ -0,0 +1,112 @@
+{
+    "id": "date_between",
+    "summary": "Between comparison for dates and times",
+    "description": "By default, this process checks whether `x` is later than or equal to `min` and before or equal to `max`.\n\nIf `exclude_max` is set to `true` the upper bound is excluded so that the process checks whether `x` is later than or equal to `min` and before `max`.\n\nLower and upper bounds are not allowed to be swapped. So `min` MUST be before or equal to `max` or otherwise the process always returns `false`.",
```
soxofaan, line 4: what about time-only intervals crossing midnight?

`between("01:00:00", min="22:00:00", max="04:00:00)`: invalid, false or true?

proposals/load_stac.json
```diff
@@ -293,6 +289,46 @@
     "exceptions": {
         "NoDataAvailable": {
             "message": "There is no data available for the given extents."
+        },
+        "TemporalExtentEmpty": {
+            "message": "The temporal extent is empty. The second instance must always be greater/later than the first instance."
```
soxofaan, line 294: instant/instance?

aggregate_temporal.json
m-mohr: No, it is required.

aggregate_temporal.json
m-mohr:
```suggestion
            "message": "At least one of the intervals is empty. The second instant in time must always be greater/later than the first instant."
```

filter_temporal.json
m-mohr: No, tz is required.

load_collection.json
m-mohr:
```suggestion
            "message": "The temporal extent is empty. The second instant in time must always be greater/later than the first instant in time."
```

filter_temporal.json
m-mohr:
```suggestion
            "message": "The temporal extent is empty. The second instant in time must always be greater/later than the first instant in time."
```

proposals/load_stac.json
m-mohr:
```suggestion
            "message": "The temporal extent is empty. The second instant in time must always be greater/later than the first instant in time."
```

proposals/date_between.json
m-mohr: I'd say invalid due to 

> Lower and upper bounds are not allowed to be swapped. So `min` MUST be before or equal to `max` or otherwise the process always returns `false`.

m-mohr:
Has anyone the capacity to review this? @dthiex @soxofaan @clausmichele @mkadunc @LukeWeidenwalker 
Thanks!

mkadunc (approved):

Issue #368, opened 2022-05-11
predict_class and predict_probabilities


soxofaan:
@JeroenVerstraelen is working on an implementation of CatBoost base ML in the VITO backend and while discussing details a couple of things came up:

- the `predict_catboost`  process would be practically identical to `predict_random_forest`, except for some textual differences in title and descriptions. Turns out that it is not really necessary to define a dedicated `predict_` process for each kind of machine learning model: all the model details are embedded in the `ml-model` object and you could just use a single `predict(data: array, model: ml-model)` for all kinds of ML models.
- for some use cases we want to predict the probability of each class instead of a single class prediction. We first considered adding a parameter to toggle between class output or probabilities output, but that would mean that the output type would change: scalar for class prediction and array for probability prediction. Moreover, the former has to be used in `reduce_dimension` and the other in `apply_dimension`. It felt error prone and confusing to let these two different patterns depend on a rather inconspicuous boolean parameter. It might be better to have a separate processes for class prediction and probabilities prediction

So with this background, the proposal is to introduce two generic ml prediction processes: 
- `predict_class(data: array, model: ml-model) -> number` 
- `predict_probabilities(data: array, model: ml-model) -> array`

m-mohr:
Sounds reasonable. What do others think? @clausmichele @LukeWeidenwalker

LukeWeidenwalker:
Some thoughts:
- `predict_class()` is conceptually just an `argmax` over the probabilities, so `predict_probabilities()` + something extra. Would it make sense to introduce an `argmax` process instead? I don't think it's asking too much of users to understand that for classification you need probabilities for each class. 
- This process only makes sense for classification models. sklearn has this taxonomy for [models](https://scikit-learn.org/stable/modules/classes.html)  (see Base classes -> `Regressor`, `Classifier`, `Cluster`, etc.). If we want to know exactly what the output from an inference process will be (scalar? array of probabilities? array of indices (in the case of clustering)?), we need to know what kind of model we're dealing with. Curious what others think, but I'd be reluctant to go down a path where openeo attempts to replicate significant pieces of sklearn.
- Without replicating/exposing large parts of already existing ML libraries (e.g. sklearn, pytorch/tensorflow, catboost), can openeo enable users to train quality models? E.g. what about model evaluation? Training data selection? Hyperparameter tuning? None of my usual ecosystem is available, so tbh I'd have a hard time building models whose performance I'm actually confident in.
[...]

PR #396, opened 2022-11-28
Generalize predict processes for ML models

m-mohr:
As discussed in #368, a first draft to generalize the former random forest specific ML processes for predictions into one process for simple class predictions and another process for class probabilities. Please check carefully, I don't have an ML background ;-)

### DIFF
--- a/proposals/predict_ml_model.json
+++ b/proposals/predict_ml_model.json
@@ -1,0 +1,42 @@
+{
+    "id": "predict_ml_model",
+    "summary": "Predict values values using a ML model",
+    "description": "Applies a machine learning model to an array and predicts a value/class for it.",
+    "categories": [
+        "machine learning",
+        "reducer"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "An array of numbers.",
+            "schema": {
+                "type": "array",
+                "items": {
+                    "type": [
+                        "number",
+                        "null"
+                    ]
+                }
+            }
+        },
+        {
+            "name": "model",
+            "description": "A ML model that can be trained with one of the ML processes such as ``fit_class_random_forest()``.",
+            "schema": {
+                "type": "object",
+                "subtype": "ml-model"
+            }
+        }
+    ],
+    "returns": {
+        "description": "The predicted value. Returns `null` if any of the given values in the array is a no-data value.",
+        "schema": {
+            "type": [
+                "number",
+                "null"
+            ]
+        }
+    }
+}
--- a/proposals/predict_ml_model_probabilities.json
+++ b/proposals/predict_ml_model_probabilities.json
@@ -1,7 +1,7 @@
 {
-    "id": "predict_random_forest",
-    "summary": "Predict values based on a Random Forest model",
-    "description": "Applies a Random Forest machine learning model to an array and predict a value for it.",
+    "id": "predict_ml_model_probabilities",
+    "summary": "Predict class probabilities using a ML model",
+    "description": "Applies a machine learning model to an array and predicts (class) probabilities for them.",
     "categories": [
         "machine learning",
         "reducer"
@@ -23,7 +23,7 @@
         },
         {
             "name": "model",
-            "description": "A model object that can be trained with the processes ``fit_regr_random_forest()`` (regression) and ``fit_class_random_forest()`` (classification).",
+            "description": "A ML model that can be trained with one of the ML processes such as ``fit_regr_random_forest()``.",
             "schema": {
                 "type": "object",
                 "subtype": "ml-model"
@@ -31,12 +31,15 @@
         }
     ],
     "returns": {
-        "description": "The predicted value. Returns `null` if any of the given values in the array is a no-data value.",
+        "description": "The predicted (class) probabilities. Returns `null` if any of the given values in the array is a no-data value.",
         "schema": {
-            "type": [
-                "number",
-                "null"
-            ]
+            "type": "array",
+            "items": {
+                "type": [
+                    "number",
+                    "null"
+                ]
+            }
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `fit_regr_random_forest`
     - `flatten_dimensions`
     - `load_ml_model`
-    - `predict_random_forest`
+    - `predict_ml_model`
+    - `predict_ml_model_probabilities`
     - `save_ml_model`
     - `unflatten_dimension`
     - `vector_buffer`
--- a/proposals/load_ml_model.json
+++ b/proposals/load_ml_model.json
@@ -36,7 +36,7 @@
         }
     ],
     "returns": {
-        "description": "A machine learning model to be used with machine learning processes such as ``predict_random_forest()``.",
+        "description": "A machine learning model to be used with machine learning processes such as ``predict_ml_model()`` or ``predict_ml_model_probabilities()``.",
         "schema": {
             "type": "object",
             "subtype": "ml-model"
--- a/proposals/predict_curve.json
+++ b/proposals/predict_curve.json
@@ -1,6 +1,6 @@
 {
     "id": "predict_curve",
-    "summary": "Predict values",
+    "summary": "Predict values using a model function",
     "description": "Predict values using a model function and pre-computed parameters. The process is primarily intended to compute values for new labels, but it can also fill gaps where existing labels contain no-data (`null`) values.",
     "categories": [
         "cubes",


proposals/predict_ml_model.json
```diff
@@ -1,7 +1,7 @@
 {
-    "id": "predict_random_forest",
-    "summary": "Predict values based on a Random Forest model",
-    "description": "Applies a Random Forest machine learning model to an array and predict a value for it.",
+    "id": "predict_ml_model",
```
soxofaan, line 2: `predict_ml_model` looks a bit weird to me, it reads like you will be predicting the model (which is a confusing statement), instead of letting the model predict classes.

Something like `predict_class`, `ml_model_predict` or even `ml_predict` would feel better.

Note that we for the array, text, date related processes, we also use a prefix based naming (`array_append`, `array_apply`, `date_shift`) instead of a postfix based naming.

proposals/predict_ml_model.json
```diff
@@ -1,7 +1,7 @@
...
+    "summary": "Predict values values using a ML model",
```
soxofaan, line 3:
```suggestion
    "summary": "Predict classification or regression values using a ML model",
```

proposals/predict_ml_model.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Applies a machine learning model to an array and predicts a value/class for it.",
```
soxofaan, line 4:
```suggestion
    "description": "Applies a machine learning model to an array of input features and predicts an output value or class for it.",
```

proposals/predict_ml_model.json
```diff
@@ -23,7 +23,7 @@
         },
         {
             "name": "model",
-            "description": "A model object that can be trained with the processes ``fit_regr_random_forest()`` (regression) and ``fit_class_random_forest()`` (classification).",
+            "description": "A ML model that can be trained with one of the ML processes such as ``fit_class_random_forest()``.",
```
soxofaan, line 26:
```suggestion
            "description": "A ML model that was trained with one of the ML training processes such as ``fit_class_random_forest()``.",
```

proposals/predict_ml_model_probabilities.json
```diff
@@ -0,0 +1,45 @@
+{
+    "id": "predict_ml_model_probabilities",
```
soxofaan, line 2: Like above, this process name looks a bit weird to me.
I'd prefer something like `predict_probabilities`, `ml_model_predict_probabilities` or `ml_predict_probabilities`

proposals/predict_ml_model_probabilities.json
```diff
@@ -0,0 +1,45 @@
...
+    "summary": "Predict class probabilities using a ML model",
+    "description": "Applies a machine learning model to an array and predicts (class) probabilities for them.",
```
soxofaan, line 4:
```suggestion
    "description": "Applies a machine learning model to an array of input features and predicts the probabilities of the output classes.",
```

proposals/predict_ml_model_probabilities.json
```diff
@@ -0,0 +1,45 @@
...
+    "categories": [
+        "machine learning",
+        "reducer"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "An array of numbers.",
```
soxofaan, line 13:
```suggestion
            "description": "An array of input features.",
```

proposals/predict_ml_model_probabilities.json
```diff
@@ -0,0 +1,45 @@
...
+                    "type": [
+                        "number",
+                        "null"
+                    ]
+                }
+            }
+        },
+        {
+            "name": "model",
+            "description": "A ML model that can be trained with one of the ML processes such as ``fit_regr_random_forest()``.",
```
soxofaan, line 26:
```suggestion
            "description": "A ML model that was trained with one of the ML training processes such as ``fit_regr_random_forest()``.",
```

proposals/predict_ml_model_probabilities.json
```diff
@@ -0,0 +1,45 @@
...
+            "schema": {
+                "type": "object",
+                "subtype": "ml-model"
+            }
+        }
+    ],
+    "returns": {
+        "description": "The predicted (class) probabilities. Returns `null` if any of the given values in the array is a no-data value.",
```
soxofaan, line 34:
```suggestion
        "description": "The predicted class probabilities. Returns `null` if any of the given values in the array is a no-data value.",
```

I'm not sure about the `Returns null if any ...`:  that depends on the capability of the ML model to handle null/nodata I think

soxofaan (commented):
A couple of notes and suggestions.
However I'm not a big ML expert either, so it would be good to collect some more input from other reviewers that do ML on a daily basis

soxofaan:
Another side note (not necessarily to tackle in this PR): 
Ultimately, I think, `predict_curve` becomes obsolete in favor of what is currently named `predict_ml_model` (considering that  `fit_curve` basically returns just another ML model)

proposals/predict_ml_model.json
m-mohr: My aim was to align with `predict_curve` so that in docs they would be listed side by side. Unfortunately, we are not very consistent with prefixes/suffixes (array_apply / date_shift / load_collection / save_result)...

Is there any other case where predict_class / predict_probabilities could become useful without ML?

proposals/predict_ml_model.json
m-mohr: Are there any other options that could occur in the future? (Also the summary is too long, we aim for 60 chars in the test suite.)

proposals/predict_ml_model.json
soxofaan: The thing with the name `predict_class`  is that that process would not only work for ML classification, but also ML regression, so ideally, the term `class` should be avoided. Unless there is a good reason to have separate inference/predict processes for ML classification and ML regression, but as far as I know that's not the case.


> Is there any other case where predict_class / predict_probabilities could become useful without ML? That was the main reason I added ml_model in the first place...

To me, "predict" implies "machine learning", so having both "ml" and "predict" in the name is slightly redundant.
However, it might be better for discoverability and self-documenting reasons to keep that bit of redundancy.

proposals/predict_ml_model.json
soxofaan: I'd think anything outside of classification/regression is too exotic to be worried about now.

shorter text (if dropping the main verb is ok): "Classification or regression using a ML model"

proposals/predict_ml_model_probabilities.json
LukeWeidenwalker: This is actually tricky - some thoughts:
- E.g. in deep learning if I've anticipated missing data, I could have used `torch.nan_to_num` to replace nan with a carefully chosen value in both training and inference. If I then export that model with e.g. ONNX, that transformation will be baked in and my model will know how to deal with missing data. I'm not sure how every other frameworks handles this (e.g. sklearn or xgboost) - but it's super important that this value is the same during inference as it was during training!
- If the model handles nans, and the `predict_ml_model` process just ignores nan-values and lets it through, then I get the correct behaviour. 
- However, if it fails on nan-values, then the only way the user can fix this is to replace nan-values with the correct value in an extra openeo process before. To do that I need to know what that value was during training - this might not be easily available!
- If the model doesn't handle nans (either they didn't add nan-handling to inference code, or just didn't train on samples with nans at all, etc.), then what would happen really depends on the framework. It might just crash, or it might run through, with the nan values subtly impacting the predictions. 
- some options I can think of:
	- Make this a parameter of the process (`fail_on_nan` or similar), fail by default and allow overriding if the user knows that their model can handle nans out of the box or is willing to throw the dice.
	- Add information about what to replace nan values with to the `ml-model` STAC extension and use that if available

I think as a user my preference would be for this process to assume that the model has already been constructed to handle nans correctly in both training and inference and therefore doesn't try to interfere.

soxofaan: If I understand correctly, you agree that it would be better to drop 
> Returns `null` if any of the given values in the array is a no-data value.

from the general description of `predict_..._probabilities` ?

LukeWeidenwalker: > If I understand correctly, you agree that it would be better to drop
> 
> > Returns `null` if any of the given values in the array is a no-data value.
> 
> from the general description of `predict_..._probabilities` ?

Yeah, exactly!

proposals/predict_ml_model_probabilities.json
m-mohr: Thinking about the recent discussion (prefix is based on the primary input), it should probably be `ml_predict_probabilities` (or `ml_model_predict_probabilities` which is rather lon)

m-mohr:
Potentially interesting for "bring your own model": https://onnx.ai/

m-mohr:
We continue in #441 - We still need to bring some of the comments over.

PR #400, opened 2023-01-31
apply_dimension: Clarification of behavior #357

### LINKED ISSUE #357: apply_dimension: Clarification of behaviour
**Process ID:** apply_dimension

**Describe the issue:**
Documentation states:

> The target dimension is the source dimension if not specified otherwise in the target_dimension parameter. The pixel values in the target dimension get replaced by the computed pixel values. The name, type and reference system are preserved.

>There are three cases how the dimensions can change:
>...
>2. The source dimension is not the target dimension and the latter exists:
The number of dimensions decreases by one as the source dimension is dropped.
The target dimension properties name and type remain unchanged. All other dimension properties change as defined in the list below.

We don't understand how this is supposed to work.

For example what would the result be in the following case:
```python
data = DataCube([[1,2],[3,4]], dimensions=("temporal", "band"))
apply_dimension(data=data, process, dimension="band", target_dimension="temporal") -> ?
```

m-mohr:
Fixes #357


dthiex (approved):

LukeWeidenwalker (approved):
lgtm

Issue #402, opened 2023-02-01
merge_cubes: resample implicitly


m-mohr:
**Process ID:** merge_cubes

**Describe the issue:**

Jeroen wrote at https://discuss.eodc.eu/t/terrascope-backend-upgrade/542:
> merge_cubes and mask will now try to automatically resample input data cubes if required, reducing the need to always use resample_cube_spatial
> When catalogs contain 2 versions of the same product, we now detect this better and try to use the most recent version. This improves performance, but also reproducibility.

This is not ideal wrt reproducibility, I think. Different back-ends behave differently then.
It's non breaking as it purely adds behavior for what before just threw an error.

**Proposed solution:**

We discussed this in the PSC and it seems reasonable here to be user-friendly and implicitly add the resampling to the merge_cubes process in general. It would be described just as an implicit call to `resample_cube_spatial` where the first cube is always the baseline.

soxofaan:
This feature could also be considered for the `mask` processes

PR #405, opened 2023-02-03
Implicit resampling for spatial dimensions in mask and merge_cubes + clarifications

### LINKED ISSUE #379: `merge_cubes` description
(here we are again, discussing `merge_cubes`' description :smile: )

a couple of notes on the current description of `merge_cubes`

> The data cubes have to be compatible

I don't think we properly defined what "compatible" means. I also think the first sentence of the description here should be a more verbose version of the summary and at least mention the verb "merge".

> A merge operation without overlap should be reversible with (a set of) filter operations for each of the two cubes.

I think it's a bit strange to have this as second sentence of the description. We haven't properly defined "overlap" yet. I guess the sentence tries to define "merge_cubes" as the inverse of "filter_*", but don't think that is very clear at the moment.

> The process performs the join on overlapping dimensions, with the same name and type.

I think using the wording "join on a dimension" is not ideal because "join on" is classic  relational database terminology and is closer to "concatenate" than "merge" (which is actually meant here)

> An overlapping dimension has the same name, type, reference system and resolution in both dimensions, but can have different labels.

There is something wrong here: "A ... dimension has the same name ... in both dimensions". I guess what is meant here:

> Overlapping dimensions have the same name, type, reference system and resolution, but can have different labels

But then: isn't overlap about having the *same labels*, so why does this talk about *different labels*?

> One of the dimensions can have different labels, for all other dimensions the labels must be equal. If data overlaps, the parameter overlap_resolver must be specified to resolve the overlap.

- It seems a bit arbitrary that only one dimension can have different labels, and all the others must be equal. E.g. as spatial dimensions usually come as a pair (x and y): isn't it allowed to merge to cubes from different bboxes?
- "all other dimensions the labels must be equal" seems to imply that there is always overlap, while the next sentence contradicts this.

m-mohr:
`mask` and `merge_cubes`: The spatial dimensions `x` and `y` can now be resampled implicitly instead of throwing an error.
Also, clarify a couple of inconsistencies and conflicting descriptions. Depending on what the reader assumed as the truth, this might be breaking.

Implements #402
Fixes #379


soxofaan (commented):
some quick notes

merge_cubes.json
```diff
@@ -1,22 +1,22 @@
...
... the merged data cube: An overlap resolver is *not needed*, but for each data cube you need to add a new dimension using ``add_dimension()``. The new dimensions must be equal, except that the labels for the new dimensions must differ. The merged data cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.\n   2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all values. The merged data cube has the same dimensions and labels as the original data cubes, but all values have been processed by the overlap resolver.\n4. A data cube with dimensions (`x`, `y`, `t` / `bands`) or (`x`, `y`, `t`, `bands`) and another data cube with dimensions (`x`, `y`) have the same dimension labels in `x` and `y`. Merging them will join dimensions `x` and `y`, so the lower dimension cube is merged with each time step and band available in the higher dimensional cube. A use case for this is applying a digital elevation model to a spatio-temporal data cube. An overlap resolver is *required* to resolve the overlap for all pixels.",
```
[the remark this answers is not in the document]
soxofaan: This expression (the first sentence of the description) just doesn't make sense to me:

> merge cubes by performing a join on dimensions

The "perform a join" sound very vague and I have no idea how to interpret it. So that's why my suggestion contains more concrete terms like "concatenation" and "data fusion" (which are the main use cases for `merge_cubes` usage at the moment).

merge_cubes.json
```diff
@@ -1,22 +1,22 @@
...
```
dthiex, line 4: > but the labels for the dimension `bands` are `B1` and `B2` for the base data cube and `B3` and `B4`.

I think we are missing here that B3/B4 belong to the cube to be merged/the other.

dthiex (approved):
From simply reading this both descriptions are clear to me/understandable (would still add the small addition from my other comment).

merge_cubes.json
m-mohr:
```suggestion
    "description": "The process merges two 'compatible' data cubes by performing a join on dimensions that are not equal.\n\nThe data cubes have to be compatible, which means that they must share a common subset of equal dimensions. To conveniently get to such a subset of equal dimensions, the process tries to align the horizontal spatial dimensions (axes `x` and `y`) implicitly with ``resample_cube_spatial()`` if required. `cube1` is the target data cube for the resampling. The equality for geometries follows the definition in the Simple Features standard by the OGC.\n\nAll dimensions share the same properties, such as name, type, reference system, and resolution. Dimensions can have disjoint or overlapping labels. If there is any overlap between the dimension labels, the parameter `overlap_resolver` must be specified to combine the two values for these overlapping labels. A merge operation without overlap should be reversible with (a set of) filter operations for each of the two cubes, if no implicit resampling was applied.\n\nIt is not possible to merge a vector and a raster data cube. Merging vector data cubes with different base geometry types (points, lines/line strings, polygons) is not possible and throws the `IncompatibleGeometryTypes` exception. The base geometry types can be merged with their corresponding multi geometry types.\n\nAfter the merge, the dimensions with a natural/inherent label order (with a reference system this is each spatial and temporal dimensions) still have all dimension labels sorted. For other dimensions without inherent order, including bands, the dimension labels keep the order in which they are present in the original data cubes, and the dimension labels of `cube2` get appended to the dimension labels of `cube1`.\n\n**Examples for merging two data cubes:**\n\n1. Data cubes with the dimensions (`x`, `y`, `t`, `bands`) have the same dimension labels in `x`, `y` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the base data cube and `B3` and `B4` for the other. An overlap resolver is *not needed*. The merged data cube has the dimensions `x`, `y`, `t`, `bands`, and the dimension `bands` has four dimension labels: `B1`, `B2`, `B3`, `B4`.\n2. Data cubes with the dimensions (`x`, `y`, `t`, `bands`) have the same dimension labels in `x`, `y` and `t`, but the labels for the dimension `bands` are `B1` and `B2` for the base data cube and `B2` and `B3` for the other. An overlap resolver is *required* to resolve overlap in band `B2`. The merged data cube has the dimensions `x`, `y`, `t` and `bands` and the dimension `bands` has three dimension labels: `B1`, `B2`, `B3`.\n3. Data cubes with the dimensions (`x`, `y`, `t`) have the same dimension labels in `x`, `y` and `t`. There are two options:\n   1. Keep the overlapping values separately in the merged data cube: An overlap resolver is *not needed*, but for each data cube you need to add a new dimension using ``add_dimension()``. The new dimensions must be equal, except that the labels for the new dimensions must differ. The merged data cube has the same dimensions and labels as the original data cubes, plus the dimension added with ``add_dimension()``, which has the two dimension labels after the merge.\n   2. Combine the overlapping values into a single value: An overlap resolver is *required* to resolve the overlap for all values. The merged data cube has the same dimensions and labels as the original data cubes, but all values have been processed by the overlap resolver.\n4. A data cube with dimensions (`x`, `y`, `t` / `bands`) or (`x`, `y`, `t`, `bands`) and another data cube with dimensions (`x`, `y`) have the same dimension labels in `x` and `y`. Merging them will join dimensions `x` and `y`, so the lower dimension cube is merged with each time step and band available in the higher dimensional cube. A use case for this is applying a digital elevation model to a spatio-temporal data cube. An overlap resolver is *required* to resolve the overlap for all pixels.",
```

merge_cubes.json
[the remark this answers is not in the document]
m-mohr: "concatenation" doesn't work because it's not always just an "append". Maybe we should just leave it at "The process combines two 'compatible' data cubes to a single data cube."

mask.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Applies a mask to a raster data cube. To apply a polygon as a mask, use ``mask_polygon()``.\n\nA mask is a raster data cube for which corresponding pixels among `data` and `mask` are compared and those pixels in `data` are replaced whose pixels in `mask` are non-zero (for numbers) or `true` (for boolean values). The pixel values are replaced with the value specified for `replacement`, which defaults to `null` (no data).\n\nThe data cubes have to be compatible except that the horizontal spatial dimensions (axes `x` and `y`) will be aligned implicitly by ``resample_cube_spatial()`` with `data` being the target data cube. All other dimensions in the mask must also be available in the raster data cube with the same name, type, reference system, resolution and labels. Dimensions can be missing in the mask with the result that the mask is applied to each label of the dimension in `data` that is missing in the data cube of the mask. The process fails if there's an incompatibility found between the raster data cube and the mask.",
```
LukeWeidenwalker, line 4: I know, implementation detail, but does this mean that the client will check for this and add a `resample_cube_spatial` node at the right place in the graph, or do we need to handle this in the `merge_cubes` implementation?

m-mohr: in merge_cubes

LukeWeidenwalker:
I understand that the reproducability issue with this change has already been discussed (and accepted) in the PSC. With that disclaimed, I still feel obligated to ask: what exactly is the principle by which it's okay for this process to violate reproducability? And is this really so difficult to fix? What would be wrong about a `resampler` parameter, that at least explicitly encodes what's going on as a sub-process-graph? If that's too much effort, `resample_cube_spatial` lists loads of resampling methods, couldn't we then at least pick one from those and commit `merge_cubes` in its current form to always use that (under the assumption that the resampling method used is the only source of ambiguity here)? 
Let me know if I'm missing something here!

Btw, if folks are totally over this discussion and this is deemed not worth the effort by everyone, just ignore this comment - the content of the PR looks fine to me!

m-mohr:
@LukeWeidenwalker This has not been accepted by the PSC yet, we just discussed it but did not vote on it yet.

I'm not 100% sure why this is hurting reproducibility? What we propose here is to just implicitly call the resample process with default parameters if required. As such this should be as reproducible as an explicit call to the resample process.
I guess it's not very explicit in the process description that this is meant to use the default parameters?

LukeWeidenwalker:
> I'm not 100% sure why this is hurting reproducibility? 

Because without more info this doesn't specify which resampling method was used!

> What we propose here is to just implicitly call the resample process with default parameters if required. As such this should be as reproducible as an explicit call to the resample process. I guess it's not very explicit in the process description that this is meant to use the default parameters?

Ah, yes, IMHO this should be explicit. And since the default parameters can change, I'd also hard-code what this default parameter is expected to be. E.g. "The implicit resampling is performed as a `resample_cube_spatial` with `method="near"`".

m-mohr:
Okay, I'll make this more explicit. I'm not suire whether It's not a good idea to also mention the default values.
I assume by default back-ends don't change the default and if they change it then due to the reason that they don't support it. This means they could not implement this functionality here either and need to change the description, too. So it's not reproducible anyway...

PR #412, opened 2023-03-08
Streamline GeoJSON import #346

m-mohr:
Streamlines the GeoJSON import #346

The load_geojson function probably needs a bit better description on how to construct the data cube.


clausmichele:
If I understand correctly, after that this PR will be merged, when a user will select a polygon aoi in a `load_collection` process in the Web Editor there will be two processes added: `load_geojson` and `load_collection` ?

m-mohr:
In the Web Editor it would complicate things a bit as you'd need to add the load_geojson specifically to the model builder and then connect it to the temporal_extent parameter, but on the other hand it simplifies, because you can reuse it in subsequent processes like aggregate_spatial.

soxofaan:
I just had a quick look at this PR, but this looks like a heavily breaking change.
Can't we just first add the `load_geojson` option in non-breaking way (keep inline geosjon option), 
and then in a later phase remove the original inline geojson option?

m-mohr:
You can always still support geojson by adding the missing schema again. Having it still in here in 2.0 as deprecated would also mean new implementations would face it, which is not ideal.

m-mohr:
- Deprecate in stable processes
- Don't add to new processes
- Remove from proposals

8+

proposals/load_geojson.json
```diff
@@ -0,0 +1,68 @@
...
+                    "subtype": "file-path",
+                    "pattern": "^[^\r\n\\:'\"]+$"
+                }
+            ]
+        },
+        {
+            "name": "properties",
+            "description": "One or more properties from the GeoJSON file to construct a additional dimension from. Only applies for GeoJSON Features and FeatureCollections. Missing values in are generally set to no-data (`null`).\n\nDepending on the number of properties provided the process constructs and names the dimension differently:\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nDepending on the number of properties the dimension name is:\n\n- Single property: Name of the property\n- Multiple properties: `properties`.",
```
thomasstorm, line 38:
```suggestion
            "description": "One or more properties from the GeoJSON file to construct an additional dimension from. Only applies for GeoJSON Features and FeatureCollections. Missing values are generally set to no-data (`null`).\n\nDepending on the number of properties provided the process constructs and names the dimensions differently:\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nDepending on the number of properties the dimension name is:\n\n- Single property: Name of the property\n- Multiple properties: `properties`.",
```

m-mohr: Thanks.

LukeWeidenwalker (approved):

proposals/load_geojson.json
```diff
@@ -0,0 +1,68 @@
...
+                    "subtype": "file-path",
+                    "pattern": "^[^\r\n\\:'\"]+$"
+                }
+            ]
+        },
+        {
+            "name": "properties",
+            "description": "One or more properties from the GeoJSON file to construct an additional dimension from. Only applies for GeoJSON Features and FeatureCollections. Missing values are generally set to no-data (`null`).\n\nDepending on the number of properties provided the process constructs and names the dimension differently:\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nDepending on the number of properties the dimension name is:\n\n- Single property: Name of the property\n- Multiple properties: `properties`.",
```
soxofaan, line 38: >    Depending on the number of properties provided the process constructs and names the dimension differently:

What then follows is a discussion about the dimension labels (not dimension names).
Explanation of the dimension name is in another paragraph

soxofaan: > Depending on the number of properties the dimension name is:\n\n- ... Multiple properties: properties.",

related to the discussion about dim name geometries/geometry (https://github.com/Open-EO/openeo-api/issues/479 settled on singular geometry), this should be singular "property" as well, right?

m-mohr:
Applies the deprecation as discussed. Will create a new PR for load_geojson.

proposals/load_geojson.json
m-mohr: 1) True, improved.
2) Not sure, it was just about recommended names. There we opted for geometry due to the Python ecosystem, but here we are clearly following GeoJSON / SF terminology.

Issue #384, opened 2022-09-14
load_ml_model id: overlap between batch job id and file path?


soxofaan:
https://github.com/Open-EO/openeo-processes/blob/be148835a725966fbf1fff65215dbf2bca82a4a7/proposals/load_ml_model.json#L23-L33

Batch job id has regex `"^[\\w\\-\\.~]+$"` and user-uploaded file has regex `"^[^\r\n\\:'\"]+$"`.
Both regexes will match on simple things like `foo-bar`, so a back-end will not be able to determine if given `id` is intended as job id or a user uploaded file? Or is a back-end supposed to try the available options and pick the first match?

m-mohr:
True, that's not ideal.

m-mohr:
The only thing I could think of right now is to require e.g. a "./" at the beginning of file paths.
Alternatively, back-ends would need to be given a priority list, I guess... Any opinions or ideas?

soxofaan:
another option is requiring some kind of  scheme prefix when there is risk of confusion, e.g. `workspace://mymodel.foo`,  which extends easily to other storage solutions or "namespaces" (`s3://...`, `https://...`, ... or even `job:...` and `job://otherbackend/...`)

m-mohr:
Yeah, I thought about that too, but I found that even more confusing/unexpected. 🤔

m-mohr:
Maybe define two processes?

m-mohr:
Solved by #413, I think.

PR #413, opened 2023-03-08
load_stac and remove job-id subtype #384

m-mohr:
To avoid ambiguities, I think we can remove the batch job id subtype. A batch job ID can also just be provided via the uri subtype. Either you provide the canonical (i.e. "public") link and then it's just like any external data. Or you can simply provide a URI such as `https://example.com/api/v1.0/jobs/12345/results` and then you can easily detect the job id from it. The clients can help with it by allowing Job objects as input for example.


LukeWeidenwalker (approved):
I like this change! Subsuming the openeo-internal "result" concept into the generally-applicable "datacube-described-by-STAC" concept (which exists outside of openeo) makes total sense to me.

What prevents us from doing something similar for `load_collection`? At the end of the day each collection is also just a STAC collection, right? Is the "collection" concept in openeo anything other than "STAC collection openeo platform knows about and has an explicit name for"? The `geometry` parameter that `load_collection` has could just as well be added to `load_stac`. 

(If my understanding of the collection abstraction is somehow inaccurate, I'd also love to hear in what way 😇)

jdries:
@LukeWeidenwalker  FYI, not each openEO collection is a STAC collection. In Terrascope backend for instance, the majority of collections is based on opensearch or even simply finding tif or netcdf files on disk by globbing.

m-mohr:
Change id to url.

m-mohr:
Ready for review.

proposals/load_stac.json
```diff
@@ -1,35 +1,27 @@
...
...\nA batch job result can be loaded by providing the URL to the STAC metadata for a batch job result. This is usually a signed URL that is provided by some back-ends since openEO API version 1.1.0 through the `canonical` link relation in the batch job result metadata.\n\nIf supported by the underlying metadata and file format, the data that is added to the data cube can be restricted with the parameters `spatial_extent`, `temporal_extent` and `bands`. If no data is available for the given extents, a `NoDataAvailable` exception is thrown.\n\n**Remarks:**\n\n* The bands (and all dimensions that specify nominal dimension labels) are expected to be ordered as specified in the metadata if the `bands` parameter is set to `null`.\n* If no additional parameter is specified this would imply that the whole data set is expected to be loaded. Due to the large size of many data sets, this is not recommended and may be optimized by back-ends to only load the data that is actually required after evaluating subsequent processes such as filters. This means that the values should be processed only after the data has been limited to the required extent and as a consequence also to a manageable size.",
```
soxofaan, line 4: > This is usually a signed URL that is provided by some back-ends since openEO API version 1.1.0 through the `canonical` link relation in the batch job result metadata.

maybe I'm misreading, but I can't find this "canonical" feature in the 1.1.0 of the API, only in current draft. Am I looking wrong?

m-mohr: Indeed, it should be 1.2.0.

soxofaan (approved):

PR #420, opened 2023-03-14
New definitions for fit_curve and predict_curve

m-mohr:
Make fit_curve usable in apply_dimension, scrap gap filling for predict_curve, clarify no data handling, etc.


LukeWeidenwalker:
@ValentinaHutter you know this better than I do! :)

m-mohr:
It was meant to cater for invalid values in the input data, but not sure whether and how we should cater for other cases?

proposals/fit_curve.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "fit_curve",
     "summary": "Curve fitting",
-    "description": "Use non-linear least squares to fit a model function `y = f(x, parameters)` to data.\n\nThe process throws an `InvalidValues` exception if invalid values are encountered. Invalid values are finite numbers (see also ``is_valid()``).",
+    "description": "Use non-linear least squares to fit a model function `y = f(x, parameters)` to data.\n\nThe process throws an `InvalidValues` exception if invalid values are encountered. Valid values are finite numbers (see also ``is_valid()``).",
```
soxofaan, line 4:
```suggestion
    "description": "Use non-linear least squares to fit a model function `y = f(x, parameters)` to data.\n\nThe input data cube should only contain finite numerical values (also see ``is_valid()``), otherwise an `InvalidValues` error is thrown.",
```
?

soxofaan (commented):
just a suggestion

clausmichele:
This process must be able to handle invalid data internally and should not return an error if some invalid points are present in the time series.

m-mohr:
@clausmichele So you are saying we should remove the `InvalidValues` exception and all mentions of it and instead say that invalid values must be dealt with smoothly? Is this possible in all cases?

clausmichele:
Indeed. I can't think about _all_ the possible scenarios, but the one for which we created this process, fitting an harmonic curve to each pixel of a raster-cube, must handle NaNs occurring due to cloud masked or or invalid pixels along the time series.
Should we make this part explicit? Adding a default `ignore_nodata = true` ? I don't know if there could be a case where the user wants to be warned if there are invalid values in the raster-cube/time series.

soxofaan:
Doesn't it actually all depend on how the callback provided in the `function` parameter works?
https://github.com/Open-EO/openeo-processes/blob/0dd3ab0d81f67506547136532af39b5c9a16771e/proposals/fit_curve.json#L38-L43

If that function properly supports handling invalid data, then the whole `fit_curve` call will work. If it doesn't, then `fir_curve` will fail when given data that contains invalid values.

So this seems like to user's responsibility to make it work with or without invalid data support?


Or,  would it be a feature of `fit_curve`  to skip parts of the raster data cube with invalid values and only provide `function`  the parts where the `x` input is fully valid?

clausmichele:
@soxofaan yes exactly! So `fit_curve` shouldn't check the input data, but it's the fitting `function` that takes care of it.

m-mohr:
I cherry-picked the single word change from this PR which seemed okay. The issues that came up now are somewhat related, but not strictly related to the intial PR. So having the first action done, we can now discuss additional changes here independantly.

m-mohr:
Example from discussions:

labels = dimension_labels(datacube)
seconds = array_apply(labels, fn = to_seconds)
cube2 = apply_dimension(datacube, fn, "t", seconds)

fn fit:
	params = fit_curve(context.seconds, [1,1,1], fn);


predictions = predict_curve(cube2)

mkadunc (approved):

proposals/fit_curve.json
```diff
@@ -1,38 +1,33 @@
...
             "name": "data",
-            "description": "A data cube.",
+            "description": "An array.",
             "schema": {
-                "type": "object",
-                "subtype": "datacube"
+                "type": "array",
+                "items": {
+                    "type": "number"
+                }
```
soxofaan, line 18: I think we are still missing a source for the `x` values here. Possible solutions:
- `data` is a labeled array, with x values as labels
- add another array parameter for the `x` values (and rename `data` to `y` for clarity?)

clausmichele: A labeled array is what was used already, so I vote for that.

m-mohr: Updated.

Issue #409, opened 2023-02-27
order: clarify ordering of ties


LukeWeidenwalker:
**Process ID:** order

**Describe the issue:**
The simplest implementation to support the `asc` parameter is to simply flip the input array before running argsort if descending order is desired. However, the `order` process currently requires that `Ties will be left in their original ordering`, regardless of whether `asc=True` or `asc=False`. This means that we cannot simply flip the ~input array~ array with the ordered indices, because ties would be flipped too. This makes the implementation more unwieldy than it has to be.

**Proposed solution:**
Change this requirement such that ties will be in the original ordering if asc=True and flipped otherwise.

**Additional context:**
See [here](https://github.com/Open-EO/openeo-processes-dask/pull/65#discussion_r1116752033) for initial discussion

m-mohr:
@soxofaan @dthiex  Do you know how your implementations handle this?

soxofaan:
we currently use this implementation from the legacy openeo-processes-python repo: https://github.com/Open-EO/openeo-processes-python/blob/f57c663e8708061a8222851a8aec2762ac46b0bb/src/openeo_processes/arrays.py#L745-L817

which behaves like the current spec if I understand correctly:
```python
import numpy as np

data = np.array([1, 2, 2, 3])
print(data)
# [1 2 2 3]

print(numpy.argsort(data, kind="mergesort"))
# [0 1 2 3]
print(numpy.argsort(-data, kind="mergesort"))
# [3 1 2 0]
```

soxofaan:
> The simplest implementation to support the asc parameter is to simply flip the input array before running argsort if descending order is desired. However, the order process currently requires that Ties will be left in their original ordering, regardless of whether asc=True or asc=False. This means that we cannot simply flip the input array, because ties would be flipped too.

maybe I'm misunderstanding, but if you mean flipping the sign of the input data with "flipping the array", I think the descending mode works just fine with `numpy.argsort` as illustrated above

LukeWeidenwalker:
Apologies, typo in my description:  
I mean after applying `argsort`, flipping (aka reversing) the array with the ordered indices with `np.flip` to get the opposite ordering.

m-mohr:
If it works with argsort in the old implementation, why do we need a change here? Can you give some examples, @LukeWeidenwalker?

LukeWeidenwalker:
Adding the minus-sign like in the old implementation only works on numerical arrays:
`-data`

[...]

PR #424, opened 2023-03-15
Improve summary of order process

soxofaan:
spin-off from #409   :stuck_out_tongue:

### DIFF
--- a/order.json
+++ b/order.json
@@ -1,7 +1,7 @@
 {
     "id": "order",
-    "summary": "Get the order of array items",
-    "description": "Computes a permutation which allows rearranging the data into ascending or descending order. In other words, this process computes the ranked (sorted) element positions in the original list.\n\n**Remarks:**\n\n* The positions in the result are zero-based.\n* The ordering of ties is implementation-dependent.\n* Temporal strings can *not* be compared based on their string representation due to the time zone/time-offset representations.",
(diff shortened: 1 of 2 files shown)


m-mohr:
A change here should probably be aligned with rearrange.

For me in this proposal it's not clear what the order is. This sounds a bit like whether it would return "ascending" or "descending" or "unsorted" and not the indices for the array in a sorted order ;-)

soxofaan:
> A change here should probably be aligned with rearrange.

Because `rearrange` uses "permutation" in its summary?  I think that's fine.

> For me in this proposal it's not clear what the order is. This sounds a bit like whether it would return "ascending" or "descending" or "unsorted" 

Ok fair. An alternative, stealing from numpy argsort:

> Get indices that would sort the array


FYI: what I don't like about the current summary "Create a permutation":
- "create a" sound like the process just randomly generates some permutation, regardless of the original data
- "permutation": it's not always obvious if "permutation" refers to the set of indices to implement the rearrangement, or if it is about the result after rearranging 
so "create a permutation" roughly sounds like "randomly shuffle the array" to me

m-mohr:
@soxofaan I made a couple of changes to the PR, especially to link between the processes. Can you check whether this works for you?

soxofaan:
yes it's better now

apparently I can approve this PR, I guess because I started it myself

Issue #415, opened 2023-03-10
Generalize `load_geojson`


soxofaan:
I considered raising this in #412 or #322 but maybe it's better to keep this a separate thread:

`load_geojson` (as introduced in #412)  is closely tied to GeoJSON (by name and supported format), which has the following problems:
- it's out of line with other `load_` processes which support multiple formats, and use `/file_formats` to express that
- it supports loading data from inlined data in the process graph and from remote URLs, but remote vector data might also come in other formats (geopackage, shapefiles, next generation GeoJSON, ...), so why not support that too? (Inline data can also come in other formats actually: WKT string, CovJSON, ...)

Couple of lines of thought:
- generalize `load_geojson`, e.g. to `load_vector`, to also support other vector formats
- generalize  even further to e.g. `load_cube` to also support raster data, not only something like a remote GeoTIFF, but also some kind of inline raster data cube construct (also see #204)
- generalize, but split the inline and external URL case (to avoid some schema specification issues) :
    -  `load_inline`  which could support inline GeoJSON, WKT, other JSON based vector format (e.g. CovJson),  some kind of inline raster data format (again see #204), ...
    - `load_external` to load from a URL

jdries:
A reminder perhaps, but another issue with a specific process like load_geojson is in UDP's where the data to load is a parameter, and the UDP designer would like a user friendly approach to allow loading vector data from different sources without having to make a complex construction inside the UDP.
Could potentially be solved by simply having a vector cube as a udp parameter but then we do have more reliance on the ability to construct vector cubes loading logic client side.

m-mohr:
@jdries Yes, a vector cube should probably be passed in. geojson as subtype has be deprecated (except for load_geojson, if accepted).

PR #427, opened 2023-03-31
Add load_geojson

m-mohr:
Implements #346 and #415


proposals/load_geojson.json
```diff
@@ -0,0 +1,51 @@
+{
+    "id": "load_geojson",
+    "summary": "Load GeoJSON data",
+    "description": "Loads GeoJSON data as defined by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946.html) into a vector data cube. Feature properties are preserved.",
```
soxofaan, line 4:
```suggestion
    "description": "Loads inline GeoJSON data as defined by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946.html) into a vector data cube. Feature properties are preserved.",
```

proposals/load_geojson.json
```diff
@@ -0,0 +1,51 @@
...
+            "schema": {
+                "type": "object",
+                "subtype": "geojson"
+            }
+        },
+        {
+            "name": "properties",
+            "description": "One or more properties from the GeoJSON file to construct an additional dimension from. Only applies for GeoJSON Features and FeatureCollections. Missing values are generally set to no-data (`null`).\n\nDepending on the number of properties provided, the process creates the dimension differently:\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nDepending on the number of properties the dimension name is:\n\n- Single property: Name of the property\n- Multiple properties: `properties`.",
```
soxofaan, line 21: So at least one property should be specified here?
what if the geojson has no associated properties (e.g. a single `Polygon`, or `FeatureCollection` where all `properties` are empty?)

proposals/load_geojson.json
m-mohr: I'm wondering whether load is the right term. We should probably stick to it in the id, but maybe in the description we can say something like "Processes inline GeoJSON" or so?

m-mohr: Updated

mkadunc (approved):

LukeWeidenwalker (approved):

proposals/load_geojson.json
```diff
@@ -0,0 +1,53 @@
...
+    "description": "Converts GeoJSON data as defined by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946.html) into a vector data cube. Feature properties are preserved.",
+    "categories": [
+        "import",
+        "vector"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "A GeoJSON object to inline into the process. The GeoJSON type `GeometryCollection` is not supported. Each geometry in the GeoJSON data results in a dimension label in the `geometries` dimension.",
```
soxofaan, line 13: > A GeoJSON object to inline into the process.

Statement looks a bit weird to me: the object is not to be inlined into the process, it is inlined in the process graph and is to be converted to a vector cube.

```suggestion
            "description": "A GeoJSON object to convert to a vector cube. The GeoJSON type `GeometryCollection` is not supported. Each geometry in the GeoJSON data results in a dimension label in the `geometries` dimension.",
```

proposals/load_geojson.json
```diff
@@ -0,0 +1,53 @@
...
+            "schema": {
+                "type": "object",
+                "subtype": "geojson"
+            }
+        },
+        {
+            "name": "properties",
+            "description": "A list of properties from the GeoJSON file to construct an additional dimension from. A new dimension is created if at least one property is provided. Only applies for GeoJSON Features and FeatureCollections. Missing values are generally set to no-data (`null`).\n\nDepending on the number of properties provided, the process creates the dimension differently:\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nDepending on the number of properties the dimension name is:\n\n- Single property: Name of the property\n- Multiple properties: `properties`.",
```
soxofaan, line 21: Some notes about the "properties" parameter. A bit nitpicking, so not necessarily blockers to merge already (can be moved to follow up ticket).

> Single property with scalar values: ...
> Single property of type array: ...

This assumes the a feature collection is internally consistent: the property is available for each geometry and  it has same type for each geometry. When this is not the case: should this raise an error? Or should we describe type coercion rules: e.g. missing/null -> empty array,  scalar value -> array with one item, so that everything can be described in terms of arrays.

> Depending on the number of properties the dimension name is:\n\n- Single property: Name of the property\n- Multiple properties: `properties`.

I find it a bit strange to change the dimension name based on number of labels. We don't do that with other dimensions I think

soxofaan (commented):
some more nitpicking notes (not necessarily to block merging current state already)

proposals/load_geojson.json
m-mohr:
```suggestion
            "description": "A GeoJSON object to convert to a vector data cube. The GeoJSON type `GeometryCollection` is not supported. Each geometry in the GeoJSON data results in a dimension label in the `geometries` dimension.",
```

m-mohr: Updated.

proposals/load_geojson.json
m-mohr: > This assumes the a feature collection is internally consistent: the property is available for each geometry and it has same type for each geometry. When this is not the case: should this raise an error? Or should we describe type coercion rules: e.g. missing/null -> empty array, scalar value -> array with one item, so that everything can be described in terms of arrays.

As we also discusssed in the last meeting, openEO allows mixed-type data cubes, so in theory it should not throw an error, but if a "back-end limitation" requires consistent types and they are not consistent, yes, the implementation should throw an error.

> I find it a bit strange to change the dimension name based on number of labels. We don't do that with other dimensions I think

Changed to be always `properties` (and type: 'other').

m-mohr: If you think there's more room for improvements, please open an issue as I'll merge this now. Thanks :-)

Issue #155, opened 2020-06-09
raster <-> vector processes


jdries:
For an internal use case, we'd like to apply raster to vector conversion:
1. input is a raster datacube
2. output is a 'vector datacube'  that can be exported as geojson

This is not uncommon, so we'd like to see if we can provide this as higher level process. 
We will probably implement something internally first, to see how it can look like.

m-mohr:
As soon as you have a process definition I'd suggest to open a draft PR so that we can shape it together.

jdries:
Some inspiration for naming:

- GDAL has gdal_polygonize
- GEE has reduceToVector, and also has behaviour for the multiband case https://developers.google.com/earth-engine/reducers_reduce_to_vectors
- GRASS GIS has r.to.vect https://grass.osgeo.org/grass78/manuals/r.to.vect.html

m-mohr:
Can we model this as reducer in openEO? If not, reduce would be a bit confusing, I think. polygonize only works if it's always working on polygons only. I think I'd prefer something easy to understand and aligned with other processes such as `create_raster_cube` like `vector_cube`, `to_vector` or `convert_cube`. I'm not set on anything yet.

m-mohr:
@jdries We should complete the description and PR it for discussion.

PondiB:
@jdries  and @m-mohr, is this being worked on?

m-mohr:
VITO has an example implementation, but we still need to get a version into openeo-processes. @PondiB

PondiB:
Cool, I will pick it up then.

m-mohr:
Cool, please consider a survey of existing implementations, e.g. at VITO (already openEO), GRASSGIS, Python ecosystem (shapely?), R, ... @PondiB

PondiB:
@m-mohr , Sure, will do that.

PR #434, opened 2023-04-25
Raster to vector Conversion

PondiB:
closes issue #155 

An optional "projection" parameter has been inspired by the Rasterio function and GEE.  Not sure if we should keep it.

### DIFF
--- a/proposals/raster_to_vector.json
+++ b/proposals/raster_to_vector.json
@@ -1,7 +1,7 @@
 {
     "id": "raster_to_vector",
-    "summary": "Converts raster data to vector data",
-    "description": "Converts a raster data cube into a vector data cube. The bounding polygon of similar areas of pixels is constructed.",
+    "summary": "Converts single band raster data cube to vector data",
+    "description": "Converts a single band raster data cube into a vector data cube. The bounding polygon of uniform areas of pixels is constructed.",
     "categories": [
         "cubes"
     ],
@@ -17,7 +17,7 @@
         },
         {
             "name": "projection",
-            "description": "Coordinate reference system to reproject the converted vector. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\nThe default value is of the  input raster cube.",
+            "description": "Coordinate reference system to reproject the converted vector. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\n When no projection (or null) is given, the reference system of the input raster data cube is used.",
             "schema": [
                 {
                     "title": "EPSG Code",
--- a/proposals/raster_to_vector.json
+++ b/proposals/raster_to_vector.json
@@ -1,6 +1,6 @@
 {
     "id": "raster_to_vector",
-    "summary": "Converts a raster to a vector",
+    "summary": "Converts raster data to vector data",
     "description": "Converts a raster data cube into a vector data cube. The bounding polygon of similar areas of pixels is constructed.",
     "categories": [
         "cubes"
--- a/proposals/raster_to_vector.json
+++ b/proposals/raster_to_vector.json
@@ -39,7 +39,7 @@
         }
     ],
     "returns": {
-        "description": "Returns a converted vector data cube.",
+        "description": "A converted vector data cube.",
         "schema": {
             "type": "object",
             "subtype": "datacube",
--- a/proposals/raster_to_vector.json
+++ b/proposals/raster_to_vector.json
@@ -17,7 +17,7 @@
         },
         {
             "name": "projection",
-            "description": "Coordinate reference system to reproject the converted vector. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\n When no projection (or null) is given, the reference system of the input raster data cube is used.",
+            "description": "Coordinate reference system to reproject the converted vector to. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\nWhen no projection (or null) is given, the reference system of the input raster data cube is used.",
             "schema": [
                 {
                     "title": "EPSG Code",
--- a/proposals/raster_to_vector.json
+++ b/proposals/raster_to_vector.json
@@ -34,7 +34,7 @@
                     "subtype": "wkt2-definition"
                 },
                 {
-                    "type": null
+                    "type": "null"
                 }
             ],
             "optional": true,
--- a/proposals/raster_to_vector.json
+++ b/proposals/raster_to_vector.json
@@ -34,8 +34,7 @@
                     "subtype": "wkt2-definition"
                 }
             ],
-            "optional": true,
-            "default": null
+            "optional": true
         }
     ],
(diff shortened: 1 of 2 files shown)


PondiB:
FYI; I'll try and figure out what causes the failures.

m-mohr:
@PondiB experimental: true is missing, remove dot at the end of summary and a spellcheck issue :-)

PondiB:
> @PondiB experimental: true is missing, remove dot at the end of summary and a spellcheck issue :-)

Aha, awesome, thanks @m-mohr

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,53 @@
+{
+    "id": "raster_to_vector",
+    "summary": "Converts a raster to a vector",
```
soxofaan, line 3: I don't think "a raster" and "a vector" is proper terminology. I would at least say 
```suggestion
    "summary": "Converts raster data to vector data",
```

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,53 @@
...
+    "description": "Converts a raster data cube into a vector data cube. The bounding polygon of similar areas of pixels is constructed.",
```
soxofaan, line 4: What does "similar" mean? It suggests there is some room for small deviations, but then there should be a parameter to set thresholds I think. Or a parameter to set the number of desired quantization levels.

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,53 @@
...
+                    "type": "string",
+                    "subtype": "wkt2-definition"
+                }
+            ],
+            "optional": true,
+            "default": null
+        }
+    ],
+    "returns": {
+        "description": "Returns a converted vector data cube.",
```
soxofaan, line 42: I don't think that the description should use the verb "returns"

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,53 @@
...
+            "name": "data",
+            "description": "A data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
+            }
+        },
+        {
+            "name": "projection",
+            "description": "Coordinate reference system to reproject the converted vector. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\nThe default value is of the  input raster cube.",
```
soxofaan, line 20: > The default value is of the input raster cube.

doesn't sound very clean. I think it should be something like

> When no projection (or null) is given, the reference system of the input raster data cube is used

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,53 @@
...
+                    ]
+                },
+                {
+                    "title": "WKT2",
+                    "type": "string",
+                    "subtype": "wkt2-definition"
+                }
+            ],
+            "optional": true,
+            "default": null
```
soxofaan, line 38: default is `null` but, that does not match the schema above, is that valid @m-mohr ?

proposals/raster_to_vector.json
soxofaan: Another related question: how are bands handled here? Is each band converted separately, so you have different vector segmentation for each band? If so, I don't think that can represented as a single openEO vector cube, because the geometries are the labels of the "geometry" dimension, so the output would be a set of vector cubes, one for each input band.

Or does segmentation happens on all bands simultaneously,  e.g. RGB-color based segmentation? But then the definition of "similar" or color quantization becomes a lot harder I guess.

Maybe raster_to_vector should only be allowed on single band raster cubes?

proposals/raster_to_vector.json
PondiB: Noted.

proposals/raster_to_vector.json
PondiB: Similar refers to uniform or homogenous. I'll use a uniform in this case for clarity. 

The concerns on how to deal with multiple bands are legit. I have not come across any platform that does that and hence the conversion of a single band raster cube seems ideal.

proposals/raster_to_vector.json
PondiB: Adopting the suggestion:)

proposals/raster_to_vector.json
soxofaan: If you really mean that pixel values should be the same, I would just state that, e.g.

> The bounding polygon of areas with equal pixel values is constructed.

However, this works probably good enough when your input data is integer valued (e.g. classification values),
but when it's float data (e.g. NDVI values), a user probably wants to to define quantization in some way. Is this in scope of this PR, or is that not necessary yet?

proposals/raster_to_vector.json
m-mohr: Indeed, null needs to be specified in the schema explicitly.

PondiB: Noted, I will actually remove the default from being null as when this optional field is not specified then the CRS of the output vector data will be the same as the input raster data.

m-mohr: Optional fields always need a default value though!

PondiB: Okay, so in this guess default will vary based on the projection of the input raster, not sure how I should define that.

soxofaan: The default can be `null`, but you have to add that as one of the options in the schema above

PondiB: Okay noted, will update it to that effect.

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,56 @@
+{
+    "id": "raster_to_vector",
+    "summary": "Converts single band raster data cube to vector data",
+    "description": "Converts a single band raster data cube into a vector data cube. The bounding polygon of uniform areas of pixels is constructed.",
```
m-mohr, line 4:
```suggestion
    "description": "Converts a single band raster data cube into a vector data cube. The bounding polygon of uniform areas of pixels is constructed.",
```

I think this may need some more information about the conversion process. Is this a union polygon of all pixel centers? Or really the fully area of all pixel boundaries?

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,56 @@
...
+        "cubes"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "A data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
```
m-mohr, line 15: This requires at least x and y dimensions, right? Should be added as requirement.

proposals/raster_to_vector.json
```diff
@@ -0,0 +1,56 @@
...
+            }
+        },
+        {
+            "name": "projection",
+            "description": "Coordinate reference system to reproject the converted vector. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\n When no projection (or null) is given, the reference system of the input raster data cube is used.",
```
m-mohr, line 20:
```suggestion
            "description": "Coordinate reference system to reproject the converted vector to. Specified as an [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).\nWhen no projection (or null) is given, the reference system of the input raster data cube is used.",
```

m-mohr (changes requested):
See comments above.

PondiB:
> To be testable, this needs more specific information about how the conversion is done. See comments above.

Factored that in.

Issue #414, opened 2023-03-10
How to handle exceptions in the specification


m-mohr:
Should errors be "optional"?

Is the specification a best-practice and implementations can change (throw an error) or should always keep the option open (may throw an error).

m-mohr:
Candidates:

- apply_kernel
- aggregate_temporal?
- vector units
- log / ln / divide / mod
- merge_cubes? (see also #402)
- ndvi
- save_result

m-mohr:
We discussed this today. The consensus was to be strict in the user-facing documentation (i.e. the JSON process specification) so that errors are clearly stated without a "uncertainty" such as "may throw".  Back-ends can extend their implementations and remove the exception from the documentation. For back-end implementations we add more documentation to the back-end implementation guide (and promote/use it more).

PR #436, opened 2023-05-02
Handle units in vector processes #330

### LINKED ISSUE #330: Unit in vector processes
With mostly all vector processes we hit the point where you need to specify a length/distance/... and that needs to be specified in a specific unit. So we have seen different proposals across the PRs:

1. Always meters
2. Always degree
3. Unit of the SRS (usually either meter or degree)

There are several issues with the options:
1. What does "meters" mean specifically? e.g. https://github.com/Open-EO/openeo-processes/pull/314#discussion_r792528112 This is probably the most user-friendly option?
2. What does "degree" mean specifically? e.g. https://github.com/Open-EO/openeo-processes/pull/315#discussion_r814799187
3. Users need to explicitly understand the SRS, which is the least user-friendly option.

I think it's not a good idea to discuss this for every new process. So what's the preferred option that we adopt in general?

PS: In existing (raster) processes we have:
- apply_neighborhood: unit of srs or m or px
- resample_spatial: unit of srs

m-mohr:
Closes #330 as discussed there and in #414.

I tried to make this as future-proof as possible so that we can decide to allow more than meters in the future.

A general question came up (how to encode values with units), please chime in here: https://github.com/Open-EO/openeo-processes/pull/436#discussion_r1183899826


clausmichele (approved):

proposals/vector_to_regular_points.json
```diff
@@ -23,12 +23,22 @@
...
             "schema": {
                 "type": "number",
                 "minimumExclusive": 0
             }
         },
+        {
+            "name": "unit",
+            "description": "Specifies the unit the distance is given in. The only allowed value is meters (`m`), which is the default value. This parameter exists to be future-proof and might be extended in the future.",
+            "optional": true,
+            "default": "m",
```
LukeWeidenwalker, line 36: Only a nitpick, but I'd prefer this to be `meters` to be extra unambiguous.

LukeWeidenwalker (approved):
lgtm

proposals/vector_to_regular_points.json
m-mohr: We used `m` elsewhere and it's the official SI unit so we should stick to it.

LukeWeidenwalker: Ah good point!

proposals/vector_to_regular_points.json
```diff
@@ -23,12 +23,22 @@
...
             "schema": {
                 "type": "number",
                 "minimumExclusive": 0
             }
         },
+        {
+            "name": "unit",
+            "description": "Specifies the unit the distance is given in. The only allowed value is meters (`m`), which is the default value. This parameter exists to be future-proof and might be extended in the future.",
```
soxofaan, line 34: any reason this parameter wasn't added  to `vector_buffer`?

I'm also not sure it's necessary to define this parameter to be future-proof. Just omitting it now and having meter by default later will also be future proof I think

m-mohr: Yes, we can simply add it there as it's the last parameter in the list.
We need to add it in vector_to_regular_points.json because it's not the last parameter and there are languages like JS where the order or params matters.

m-mohr: >  Just omitting it now and having meter by default later will also be future proof I think

No, not all languages allow Parameters by name like Python.

soxofaan: ok I understand.

If you want to keep distance and unit better together, isn't it more logical to use group these aspects in a single object, like we currently already do in apply_neightborhood? This would allow to omit it now and add support for a "distance object" later in a future-proof way.

soxofaan: >  (instead of creating wrapper objects that - right now - would look very meaningless).

well, I meant to just define `distance` as a simple number for now (implicit unit meters)  and later, when we want more units, we also allow `distance` to be an object, but keep supporting simple numbers (in implicit meters)

m-mohr: Hmm... that makes the schema more complex though. Also fine for me, I don't have a strong preference.

m-mohr: @clausmichele @LukeWeidenwalker @soxofaan and all others:

The question basically is, how do we encode values with units when passed from users to processes.

What's the preferred way generally for openEO?
1. One parameter (e.g. `distance`) encoded as object (e.g. `{"value": 123, "unit": "m"}`)
2. Two parameters (e.g. `distance` and `unit`) endcoded as scalars (e.g. `123` and `"m"`)

mkadunc (approved):

proposals/vector_to_regular_points.json
LukeWeidenwalker: Slight preference for `One parameter (e.g. distance) encoded as object (e.g. {"value": 123, "unit": "m"})`

LukeWeidenwalker (approved):

proposals/vector_to_regular_points.json
clausmichele: I'm also for just one parameter as an object including value and unit.

proposals/vector_buffer.json
```diff
@@ -22,7 +22,7 @@
         },
         {
             "name": "distance",
-            "description": "The distance of the buffer in the unit of the spatial reference system. A positive distance expands the geometries and results in outward buffering (dilation) while a negative distance shrinks the geometries and results in inward buffering (erosion).",
+            "description": "The distance of the buffer in meters. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert to a suitable spatial reference system.\n\nA positive distance expands the geometries and results in outward buffering (dilation) while a negative distance shrinks the geometries and results in inward buffering (erosion).",
```
dthiex, line 25:
```suggestion
            "description": "The distance of the buffer in meters. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert your vector data cube to a suitable spatial reference system.\n\nA positive distance expands the geometries and results in outward buffering (dilation) while a negative distance shrinks the geometries and results in inward buffering (erosion).",
```

proposals/vector_to_regular_points.json
```diff
@@ -23,12 +23,22 @@
...
+            "description": "Defines the minimum distance in meters that is required between two samples generated *inside* a single geometry. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert to a suitable spatial reference system.\n\n- For **polygons**, the distance defines the cell sizes of a regular grid that starts at the upper-left bound of each polygon. The centroid of each cell is then a sample point. If the centroid is not enclosed in the polygon, no point is sampled. If no point can be sampled for the geometry at all, the first coordinate of the geometry is returned as point.\n- For **lines** (line strings), the sampling starts with a point at the first coordinate of the line and then walks along the line and samples a new point each time the distance to the previous point has been reached again.\n- For **points**, the point is returned as given.",
```
dthiex, line 26:
```suggestion
            "description": "Defines the minimum distance in meters that is required between two samples generated *inside* a single geometry. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert your vector data cube to a suitable spatial reference system.\n\n- For **polygons**, the distance defines the cell sizes of a regular grid that starts at the upper-left bound of each polygon. The centroid of each cell is then a sample point. If the centroid is not enclosed in the polygon, no point is sampled. If no point can be sampled for the geometry at all, the first coordinate of the geometry is returned as point.\n- For **lines** (line strings), the sampling starts with a point at the first coordinate of the line and then walks along the line and samples a new point each time the distance to the previous point has been reached again.\n- For **points**, the point is returned as given.",
```

dthiex (changes requested):

proposals/vector_to_regular_points.json
m-mohr:
```suggestion
            "description": "Defines the minimum distance in meters that is required between two samples generated *inside* a single geometry. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert the geometries to a suitable spatial reference system.\n\n- For **polygons**, the distance defines the cell sizes of a regular grid that starts at the upper-left bound of each polygon. The centroid of each cell is then a sample point. If the centroid is not enclosed in the polygon, no point is sampled. If no point can be sampled for the geometry at all, the first coordinate of the geometry is returned as point.\n- For **lines** (line strings), the sampling starts with a point at the first coordinate of the line and then walks along the line and samples a new point each time the distance to the previous point has been reached again.\n- For **points**, the point is returned as given.",
```

proposals/vector_buffer.json
m-mohr:
```suggestion
            "description": "The distance of the buffer in meters. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert the geometries to a suitable spatial reference system.\n\nA positive distance expands the geometries and results in outward buffering (dilation) while a negative distance shrinks the geometries and results in inward buffering (erosion).",
```

proposals/vector_buffer.json
```diff
@@ -22,7 +22,7 @@
         },
         {
             "name": "distance",
-            "description": "The distance of the buffer in the unit of the spatial reference system. A positive distance expands the geometries and results in outward buffering (dilation) while a negative distance shrinks the geometries and results in inward buffering (erosion).",
+            "description": "The distance of the buffer in meters. If the unit of the spatial reference system is not meters, a `UnitMismatch` error is thrown. Use ``vector_reproject()`` to convert the geometries to a suitable spatial reference system.\n\nA positive distance expands the geometries and results in outward buffering (dilation) while a negative distance shrinks the geometries and results in inward buffering (erosion).",
```
neteler, line 25: Too late since it is already merged, but the wording could be:

_A positive distance expands the geometries, resulting in outward buffering (dilation), while a negative distance shrinks the geometries, resulting in inward buffering (erosion)._

m-mohr: I can still integrate that later, thanks :-)

Issue #416, opened 2023-03-10
Move ML processes to 2.1.0


m-mohr:
As they are very unstable and a lot of discussions are still ongoing, I'd propose to remove the ML-related processes from the 2.0.0 release branch and keep them in a separate branch for 2.1.0.

soxofaan:
should we also remove `fit_curve` (which is also some kind of ML process)?
(also see #425)

PR #441, opened 2023-05-15
Machine Learning for openEO

m-mohr:
- Readds the basic ML processes #416 (supersedes #418)
- Generalizes the predict functionalities #368 (supersedes #396)
- ...

### DIFF
--- a/proposals/save_ml_model.json
+++ b/proposals/save_ml_model.json
@@ -1,0 +1,44 @@
+{
+    "id": "save_ml_model",
+    "summary": "Save a ML model",
+    "description": "Saves a machine learning model as part of a batch job.\n\nThe model will be accompanied by a separate STAC Item that implements the [ml-model extension](https://github.com/stac-extensions/ml-model).",
+    "categories": [
+        "machine learning",
+        "import"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "The data to store as a machine learning model.",
+            "schema": {
+                "type": "object",
+                "subtype": "ml-model"
+            }
+        },
+        {
+            "name": "options",
+            "description": "Additional parameters to create the file(s).",
+            "schema": {
+                "type": "object",
+                "additionalParameters": false
+            },
+            "default": {},
+            "optional": true
+        }
+    ],
+    "returns": {
+        "description": "Returns `false` if the process failed to store the model, `true` otherwise.",
+        "schema": {
+            "type": "boolean"
+        }
+    },
+    "links": [
+        {
+            "href": "https://github.com/stac-extensions/ml-model",
+            "title": "STAC ml-model extension",
+            "type": "text/html",
+            "rel": "about"
+        }
+    ]
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `filter_vector`
     - `flatten_dimensions`
     - `load_geojson`
+    - `load_ml_model`
     - `load_url`
+    - `predict_ml_model`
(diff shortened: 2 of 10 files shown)


m-mohr:
Some things from the ML meeting:

image

image

Regularization may consist of (and is mapped to openEO processes):
- resample resolution -> resample_spatial
- temporal aggregation -> aggregate_temporal_period
- filter extent -> filter_bbox/spatial
- cloud removal -> cloud_detection or masking based on cloud bands

-> combine these to a new openEO process with some arguments that are commonly used with reasonable defaults

m-mohr:
datacube = load_collection("s2", temporal_extent = ..., spatial_extent = ..., bands = ...)
model = load_ml_model("my-model-id")

# variant 1
function fn(data) {
	let values = ml_predict(data, model)
	return array_element(values, 0)
}
datacube2 = reduce_dimension(datacube, dimension = "bands", reducer = fn)

function fn2(data) {
	return ml_predict(data, model)
}
datacube3 = apply_dimension(datacube, dimension = "bands", target_dimension = "predictions", process = fn2)

# variant 2
datacube2 = ml_predict(datacube, dimension = "bands", model = model)
datacube2 = drop_dimension(datacube, name = "predictions")

datacube3 = ml_predict(datacube, dimension = "bands", model = model)

PondiB:
@m-mohr , is there a reason why *_fit_* is used instead of *_train_*?
I will work on data cube regularization and do a different PR.

m-mohr:
Just to align with fit_cuve, I guess. Train is also fine...

PondiB:
> Just to align with fit_cuve, I guess. Train is also fine...

Cool

PondiB (approved):
Looks good. Approving this now. We can continuously improve during implementations if need be.

PondiB:
@m-mohr is renaming these two to follow with a prefix of "ml" a possible alternative?
*load_ml_model()* >  *ml_load_model()*
*save_ml_model()* > *ml_save_model()*

m-mohr:
Why? The current proposal follows the load_* and save_result schema.

PondiB:
> Why? The current proposal follows the load_* and save_result schema.

Cool makes sense to follow the previous schema. My initial thoughts came from the perception that it would be good for a general user if most of the ml operations start with that prefix i.e. "ml_".

m-mohr:
The STAC ML Model extension may get deprecated in favor of https://github.com/crim-ca/dlm-extension
@PondiB I think it would be great to get in touch with the folks so that we can influence that it also works for openEO.

proposals/save_ml_model.json
```diff
@@ -0,0 +1,44 @@
+{
+    "id": "save_ml_model",
+    "summary": "Save a ML model",
+    "description": "Saves a machine learning model as part of a batch job.\n\nThe model will be accompanied by a separate STAC Item that implements the [ml-model extension](https://github.com/stac-extensions/ml-model).",
```
soxofaan, line 4: > The model will be accompanied by a separate STAC Item that

What does "accompanied" practically mean? Should there be an additional job result asset? Or should this be an job result link item?

The reason I'm asking is that we want to streamline the detection of the model's URL at the client side.

e.g. see https://github.com/Open-EO/openeo-python-client/pull/576 we we currently have a highly implementation-specific hack
```
ml_model_metadata_url = [
    link 
    for link in links if 'ml_model_metadata.json' in link['href']
][0]['href']
```

m-mohr: Good question, I guess we should clarify that.

On the other hand, please note that this PR is implicitly outdated as the ML Model extension in STAC is likely going to be replaced by another extension. So this generally needs more work (which I have no plans to do anytime soon).

m-mohr: https://github.com/crim-ca/mlm-extension

PR #456, opened 2023-08-23
Adding imaginary unit

clausmichele:
Adding a process expressing the imaginary unit. First step towards complex numbers in openEO.

PR #458, opened 2023-08-23
Adding imaginary unit - 2

clausmichele:
Same content as https://github.com/Open-EO/openeo-processes/pull/456 but with the correct base branch.


clausmichele:
Sorry, apparently it doesn't include the fixed version of the json. I'll try again.

m-mohr:
You could've just pushed to the branch again :-D

Issue #455, opened 2023-08-23
Complex numbers in openEO


clausmichele:
At Eurac we are working on the improvements of the SAR2Cube project, where we are using openEO to process Sentinel-1 SLC complex data. @aljacob 

I create this issue to collect ideas and feedbacks about introducing the possibility to handle complex numbers in openEO.

In my opinion, the first step should be introducing a new process in Math->Constants for the imaginary unit `j` (or `i`, depending on how we want to name it).
If the complex numbers will be compatible with the already existing implementation depends on the back-ends, but for the Python based ones it will be doable. For sure, a process like `absolute` should be firstly updated with a mention to the support of complex numbers if we proceed.

clausmichele:
https://github.com/Open-EO/openeo-processes/pull/456

clausmichele:
@soxofaan the possibilities will include handling datacubes with complex data and also creation of new datacubes based on real + imag bands (imagine a reduce_dimension over bands where we do real + i*imag).

soxofaan:
>  the first step should be introducing a new process in Math->Constants for the imaginary unit

I'm not sure this is the necessary first step. It implies that complex numbers are a thing already, and as far as I known a JSONSchema `number` is just a real number. So if you want to define a "process" `i`, you first have to define a schema for complex numbers so you can use that for process `i`. After that you also have to add it to the specification of all processes where you want to handle complex numbers (e.g. all math processes at least I guess, array manipulation processes, ... ). Not only induces this a lot of changes in a lot of processes, I'm also afraid this will cause a schism in process specs between back-ends that do and don't want to or can support complex numbers.


An alternative solution is introducing a new type of dimension, e.g. "complex", which acts like a "band" dimension, but has just 2 pre-defined "bands": "real" and "imag". This approach does not require to rewrite a lot of (math related) processes, and you can just do complex number arithmetic with the existing openEO API and processes.

m-mohr:
I like the thoights of @soxofaan. We may even want to create a new dimension type for this instead of using "other" so that data cubes with complex numbers can be "enforced" in schemas, similar to the recent addition of the geometry/vector dimension type.

clausmichele:
Ok, so `save_result` could check if this dimension exists and act accordingly, merging real and imaginary parts in a single output.

PR #459, opened 2023-08-23
Adding imaginary unit

clausmichele:
This should be the good one @m-mohr. Sorry for the chaos with the previous PRs, I did a mess.

### DIFF
--- a/proposals/i.json
+++ b/proposals/i.json
@@ -1,0 +1,23 @@
+{
+    "id": "i",
+    "summary": "Imaginary unit (i)",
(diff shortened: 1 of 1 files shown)


m-mohr (approved):

soxofaan:
just curious: what is this for, and how are complex numbers expected to be handled in data cubes? E.g. do you get a data cube with some band-like dimension, having a real and imaginary bands?

soxofaan:
hmm I wonder if this should have been merged,
I don't think we were already sure that adding the imaginary unit was a good first step for #455

PondiB:
@soxofaan, Well noted, my bad. It's reverted on #464 . Could you approve the revert?

PR #470, opened 2023-10-12
`filter_spatial`: Clarify masking

### LINKED ISSUE #469: filter_spatial: clarify behaviour
The filter_spatial process definition https://processes.openeo.org/#filter_spatial is not clear, since it mentions keeping only the pixels intersected by the geometry and later, under the geometries parameter, mentions that no masking is applied.

We need to decide what's the expected output.
Is it only a filter_bbox process that automatically extracts the AOI from the bounds of the geometries?
Is it the same as mask_polygon?

@jdries said that on their side masking is applied, code here: https://github.com/Open-EO/openeo-geopyspark-driver/commit/4e3ef048550bf37b9ed43152d6b8733363cd0f63

We are waiting for a clarification to proceed with this PR: https://github.com/Open-EO/openeo-processes-dask/pull/170

Tagging some other people for opinions:
@ValentinaHutter @dthiex @soxofaan @SerRichard

m-mohr:
Fixes #469


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased / Draft
 
+### Fixed
+
+- `filter_spatial`: Clarified that a masking get applied for the given geometries. [#469](https://github.com/Open-EO/openeo-processes/issues/469)
```
soxofaan, line 11:
```suggestion
- `filter_spatial`: Clarified that masking is applied using the given geometries. [#469](https://github.com/Open-EO/openeo-processes/issues/469)
```

filter_spatial.json
```diff
@@ -26,7 +26,7 @@
         },
         {
             "name": "geometries",
-            "description": "One or more geometries used for filtering, given as GeoJSON or vector data cube. If multiple geometries are provided, the union of them is used. Empty geometries are ignored.\n\nLimits the data cube to the bounding box of the given geometries. No implicit masking gets applied. To mask the pixels of the data cube use ``mask_polygon()``.",
+            "description": "One or more geometries used for filtering, given as GeoJSON or vector data cube.",
```
soxofaan, line 29:
```suggestion
            "description": "One or more geometries used for spatial filtering and masking, given as GeoJSON or vector data cube.",
```

filter_spatial.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Limits the raster data cube over the spatial dimensions to the specified geometries.\n\n- For **polygons**, the filter retains a pixel in the data cube if the point at the pixel center intersects with at least one of the polygons (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nMore specifically, pixels outside of the bounding box of the given geometries will not be available after filtering. All pixels inside the bounding box that are not retained will be set to `null` (no data).\n\n Alternatively, use ``filter_bbox()`` to filter with a bounding box or ``filter_vector()`` to filter a vector data cube based on geometries.",
```
soxofaan, line 4:
```suggestion
    "description": "Limits the raster data cube over the spatial dimensions to the specified geometries.\n\n- For **polygons**, the filter retains a pixel in the data cube if the point at the pixel center intersects with at least one of the polygons (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nMore specifically, pixels outside of the bounding box of the given geometries will not be available after filtering. All pixels inside the bounding box that are not retained will be set to `null` (no data).\n\n Alternatively, use ``filter_bbox()`` to filter with a bounding box or ``filter_vector()`` to filter a vector data cube based on geometries. Use ``mask_polygon()`` to just mask without changing the spatial extent of your data cube.",
```

soxofaan (commented):
some minor suggestions

filter_spatial.json
m-mohr:
```suggestion
    "description": "Limits the raster data cube over the spatial dimensions to the specified geometries.\n\n- For **polygons**, the filter retains a pixel in the data cube if the point at the pixel center intersects with at least one of the polygons (as defined in the Simple Features standard by the OGC).\n- For **points**, the process considers the closest pixel center.\n- For **lines** (line strings), the process considers all the pixels whose centers are closest to at least one point on the line.\n\nMore specifically, pixels outside of the bounding box of the given geometries will not be available after filtering. All pixels inside the bounding box that are not retained will be set to `null` (no data).\n\n Alternatively, use ``filter_bbox()`` to filter with a bounding box or ``filter_vector()`` to filter a vector data cube based on geometries. Use ``mask_polygon()`` to mask without changing the spatial extent of your data cube.",
```

soxofaan:
> How is filter_bbox handled if the CRS of the bbox is different from the data cube CRS? Do you reproject the bbox to the data cube CRS first to align? Otherwise, you may also get an outside area that you may need to null.

Yes we reproject (e.g. typically from lon-lat to  UTM zone) and I should double check the implementation details, but I think the intent is to use the tightest bbox in target CRS (e.g. UTM) that covers the specified bbox (e.g. in lon-lat). 
So indeed, you might get more pixels outside of the original bbox, but I don't think that is a problem in most use cases. And if it is a problem you should use explicit masking (e.g. with mask_polygon or filter_spatial) anyway.

clausmichele:
We also reproject the bbox first to match the data projection, here is the implementation: https://github.com/Open-EO/openeo-processes-dask/blob/597d1005a91bb36823d191cb14507f5a1ee716f1/openeo_processes_dask/process_implementations/cubes/_filter.py#L105

m-mohr:
Thanks, I've added a clarifying remark to filter_bbox in this PR.

clausmichele (approved):

filter_bbox.json
```diff
@@ -39,7 +39,7 @@
         },
         {
             "name": "extent",
-            "description": "A bounding box, which may include a vertical axis (see `base` and `height`).",
+            "description": "A bounding box, which may include a vertical axis (see `base` and `height`).\n\nIf the bounding box is not provided in the coordinate reference system (CRS) of the data cube, the bounding box is reprojected to the CRS of the spatial data cube dimensions.",
```
soxofaan, line 42: This clarification  also applies to `spatial_extent` in `load_collection`, but it's probably a bit messier to do there as multiple data types are allowed

soxofaan (approved):
Fine for me as-is

filter_bbox.json
m-mohr: Added.

Issue #472, opened 2023-10-24
clip: Document edge cases


m-mohr:
**Process ID:** clip

**Describe the issue:**

During the test suite creation, I identified that the clip process doesn't mention what happens if min > max for example.

**Proposed solution:**
We should check the tests and document what's missing, e.g. throw an error if min > max? Check what similar processes do, e.g. between.

**Additional context:**
Tests in #468

m-mohr:
What do backends do? @soxofaan @guss84 @clausmichele @LukeWeidenwalker

clausmichele:
In the xarray/dask implementation, if max < min, no error is raised. The base function is provided by numpy, and they explicitly mention this [in the docs](https://numpy.org/doc/stable/reference/generated/numpy.clip.html):
_No check is performed to ensure a_min < a_max._

Anyway, we can agree to raise a predefined error if this happens.

soxofaan:
I don't think we have a check either.
Specifying to throw an error makes most sense to me too

m-mohr:
PR: #477

PR #468, opened 2023-10-12
Tests for the test suite

m-mohr:
See the [README](https://github.com/Open-EO/openeo-processes/blob/add-tests/tests/README.md) for details.

This PR includes the changes from PR #490.

PR #473, opened 2023-10-24
divide, ln, log, mod: Clarified behavior for 0 input / infinity results

m-mohr:
We mention NaN here the first time, I think.

Related: #468, #472, #474

### DIFF
--- a/divide.json
+++ b/divide.json
@@ -1,7 +1,7 @@
 {
     "id": "divide",
     "summary": "Division of two numbers",
-    "description": "Divides argument `x` by the argument `y` (*`x / y`*) and returns the computed result.\n\nNo-data values are taken into account so that `null` is returned if any element is such a value.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. Therefore, a division by zero results in ±infinity if the processing environment supports it. Otherwise, a `DivisionByZero` exception must the thrown.",
+    "description": "Divides argument `x` by the argument `y` (*`x / y`*) and returns the computed result.\n\nNo-data values are taken into account so that `null` is returned if any element is such a value.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. A division by zero results in:\n\n- +infinity for `x` > 0,\n- -infinity for `x` < 0,\n- `NaN` for `x` = 0,\n- or otherwise, throws a `DivisionByZero` exception if the other options are not supported by the processing environment.",
     "categories": [
         "math"
     ],
--- a/log.json
+++ b/log.json
@@ -1,7 +1,7 @@
 {
     "id": "log",
     "summary": "Logarithm to a base",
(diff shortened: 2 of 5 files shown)


divide.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "divide",
     "summary": "Division of two numbers",
-    "description": "Divides argument `x` by the argument `y` (*`x / y`*) and returns the computed result.\n\nNo-data values are taken into account so that `null` is returned if any element is such a value.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. Therefore, a division by zero results in ±infinity if the processing environment supports it. Otherwise, a `DivisionByZero` exception must the thrown.",
+    "description": "Divides argument `x` by the argument `y` (*`x / y`*) and returns the computed result.\n\nNo-data values are taken into account so that `null` is returned if any element is such a value.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. A division by zero results in:\n\n- +infinity for `x` > 0,\n- -infinity for `x` < 0,\n- `NaN` for `x` = 0,\n- or otherwise, throws a `DivisionByZero` exception if the other options are not supported by the processing environment.",
```
soxofaan, line 4: > +infinity for `x` > 0,\n- -infinity for `x` < 0,\n-

I'm not sure this is correct (I'm also not sure how tightened down the spec should be here): 
the sign of infinity also depends on which side you're looking from, e.g. `1/y` at `y=0` is +infinity  if you "approach" that limit it from positive y values, but -infinity if you approach it from negative values.

I think there is a bit of discrepancy between rigid mathematical definitions and what is practically feasible in a numerical/computer implementation, so I would not over-specify this. I'm kind of fine with what we had originally, or something like
> a division by zero results in ±infinity or NaN, depending on the capabilities of the processing environment.

soxofaan: I also wonder if we should recommend against (or forbid)  throwing a DivisionByZero error. If you're processing thousands/millions of pixels you probably prefer a NaN in some output pixels over your whole job failing.

m-mohr: > I also wonder if we should recommend against (or forbid) throwing a DivisionByZero error. If you're processing thousands/millions of pixels you probably prefer a NaN in some output pixels over your whole job failing.

Well, the error only occurs if your environment doesn't support Nan/Infinity. It's difficult to return NaN if Nan is unsupported ;-)

soxofaan: > It's difficult to return NaN if NaN is unsupported

sure, but we could recommend against throwing DivisionByZero if the environment *does* support NaN

log.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "log",
     "summary": "Logarithm to a base",
-    "description": "Logarithm to the base `base` of the number `x` is defined to be the inverse function of taking b to the power of x.\n\nThe no-data value `null` is passed through and therefore gets propagated if any of the arguments is `null`.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. Therefore, `log(0, 2)` results in ±infinity if the processing environment supports it or otherwise an exception is thrown.",
+    "description": "Logarithm to the base `base` of the number `x` is defined to be the inverse function of taking b to the power of x.\n\nThe no-data value `null` is passed through and therefore gets propagated if any of the arguments is `null`.\n\nThe computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. Therefore, having `x` set to `0` with any base results in -infinity if the processing environment supports it. Otherwise, an exception is thrown.",
```
soxofaan, line 4: We should probably  specify here (and for `ln`) that NaN is returned when x is negative too

m-mohr:
I'll actually close here and merge this into #476 because the changes are overlapping.

log.json
m-mohr: Should be part of #476 now.

PR #476, opened 2023-10-24
Math functions: Clarified value ranges and NaN

m-mohr:
Related to #468


arcsin.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "arcsin",
     "summary": "Inverse sine",
-    "description": "Computes the arc sine of `x`. The arc sine is the inverse function of the sine so that *`arcsin(sin(x)) = x`*.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated.",
```
soxofaan, line 4: add return min (-pi/2) and max (pi/2) like with arccos?

arccos.json
```diff
@@ -23,7 +23,8 @@
             "type": [
                 "number",
                 "null"
-            ]
+            ],
+            "minimum": 0
```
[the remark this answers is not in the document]
m-mohr: Yes, but I wasn't sure how to expose Pi properly in JSON.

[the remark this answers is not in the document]
soxofaan: just as floating point `3.1415...` will be good enough for most purposes I guess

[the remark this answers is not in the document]
m-mohr: Yeah, for most. But now that I'm writing the tests, these edge cases get in the way of that. 

If I use 3.1415, a test with 3.14159 would fail. So I'd need to set it to 3.1416, which is a bit weird. The precision is an example and could apply on any level of precition. Thus I left it open.

arcsin.json
m-mohr: I added the range to the text for all functions. I hope that makes the ranges clearer without fiddling to much with specific decimal numbers.

arccos.json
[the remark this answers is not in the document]
m-mohr: I added the range to the text for all functions. I hope that makes the ranges clearer without fiddling to much with specific decimal numbers.

clausmichele (approved):
Some functions would allow extended ranges once we will support complex numbers. For the time being I think it's fine.

arcosh.json
```diff
@@ -1,29 +1,31 @@
...
                     "null"
                 ]
-            }
+            },
+            "minimum": 1
         }
     ],
     "returns": {
-        "description": "The computed angle in radians.",
+        "description": "The computed angle in radians in the range *[0, +∞)*.",
```
soxofaan, line 22:
```suggestion
        "description": "The hyperbolic angle in the range *[0, +∞)*.",
```

arcosh.json
```diff
@@ -1,29 +1,31 @@
 {
     "id": "arcosh",
     "summary": "Inverse hyperbolic cosine",
-    "description": "Computes the inverse hyperbolic cosine of `x`. It is the inverse function of the hyperbolic cosine so that *`arcosh(cosh(x)) = x`*.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated.",
+    "description": "Computes the inverse hyperbolic cosine of `x`. It is the inverse function of the hyperbolic cosine so that *`arcosh(cosh(x)) = x`*.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated. `NaN` is returned for values outside of the allowed range.",
```
soxofaan, line 4: > Works on radians only.

I have never seen the explicit clarification of "radians" in context of hyperbolic trig functions. I don't think there is an alternative definition like you have radians vs degrees with regular trig functions. It looks confusing to me to read that "Works on radians only." and made me double check I wasn't reading about standard cosine function

artanh.json
```diff
@@ -1,19 +1,21 @@
 {
     "id": "artanh",
     "summary": "Inverse hyperbolic tangent",
-    "description": "Computes the inverse hyperbolic tangent of `x`. It is the inverse function of the hyperbolic tangent so that *`artanh(tanh(x)) = x`*.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated.",
+    "description": "Computes the inverse hyperbolic tangent of `x`. It is the inverse function of the hyperbolic tangent so that *`artanh(tanh(x)) = x`*.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated.  `NaN` is returned for values outside of the allowed range.",
```
soxofaan, line 4: like in #473 you could also specify to get -/+ infinity at -/+1

cosh.json
```diff
@@ -18,12 +18,13 @@
         }
     ],
     "returns": {
-        "description": "The computed hyperbolic cosine of `x`.",
+        "description": "The computed hyperbolic cosine in the range *[1, +∞)*.",
```
soxofaan, line 21: (can't comment on lines above)

as mentioned earlier, I would drop "works on radians only" and describe the "x" parameter as "Hyperbolic angle"

tan.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "tan",
     "summary": "Tangent",
-    "description": "Computes the tangent of `x`. The tangent is defined to be the sine of x divided by the cosine of x.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated.",
+    "description": "Computes the tangent of `x`. The tangent is defined to be the sine of x divided by the cosine of x.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated. `NaN` is returned for values where the tangent function is not defined.",
```
soxofaan, line 4: why not infinity like in #473 as there is a division by zero there?

soxofaan (commented):
some more notes

arcosh.json
m-mohr: Done.

arcosh.json
m-mohr: I guess that was copy pasted back in the days. Removed.

cosh.json
m-mohr: Done, I think...

tan.json
m-mohr: I tried to get infinity in practice (e.g. `Math.tan(Math.PI/2))` in JS and `(math.tan(math.pi / 2)` in Python) but due to floating point inaccuracies it seems mostly impossible to actually get a infinity result. So I left it out, but I guess it would be more accurate to add it...

> The computations follow [IEEE Standard 754](https://ieeexplore.ieee.org/document/8766229) whenever the processing environment supports it. Therefore, *`tan(pi()/2 + multipliy(pi(), n))`* with `n` being any integer results in ±infinity if the processing environment supports it. -infinity for negative values passed to `tan`, +infinity otherwise. If the processing environment does not supports it, an exception is thrown.

Quite lengthy though...

artanh.json
m-mohr: Done.

tan.json
soxofaan: > due to floating point inaccuracies it seems mostly impossible to actually get a infinity result.

but then you can't also produce nan as it is currently defined in the description

> `NaN` is returned for values where the tangent function is not defined.


As the description defines "The tangent is defined to be the sine of x divided by the cosine of x", you could get away with it by saying "Infinity is returned for values where the sine of x is zero"

m-mohr: I just added the lengthy one from above, that should work.

m-mohr:
Merged #473 into this PR due to the large overlap of changes.

PR #478, opened 2023-10-26
`array_append`: Add number type for labels

m-mohr:
`array_append`: Added `number` type for labels to be consistent with other processes. Default to numerical index instead of string.

### DIFF
--- a/array_append.json
+++ b/array_append.json
@@ -25,15 +25,20 @@
         },
         {
             "name": "label",
-            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as string is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
+            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as number is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
             "optional": true,
             "default": null,
-            "schema": {
-                "type": [
-                    "string",
-                    "null"
-                ]
-            }
+            "schema": [
+                {
+                    "type": "number"
+                },
+                {
+                    "type": "string"
+                },
+                {
+                    "type": "null"
+                }
+            ]
         }
     ],
     "returns": {
--- a/array_append.json
+++ b/array_append.json
@@ -25,7 +25,7 @@
         },
         {
             "name": "label",
-            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as number is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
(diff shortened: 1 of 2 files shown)


array_append.json
```diff
@@ -25,15 +25,20 @@
         },
         {
             "name": "label",
-            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as string is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
+            "description": "If the given array is a labeled array, a new label for the new value should be given. If not given or `null`, the array index as number is used as the label. If in any case the label exists, a `LabelExists` exception is thrown.",
```
soxofaan, line 28: > If not given or `null`, the array index as number is used as the label

not sure what "array index" means in this context. Does it mean something like original length? e.g. `array_append([10, 11], 12)`: length is 2 -> new label = 2

soxofaan (approved):
minor remark, but already going to approve

array_append.json
m-mohr: Yes, indeed. It's the natural next array index pretty much ;-)

soxofaan: makes me wonder: what happens with a non-labeled array and the given label is "out of range"

e.g. `array_append([10, 11], value=20, label=100)`

does that result in `[10, 11, NaN, NaN, ......, NaN, 20]`? or is that invalid

and what if the label is negative?

m-mohr: Labels can't be out of range. They behave as if they would be a string, same applies for negative.

But if you pass a non-labeled array and want to add an element with a label, I'm not sure whether we should convert or throw an error?

m-mohr: (our labeled arrays pretty much work like PHP arrays 😅)

soxofaan: > But if you pass a non-labeled array and want to add an element with a label, I'm not sure whether we should convert or throw an error?

what do you mean with convert? first convert the non-labeled array to labeled array and then use the label as-is?

> (our labeled arrays pretty much work like PHP arrays 😅)

Back in the day when I did PHP development, the frankenstein character of PHP arrays was one of the things I hated most FYI.

m-mohr: > what do you mean with convert? first convert the non-labeled array to labeled array and then use the label as-is?

The complexity probably means we should throw an error. 

> he frankenstein character of PHP arrays was one of the things I hated most

I love them and still sometimes miss them lol

clausmichele (approved):

array_append.json
soxofaan: > But if you pass a non-labeled array and want to add an element with a label, I'm not sure whether we should convert or throw an error?

I'd say throw an error. Implicit conversion of non-labeled to labeled from an append operation doesn't feel right to me

PR #483, opened 2023-12-05
ML Support Vector Classification

PondiB:
Continuation of #441 , Support Vector Classification


PondiB:
@m-mohr, my understanding is that this test fail is probably a false fail because the "save_ml_model" does not exist yet in the draft branch.

PondiB:
Just confirmed that's it. I will cross reference them later then as they need to be in my working repo too.

proposals/ml_fit_class_support_vector.json
```diff
@@ -0,0 +1,153 @@
...
+                        {
+                            "type": "other"
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "target",
+            "description": "The training sites for the classification model as a vector data cube. This is associated with the target variable for the support vector classification model. The geometry has to be associated with a class label.",
```
soxofaan, line 42: I have a hard time understanding what this description is saying about the `target` vector cube

PondiB: Redefined.

proposals/ml_fit_class_support_vector.json
```diff
@@ -0,0 +1,153 @@
...
+                        {
+                            "type": "other"
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "target",
+            "description": "Labeled information for training the support vector classification model. Each entry corresponds to a set of features (predictors).",
```
soxofaan, line 42: I still find this unclear. "Labeled information" can be interpreted as predictors plus a label, but I don't think that is intended here.
It's also risky  to use the term "features" (ML interpretation: predictor values) in another interpretation than used in the description of the predictors argument (EO interpretation: geometry). 

"Each entry corresponds to a set of .." is also a bit too loose as constraint, I think the point should be that the the `predictors` vector cube and `target` cube have a common geometry dimension, or at least that the backend is able to connect "predictor rows" from `predictors` with labels from `target`.

m-mohr:
@PondiB I think it would make sense to make PRs against the ml branch because otherwise all changes from the ML branch will also appear in this PR. This leads to confusion as you can see with @soxofaan's comments. Please rebase your changes against the ML branch if necessary and set the base branch of the PR to ml.

m-mohr (changes requested):
@PondiB I think it would make sense to make PRs against the ml branch because otherwise all changes from the ML branch will also appear in this PR. This leads to confusion as you can see with @soxofaan's comments. Please rebase your changes against the ML branch if necessary and set the base branch of the PR to ml.

PondiB:
> @PondiB I think it would make sense to make PRs against the ml branch because otherwise all changes from the ML branch will also appear in this PR. This leads to confusion. Please rebase your changes against the ML branch if necessary and set the base branch of the PR to ml.

Sure, noted

PondiB:
@m-mohr , for some reason when one runs "npm test" locally, it gets historical commits not associated with the current branch. This branch for example has got 27 changes now. Did I miss something?

PondiB:
Closing this in favor of #492

Issue #467, opened 2023-10-05
export_to_workspace


m-mohr:
**Proposed Process ID:** export_to_workspace

## Context
For the new workspace API we want a way to have a process that explicitly allows to store a result in a dedicated workspace instead of in "the normal user space"

We thought about having it as part of save_result, but then you either mix the storage location with the file formats or you get a number of parameters that depend on each other in save_result, both not ideal.

Backends may not do the intermediate step to store in the normal user space and directly store in the workspace although both processes are present.

The process usually only makes sense in batch jobs, I think.

**NOTE:** This process may have implications on save_result!

## Summary
Exports a result to a registered workspace

## Description
... 

### `data` (or so?)

One issue is that save_result returns a boolean, so we need to look into that. Maybe it should return something else that we can feed into export_to_workspace as data? Maybe a list of files? Or just a generic "result" subtype?

### `name` 

Name of the registered workspace.

Type: string
Subtype: workspace-name

**Optional:** no

### `folder` (or prefix?)

Stores the batch job with the given prefix / to the given folder in the workspace.

Default to `null`, which uses a unique identifier as prefix (e.g. batch job id).
An empty string stores them in the "root folder".

*Q: Should the process by default override files or not?*

Type: string or null

**Optional:** yes, default: null

## Return Value

boolean?

## Categories (optional)
* export

## Links to additional resources (optional)
* Link to workspace API

## Details about exceptions mentioned in the description (optional)
* FileExists

m-mohr:
cc @Open-EO/team-eodc

jdries:
Should we have most of the discussion in this issue: https://github.com/Open-EO/openeo-api/issues/376
Because there we already collected a lot of thoughts and requirements. I'm also not really clear yet on what the new workspace API looks like.

m-mohr:
User-generated Collections and user workspaces are not necessarily the same, so they deserve separate issues, I think.
I tried to collect relevant issues and links in https://github.com/openEOPlatform/architecture-docs/issues/215#issuecomment-1748497731

PR #485, opened 2023-12-08
STAC, Collection and Workspace handling in openEO

m-mohr:
### Related issues

- https://github.com/Open-EO/openeo-api/issues/376
- https://github.com/Open-EO/openeo-api/issues/135
- https://github.com/Open-EO/openeo-processes/issues/467

### Changelog

- Added `export_collection`
- Added `export_workspace`
- Added `stac_update`
- `save_results`: Returns the STAC resource instead of boolean `true`

### DIFF
--- a/proposals/export_collection.json
+++ b/proposals/export_collection.json
@@ -1,0 +1,45 @@
+{
+    "id": "export_collection",
+    "summary": "Export data to an openEO collection",
+    "description": "Exports the given processing results made available through a STAC resource (e.g., a STAC Collection or Item) to the given openEO collection. The STAC resource itself is exported with all STAC resources and assets underneath.",
+    "categories": [
+        "export",
+        "stac"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "The data to export to the openEO collection.",
+            "schema": {
+                "type": "object",
+                "subtype": "stac"
+            }
+        },
+        {
+            "name": "collection",
+            "description": "The identifier of the collection to export to. If the collection **exists**, the data will be added to the collection. If the collection **does not exist** yet, it will be created based on the given STAC metadata.",
+            "schema": [
+                {
+                    "title": "New Collection",
+                    "type": "string",
+                    "pattern": "^[A-Za-z0-9_\\-\\.~/]+$"
+                },
+                {
+                    "title": "Existing Collection",
+                    "type": "string",
+                    "subtype": "collection-id",
+                    "pattern": "^[A-Za-z0-9_\\-\\.~/]+$"
+                }
+            ]
+        }
+    ],
+    "returns": {
+        "description": "Returns the collection identifier.",
+        "schema": {
+            "type": "string",
+            "subtype": "collection-id",
+            "pattern": "^[A-Za-z0-9_\\-\\.~/]+$"
+        }
+    }
+}
--- a/proposals/stac_update.json
+++ b/proposals/stac_update.json
@@ -1,0 +1,41 @@
+{
+    "id": "stac_update",
+    "summary": "Export data to an openEO collection",
+    "description": "Exports the given processing results made available through a STAC resource (e.g., a STAC Collection or Item) to the given openEO collection. The STAC resource itself is exported with all STAC resources and assets underneath.",
+    "categories": [
+        "stac"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "The existing STAC resource.",
+            "schema": {
+                "type": "object",
+                "subtype": "stac"
+            }
+        },
+        {
+            "name": "merge",
+            "description": "A potentially incomplete STAC resource that should be merged with the existing STAC resource. It follows the [RFC 7386: JSON Merge Path](https://tools.ietf.org/html/rfc7386) specification.",
+            "schema": {
+                "type": "object"
+            }
+        }
+    ],
+    "returns": {
+        "description": "Returns the updated STAC resource.",
+        "schema": {
+            "type": "object",
+            "subtype": "stac"
+        }
+    },
+    "links": [
+        {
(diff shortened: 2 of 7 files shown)


proposals/stac_update.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id": "stac_update",
```
mkadunc, line 2: `stac_update` process has exactly the same summary and description as `export_collection` — do we need two processes at all or are they both doing the same thing?

mkadunc (approved):

proposals/stac_update.json
m-mohr: Sorry, my fault. I forgot to update the summary and description. Fixed.

jdries:
So for my use case of storing metadata of jobs in an existing STAC collection, I would use export_collection?
This collection can then optionally expose a STAC api I assume, as this is already foreseen.
By default, this would be a private collection? Do we already have a way to make such a collection public?
Sharing within a group would perhaps be via signed url?

m-mohr:
@jdries 

> So for my use case of storing metadata of jobs in an existing STAC collection, I would use export_collection?

- If you want to store it as an openEO Collection (i.e. exposed via `/collections`, usable via `load_collection`): `export_collection`
- If you just want to have a STAC Collection on cloud storage somewhere (usable via `load_stac`): `export_workspace` (if your backend exports STAC Collections for batch job results, otherwise it would be an Item)

> This collection can then optionally expose a STAC api I assume, as this is already foreseen.

export_collection would automatically expose this through the openEO API (i.e. `/collections`), which can optionally have the STAC API Items endpoints (i.e. `/collections/:id/items`).

> By default, this would be a private collection?

Yes, by default this is only accessible after authentication from the user that created the collection.

> Do we already have a way to make such a collection public?
> Sharing within a group would perhaps be via signed url?

No, I'd like to generally target sharing in a common manner across all openEO resource types. 

We pretty much have defined the behavior for making resources public (happy to adopt the public flag as implemented by VITO and signed URLs as documented in batch job results). We don't have anything more specific for users or groups. Both will need separate work items (a smaller one for public sharing, and a larger one for user/group based sharing). Unfortunately, there's currently no funding foreseen for this in any project afaik.

m-mohr:
Any further reviews? @soxofaan @jdries @GeraldIr @clausmichele @clausmichele 
Otherwise, I'd merge this soon.

meta/implementation.md
```diff
@@ -230,3 +230,24 @@ We have found some libraries that can be used for an implementation:
 - Julia: [Statistics.quantile](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.quantile!), type 7 is the default.
 - Python: [numpy](https://numpy.org/doc/stable/reference/generated/numpy.quantile.html), [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.quantile.html), [xarray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.quantile.html) - type 7 (called 'linear' for the interpolation parameter) is the default for all of them.
 - R: [quantile](https://stat.ethz.ch/R-manual/R-patched/library/stats/html/quantile.html) - type 7 is the default.
+
+## STAC support
+
+The subtype `stac` is an abstract type that refers to a STAC resource of any type (Catalog, Collection, or Item).
+It can refer to:
+- static STAC resources, e.g. hosted on cloud storage
+- "dynamic" STAC resources made available via a STAC API
```
clausmichele, line 239: Just for style alignment: should we use the quotation marks also for `static` or remove them from `dynamic`?

proposals/export_collection.json
```diff
@@ -0,0 +1,45 @@
...
+            "name": "data",
+            "description": "The data to export to the openEO collection.",
+            "schema": {
+                "type": "object",
+                "subtype": "stac"
+            }
+        },
+        {
+            "name": "collection",
+            "description": "The identifier of the collection to export to. If the collection **exists**, the data will be added to the collection. If the collection **does not exist** yet, it will be created based on the given STAC metadata.",
```
clausmichele, line 21: I guess that the already existing collection must also be a user created collection? Or not necessarily? Is it completely overwriting the existing one?

proposals/stac_modify.json
```diff
@@ -0,0 +1,41 @@
+{
+    "id": "stac_modify",
+    "summary": "Updates an existing STAC resource",
+    "description": "Modifies the given STAC resource (e.g., a STAC Collection or Item) based on the given changeset.",
```
clausmichele, line 4: I'm trying to figure out if I would need to use this process or `export_collection` in the scenario where a user created already a STAC Collection with some Items offered via an API and wants to add some new Items.

clausmichele:
> Any further reviews? @soxofaan @jdries @GeraldIr @clausmichele @clausmichele Otherwise, I'd merge this soon.

Comments added, thanks for tagging me and my twin @clausmichele for a double review

proposals/export_collection.json
m-mohr: I guess that's up to the back-end to decide?!

proposals/stac_modify.json
m-mohr: No, this is more to update metadata in a specific resource, not to add items or so.

jdries:
@bossie you implemented export_workspace, did we have any remarks on the process description?
Can you perhaps also check stac_modify? As we also plan to implement that one.

jdries:
Here's a next step in how I plan to use this to export openEO results directly into a stac api:

- save_result -> generates static stac
- stac_modify -> update metadata as needed
- export_workspace(merge="https://openeo.stac.vito.be/.../my_user_collection",workspace="my_workspace")

For this to work, the openeo service needs to have permission to write into the referenced stac collection, or a presigned url with write permission needs to be provided. (A presigned url mechanism for writing is not part of the stac spec afaik,...)

PR #490, opened 2024-01-03
Clarify no-data value / null / NaN for most processes #480

### LINKED ISSUE #480: null vs NaN
Historically, the openEO processes are using `null` to encode no-data values due to the fact that JSON can't encode NaN (and +/-Infinity). This was always only meant to be a placeholder for communication through the API. Internally, this can be anything. For some data it might be 0 or 255, for some it might be NaN or null, it pretty much depends on the underlying implementation. The actual no-data values were meant to be encoded in the collection and output metadata. Thus, it was always meant that internally the processes can pass around NaN and +/-Infinity.

Now that we are starting with the test suite for individual processes, it occurs that we write tests and expect null to be handled in internal interfaces. Also, the process definitions get a bit akward if we define behavior for null and NaN, e.g. in #479. Sometimes the behvaior is even different (i.e. our null definition derives from what IEEE 754 defines for NaN). And is NaN covered by the ignore_nodata parameters?

I'd like to discuss how to handle this.

m-mohr:
Fixes #480


array_interpolate_linear.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "array_interpolate_linear",
     "summary": "One-dimensional linear interpolation for arrays",
-    "description": "Performs a linear interpolation for each of the no-data values (`null`) in the array given, except for leading and trailing no-data values.\n\nThe linear interpolants are defined by the array indices or labels (x) and the values in the array (y).",
+    "description": "Performs a linear interpolation for each of the NaN and no-data values in the array given, except for leading and trailing NaN and no-data values.\n\nThe linear interpolants are defined by the array indices or labels (x axis) and the values in the array (y axis).",
```
soxofaan, line 4: > NaN and no-data

Isn't it easier to just speak about no-data values and consider NaN to be a case of no-data (e.g. ina IEEE float context). This note on NaN might be left implicit for reader, or mentioned explicitly.

array_find.json
```diff
@@ -19,7 +19,7 @@
         },
         {
             "name": "value",
-            "description": "Value to find in `data`. If the value is `null`, this process returns always `null`.",
+            "description": "Value to find in `data`. If the value is a no-data value (or `null`), this process returns the no-data value (or `null`).",
```
soxofaan, line 22: > If ... this process returns the no-data value (or `null`).

It looks a bit confusing to specify "return ... or ..." here. Why not just say "return the no-data value"? This no-data value might be null, but that depends on the context.

array_find.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Returns the zero-based index of the first (or last) occurrence of the value specified by `value` in the array specified by `data` or `null` if there is no match. Use the parameter `reverse` to switch from the first to the last match.\n\n**Remarks:**\n\n* Use ``array_contains()`` to check if an array contains a value regardless of the position.\n* Use ``array_find_label()`` to find the index for a label.\n* All definitions for the process ``eq()`` regarding the comparison of values apply here as well. A no-data return value from ``eq()`` is handled as `false` (no match).\n* Data types MUST be checked strictly. For example, a string with the content *1* is not equal to the number *1*.\n* An integer *1* is equal to a floating-point number *1.0* as `integer` is a sub-type of `number`. Still, this process may return unexpectedly `false` when comparing floating-point numbers due to floating-point inaccuracy in machine-based computation.\n* Temporal strings are treated as normal strings and MUST NOT be interpreted.\n* If the specified value is an array, object or null, the process always returns `null`. See the examples to find no-data values.",
```
soxofaan, line 4: > Returns the zero-based index of ..  or `null` if there is no match.

Should it be "... the no-data value if there is no match"?

arctan2.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "arctan2",
     "summary": "Inverse tangent of two numbers",
-    "description": "Computes the arc tangent of two numbers `x` and `y`. It is similar to calculating the arc tangent of *`y / x`*, except that the signs of both arguments are used to determine the quadrant of the result.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated if any of the arguments is `null`.",
+    "description": "Computes the arc tangent of two numbers `x` and `y`. It is similar to calculating the arc tangent of *`y / x`*, except that the signs of both arguments are used to determine the quadrant of the result.\n\nWorks on radians only.\nIf any argument is a no-data value, the result will be the no-data value (or `null`).",
```
soxofaan, line 4:
```suggestion
    "description": "Computes the arc tangent of two numbers `x` and `y`. It is similar to calculating the arc tangent of *`y / x`*, except that the signs of both arguments are used to determine the quadrant of the result.\n\nWorks on radians only.\nIf any argument is a no-data value, the result will be the no-data value.",
```

soxofaan (commented):
I haven't got through this giant PR yet, but here are some comments already.

Also note that some comments probably apply to multiple other places, I didn't cover yet

all.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Checks if **all** of the values in `data` are true. If no value is given (i.e. the array is empty) the process returns `null`.\n\nBy default all no-data values are ignored so that the process returns the no-data value (or `null`) if all values are no-data, `true` if all values are true and `false` otherwise. Setting the `ignore_nodata` flag to `false` takes no-data values into account and the array values are reduced pairwise according to the following truth table:\n\n```\n        || no-data | false | true\n------- || ------- | ----- | -------\nno-data || no-data | false | no-data\nfalse   || false   | false | false\ntrue    || no-data | false | true\n```\n\n**Remark:** The process evaluates all values from the first to the last element and stops once the outcome is unambiguous. A result is ambiguous unless a value is `false` or all values have been taken into account.",
```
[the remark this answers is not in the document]
m-mohr: Yeah, I've updated that.

Indeed, I just checked R and it also returns TRUE, Julia as well.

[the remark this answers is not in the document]
m-mohr: Issues with a bit more context: https://github.com/Open-EO/openeo-processes/issues/494

arccos.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "arccos",
     "summary": "Inverse cosine",
-    "description": "Computes the arc cosine of `x`. The arc cosine is the inverse function of the cosine so that *`arccos(cos(x)) = x`*.\n\nWorks on radians only.\nThe no-data value `null` is passed through and therefore gets propagated. `NaN` is returned for values outside of the allowed range.",
+    "description": "Computes the arc cosine of `x`. The arc cosine is the inverse function of the cosine so that *`arccos(cos(x)) = x`*.\n\nWorks on radians only.\nNo-data values are passed through and therefore get propagated. `NaN` is returned for values outside of the allowed range.",
```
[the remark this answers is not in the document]
m-mohr: Fixed

absolute.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "absolute",
     "summary": "Absolute value",
-    "description": "Computes the absolute value of a real number `x`, which is the \"unsigned\" portion of `x` and often denoted as *|x|*.\n\nThe no-data value `null` is passed through and therefore gets propagated.",
+    "description": "Computes the absolute value of a real number `x`, which is the \"unsigned\" portion of `x` and often denoted as *|x|*.\n\nNo-data values are passed through and therefore get propagated.",
```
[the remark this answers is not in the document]
m-mohr: Yeah, I removed " and therefore get propagated" from all process descriptions.

between.json
```diff
@@ -38,7 +38,7 @@
         }
     ],
     "returns": {
-        "description": "`true` if `x` is between the specified bounds, `null` if `x` is a no-data value, `false` otherwise.",
+        "description": "`true` if `x` is between the specified bounds, the no-data value (or `null`) if `x` is a no-data value, `false` otherwise.",
```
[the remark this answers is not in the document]
m-mohr: That is intentional as in the other processes, see the implementation guide for details.

array_interpolate_linear.json
m-mohr: no-data and NaN may be different values, so I think it's better to make it explicit.

array_find.json
m-mohr: See the implementation guide for an explanation.

array_find.json
m-mohr: Fixed

array_find.json
m-mohr: > **Array processes:** Some array processes (e.g. `array_find` or `any`) use `null` as a return value.
> In the context of data cube operations (e.g. in `reduce_dimension`), `null` values returned
> by the array processes shall be replaced with the no-data value of the data cube.
> As the processes may be used outside of data cubes where the no-data values are undefined,
> most processes fall back to `null` in this case (reflected through the mention of "(or `null`)" in the process description).
> This effectively means that `null` is the default value for an undefined no-data value in openEO.

soxofaan:
I just stumbled on this old PR where I'm expected to give a follow-up review, but in the meantime it got marked as "has conflicts that must be resolved" (but github doesn't show me more info)

Is this still a valid PR to chase?

m-mohr:
Yes, this PR is still valid. I've resolved the conflicts.

m-mohr:
Updated the PR again to the latest draft version.

This PR is important for #468 and, if we want to have it, we should certainly have it with rc.2 and not just for the stable release to ensure the changes make sense and work as expected.

CHANGELOG.md
```diff
@@ -17,18 +17,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- `apply_polygon`: Replaced outdated usage of `raster-cube` subtype with `datacube` and dimensions. [#524](https://github.com/Open-EO/openeo-processes/issues/524)
+- Clarified for various processes the handling of no-data values and null, see also the [implementation guide](meta/implementation.md).  [#480](https://github.com/Open-EO/openeo-processes/issues/480)
 - `aggregate_temporal` and `aggregate_temporal_period`: Clarified that the process throws a `DimensionNotAvailable` exception when no temporal dimension exists.
-- `aggregate_temporal_period`: Removed unused exception `DistinctDimensionLabelsRequired`
-- `aggregate_temporal_period`: Clarified that the definition of weeks follows ISO 8601
+- `aggregate_temporal_period`: Removed unused exception `DistinctDimensionLabelsRequired`.
+- `aggregate_temporal_period`: Clarified that the definition of weeks follows ISO 8601.
+- `apply_polygon`: Replaced outdated usage of `raster-cube` subtype with `datacube` and dimensions. [#524](https://github.com/Open-EO/openeo-processes/issues/524)
+- `array_interpolate_linear`: Apply interpolation to NaN and no-data values.
```
soxofaan, line 39: out curiosity: is there a reason for not only adding changelog items, but also changing the order of non-related items?

soxofaan (approved):

soxofaan:
I did another superficial pass and I think this is fine to merge, at least to unblock other progress

PR #491, opened 2024-01-03
`load_collection` and `load_stac`: Clarify the order of dimension labels for nominal labels. #488

### LINKED ISSUE #488: load_stac: band order remark
https://github.com/Open-EO/openeo-processes/blob/ad8a2f3551fe9d69c1689e310bf5307e20d86113/proposals/load_stac.json#L4

has this remark:

> The bands (and all dimensions that specify nominal dimension labels) are expected to be ordered as specified in the metadata if the bands parameter is set to null.

What does "ordered as specified in the metadata" mean practically? 
Also, doesn't this heavily depends on STAC extensions in play (if any)?

For example, take a STAC Item like this (using the eo STAC extension):
```json
{
  "type": "Feature",
  "assets": {
    "B04": {
      "eo:bands": [{"name": "B04"}],
      ...
    "B05": {
      "eo:bands": [{"name": "B05"}],  
      ...
```

I assume the spirit of the remark above is to take band order ["B04", "B05"], but this comes from the "assets" *mapping*, which technically does not imply an order.

m-mohr:
Fixes #488


proposals/load_stac.json
```diff
@@ -1,7 +1,7 @@
...
...ilable for the given extents, a `NoDataAvailable` exception is thrown.\n\n**Remarks:**\n\n* The dimensions (e.g. names) follow the data cube metadata (`cube:dimensions`), if present. Otherwise, it tries to preserve any dimension names available in the files. Otherwise, it falls back to the recommended dimension names as specified in the openEO API (`x`, `y`, `z`, `t`, `bands`, `geometry`).\n* All dimensions that specify nominal dimension labels (e.g. bands) are expected to be ordered as specified in the data cube metadata (`cube:dimensions`) unless otherwise specified in a corresponding parameter (e.g. `bands`). If no data cube metadata is present, the order of the dimension labels will reflect the structure in the files.\n* If no additional parameter is specified this would imply that the whole data set is expected to be loaded. Due to the large size of many data sets, this is not recommended and may be optimized by back-ends to only load the data that is actually required after evaluating subsequent processes such as filters. This means that the values should be processed only after the data has been limited to the required extent and as a consequence also to a manageable size.",
```
soxofaan, line 4: > If no data cube metadata is present, the order of the dimension labels will reflect the structure in the files.

Isn't it safer to say that the order will be undefined? 
Otherwise users might be tricked into depending on  implementation details that can suddenly change. Also, "structure in the files" might be not as straightforward practically as it sounds (e.g. multiple files where band order is different).

If a predictable order is important, we might also prescribe to use alphabetical order if no other reliable band order source is available.

m-mohr: For us, maybe. For the user, undefined is the worst case scenario, I think.

load_stac is anyway implementation heavily implementation detail dependant, the order of the bands is just a tiny detail of it. If you change band order in the source files, you may change a lot more (think of recent S2 changes with the offsets etc).  So I don't see an issue here (at least not only for bands).

Alphabetical order as in A1, A10, A2 or A1, A2, A10? ;-)

load_collection.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Loads a collection from the current back-end by its id and returns it as a processable data cube. The data that is added to the data cube can be restricted with the parameters `spatial_extent`, `temporal_extent`, `bands` and `properties`. If no data is available for the given extents, a `NoDataAvailable` exception is thrown.\n\n**Remarks:**\n\n* All dimensions that specify nominal dimension labels (e.g. bands) are expected to be ordered as specified in the data cube metadata (`cube:dimensions`) unless otherwise specified in a corresponding parameter (e.g. `bands`).\n* If no additional parameter is specified this would imply that the whole data set is expected to be loaded. Due to the large size of many data sets, this is not recommended and may be optimized by back-ends to only load the data that is actually required after evaluating subsequent processes such as filters. This means that the values in the data cube should be processed only after the data has been limited to the required extent and as a consequence also to a manageable size.",
```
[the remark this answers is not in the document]
m-mohr: Changed.

proposals/load_stac.json
soxofaan: > For the user, undefined is the worst case scenario, I think.

I'm not sure. I'd rather prefer an simple honest statement that the order is undefined than getting the impression that the order is static (unless you understand the fine print with a lot of technical jargon).

But anyway, I guess the main message should be: dear user, if you care about the band order, specify it explicitly

soxofaan: and if alphabetical is an option I would keep it simple to pure alphabetical: B1, B10, B2. The main goal is to be predictable, not trying to guess what makes most sense for humans.

clausmichele: I fully agree with Stefaan. And having an explicit statement telling that for some edge cases the order can't be defined automatically and the best practice consist in defining explicitly the bands to load and their order.

proposals/load_stac.json
```diff
@@ -1,7 +1,7 @@
...
... no data is available for the given extents, a `NoDataAvailable` exception is thrown.\n\n**Remarks:**\n\n* The dimensions (e.g. names) follow the data cube metadata (`cube:dimensions`), if present. Otherwise, it tries to preserve any dimension names available in the files. Otherwise, it falls back to the recommended dimension names as specified in the openEO API (`x`, `y`, `z`, `t`, `bands`, `geometry`).\n* All dimensions that specify nominal dimension labels (e.g. bands) are ordered as specified in the data cube metadata (`cube:dimensions`) unless otherwise specified in a corresponding parameter (e.g. `bands`). If no data cube metadata is present, the order of the dimension labels will reflect the structure in the files.\n* If no additional parameter is specified this would imply that the whole data set is expected to be loaded. Due to the large size of many data sets, this is not recommended and may be optimized by back-ends to only load the data that is actually required after evaluating subsequent processes such as filters. This means that the values should be processed only after the data has been limited to the required extent and as a consequence also to a manageable size.",
```
soxofaan, line 4: (the "suggestion snippet" feature of github does not seem to work properly with this long line, so I do it low-tech style here)

original: 
> If no data cube metadata is present, the order of the dimension labels will reflect the structure in the files.

proposal:
> If no data cube metadata is present, the order of the dimension labels will be implementation-dependent (e.g. based on the structure in the files).


Although I left in this proposal for now: I'm not a fan of describing this in terms of "structure in the files":
- we don't express anything else in terms of "files", which is an alien concept in openEO (intentionally, because we love cubes). Sure, we have "file formats", but that's quite a different concern.
- in STAC (e.g. items), assets are a mapping (key -> asset object), which is unordered by definition, so that undermines any promise you can make about a predictable order based on "the structure in the files"

soxofaan (commented):
trying to blow some life in this old PR I found :smile:

proposals/load_stac.json
m-mohr: I've updated it as follows:

>  All dimensions that specify nominal dimension labels (e.g. bands) are ordered as specified in the data cube metadata (`cube:dimensions`) or other relevant STAC properties (e.g. the STAC bands array) unless otherwise specified in a corresponding parameter (e.g. `bands`). If no relevant metadata is present, the order of the dimension labels is undefined. It is recommended to provide the `bands` parameter if the order is important for the following processes.

proposals/load_stac.json
m-mohr: I've updated it as follows:

>  All dimensions that specify nominal dimension labels (e.g. bands) are ordered as specified in the data cube metadata (`cube:dimensions`) or other relevant STAC properties (e.g. the STAC bands array) unless otherwise specified in a corresponding parameter (e.g. `bands`). If no relevant metadata is present, the order of the dimension labels is undefined. It is recommended to provide the `bands` parameter if the order is important for the following processes.

soxofaan (approved):

PR #493, opened 2024-01-04
Multi-layer Perceptron classifier

PondiB:

PR #484, opened 2023-12-07
ML XGBoost  Classification

PondiB:
Specification for ML XGBoost for Classification.

Depends on #441

PR #497, opened 2024-02-21
ML Generic API

PondiB:
I saw the need for this as an alternative for #441 , if we keep following the pattern of **_ml_fit_*_** as #484 , #483 , #493  and so forth. We will end up with a long list of ml processes. 

Getting inspiration from tools like [caret](https://cran.r-project.org/web/packages/caret/index.html) in R ecosystem. We can ideally achieve a minimalistic API. 

Labels can be used to identify a classification and regression problem. 

Parameters for various ml models can be passed as dictionaries: e.g.
1. Random Forest :
`ml_fit(predictor, label, ml_method ="random_forest", params = {"num_trees":200, "max_variables":"sqrt", "seed":42})`
2. Multi-layer Perceptron classifier
`ml_fit(predictor, label, ml_method = "mlp", params = {"hidden_layers": [ 512, 512, 512], "dropout_rates": [0.4, 0.3, 0.2], "epochs": 100, "activation_function": "relu", "optimizer": "adam"}`

Well, there will be need a for clear documentation dedicated to ML on the openEO website.

Additionally, a need for 2 more methods on the client side `list_ml_models()` for listing the available models and  `model_param_blueprint(model_name, task_type="classification")`  for listing the default params of the models. The information should ideally be on the website too.

Opinions, suggestions and critics are welcomed as I am still exploring this.


soxofaan:
> We can ideally achieve a minimalistic API.

I'm not sure that that should be the goal. 
You indeed have a small API at the level of the processes (just one process `ml_fit`), but you pay the cost in other places: those `list_ml_models()` and `model_param_blueprint()` have to be defined, you need to provide schema definition/validation for the `params` of each model, all clients have to start supporting that, all back-ends have to start implementing that, ... It feels like a lot of reinventing things we already have in our current ecosystem of processes.
I'm not sure there is a net win here.

> We will end up with a long list of ml processes

I don't think that is a problem on its own. The list of pre-defined processes is also long, but that is fine.

PondiB:
> I'm not sure that that should be the goal.

Thanks for your solid comments that cannot be overlooked, it reminds me of this article ; [Simple is not easy](https://www.entropywins.wtf/blog/2017/01/02/simple-is-not-easy/). There will be an initial Engineering overhead for sure as you have pointed out. My goal for this was more targeted to the general public/users i.e. Greater accessibility and ease of use.

> I don't think that is a problem on its own. The list of pre-defined processes is also long, but that is fine.

On my end, It does not mean we should have 20+ processes for ML/DL.

soxofaan:
> My goal for this was more targeted to the general public/users i.e. Greater accessibility and ease of use.

Even for the user I think it will be worse.

e.g. with the python client, the current way of working would result in something like 
```python
ml_fit_random_forest(predictors=cube1, labels=cube2, num_trees=500, max_variables="sqrt", seed=42)
```
Which is easy/simple and highly accessible I would think: it just uses the basics of how functions are called, everything is documented the normal way, not only the available arguments and their defaults, but also basics of random forest, ...

the proposed approach would result in something like
```python
ml_fit(predictors=cube1, labels=cube2, ml_method="random_forest", params = {"num_trees":200, "max_variables":"sqrt", "seed":42})
```
This is less easy/simple for the user: the `params` is now a generic and hard to document thing and heavily depends on `ml_method`, `ml_fit` has to explain all possible ML methods, there is no reasonable default for `ml_method`, ...

soxofaan:
> Therefore internally it can be determined from the labels if this is a classification or a regression task

I'm not sure this is a safe assumption. 
I can imagine there are users that use e.g. integer values as classification labels.

PondiB:
Closing this, I will continue with the already existing approach with a few changes as in #517

PR #505, opened 2024-05-17
Add text_position

m-mohr:


soxofaan:
This process is pretty similar to `array_find`, so it might be good to align with that one:
- `text_find` instead of `test_position`
- `array_find` returns null when not found instead of -1
- rename `pattern` arg to `value` too?

m-mohr:
Good point.

>     * `text_find` instead of `test_position`

Sounds good. +1

>     * `array_find` returns null when not found instead of -1

I don't have strong preferences here. +0

>     * rename `pattern` arg to `value` too?

pattern is coming from text_contains, text_begins, ...
I think having it consistent in text_* is more important than having it consistent with array_find. -1

soxofaan:
> pattern is coming from text_contains, text_begins, ...

good point


Another thing that is worth mentioning is that the returned position is a zero-based index (as mentioned in array_find)

m-mohr:
Isn't that captured by the description of the return value?

> A value >= 0 that indicates the position of the text

If not, happy to accept proposals how to improve.

soxofaan:
I just mentioned it because I noticed it's the first statement in the  description of array_find
> Returns the zero-based index of ...

while it's not explicitly mentioned in current PR of text_find

m-mohr:
Updated PR to be aligned with the draft branch.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `run_ogcapi`
 - `run_ogcapi_externally`
 - `stac_modify`
+- `text_position`
```
soxofaan, line 17:
```suggestion
- `text_find`
```

soxofaan (approved):
minor fix suggested, but fine to merge apart from that

Issue #503, opened 2024-05-08
 load_stac: scale_and_offset


clausmichele:
**Proposed Process ID:** load_stac
**Proposed Parameter Name:** scale_and_offset
**Optional:** yes, default: False

## Context
Recently, after the introduction of the new Sentinel-2 processing baseline, an offset has been introduced (additionally to the scale, which was already present). Previously, since the conversion from digital number (DN) (the actual values in the S2 files) to reflectances was performed only by applying the scale factor, for many applications it was the same as using DN (in many indexes, the scale factor is being neglected).
Now, both have to be applied in order to obtain meaningful results.

Some GitHub issues discussing about this topic:
https://github.com/Element84/earth-search/issues/9
https://github.com/Element84/earth-search/issues/23
https://github.com/opendatacube/odc-stac/issues/55

@jdries @dthiex How do you manage this for the SENTINEL2_L2A collections?

## Description
if `scale_and_offset` is True, apply them automatically. They should be available in the raster:bands extension metadata.

## Data Type
boolean

dthiex:
In SH itself we follow [this](https://forum.sentinel-hub.com/t/sentinel-2-processing-baseline-changes-and-harmonizevalues/4635) approach (we have a harmonization parameter which if set to true will already compensate the offset).

In openEO SH I believe we don't support setting the parameter so we apply the default meaning we request harmonize DN so DN = 10000*Reflectance is still true.

In my feeling this should though not be part of the load_stac process as it's very specific to the case "I load L2A raw DNs but I want to get Reflectance values".

jdries:
we currently also configure the behaviour per collection, but most of them require the user to explicitly do it, which is annoying.
For Sentinel-2 we make sure to convert to the 'standard' scaling factor of 0.0001, to avoid issues with the new processing baseline.
It would be nice to have a generic solution for load_collection as well.

soxofaan:
yeah, it would be nice if this could be addressed the same way in both load_collection and load_stac. 

Side note: I wonder if there isn't a more generic or future-proof parameter name than `scale_and_offset` to not be limited to just `scale` and `offset` transforms. For example in the SH link of Daniel I see clamping of negative values.

m-mohr:
The original idea was to have these information in metadata and then apply it automatically during data loading, so I assumed that this is already done and doesn't need a parameter.

[...]

PR #512, opened 2024-06-14
`load_stac` scale and offset clarifications

clausmichele:
1. I moved the json spec out of the proposals folder, since now the process is available on many different back-ends CDSE, VITO, openEO Platform, EODC, Eurac Research.
2. I tried to clarify the behavior of the process in case of Items/Assets with associated `scale` and `offset` parameters, which from this discussion we aggreed shouldn't be applied by default: https://github.com/Open-EO/openeo-processes/issues/503


clausmichele:
@m-mohr what should I add to make the `load_stac` tests passing? I'm not sure to understand. This is also the result of a local tests run:

```
npm test

> @openeo/processes@2.0.0 test /home/mclaus@eurac.edu/openeo-processes/tests
> openeo-processes-lint testConfig.json

/home/mclaus@eurac.edu/openeo-processes/tests/node_modules/@jest/core/build/SearchSource.js:137
  _context;
          ^

SyntaxError: Unexpected token ;
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/mclaus@eurac.edu/openeo-processes/tests/node_modules/@jest/core/build/index.js:30:44)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
npm ERR! Test failed.  See above for more details.
```

m-mohr:
Applies to both load_collection and load_stac.
We should add for both processes something like: 

> scale and offset need to be applied manually if present in metadata.

Not sure whether we should add:

> If scale and offset are not present in metadata, no scale and offset need to be applied.

soxofaan:
> We should add for both processes something like:
>
>>    scale and offset need to be applied manually if present in metadata.

Instead of "manually" I would express this in terms of "client side" or "explicitly in the process graph",
so that openEO _clients_ still can provide an automatic rescaling feature in some way.

soxofaan:
> Not sure whether we should add:
>
>>    If scale and offset are not present in metadata, no scale and offset need to be applied.

I agree, I would leave this out. There might be reasons that some kind of scaling is  necessary, but it can not be expressed in structured metadata (e.g. because the scaling factors change at a certain date, or the scaling is non-linear)

clausmichele:
I forgot about this PR, I now updated the text following your suggestions @m-mohr @soxofaan

Issue #513, opened 2024-06-14
Guideline about new processes in README should be updated


clausmichele:
The current README mentions an examples folder, which no longer exists. Should we remove that line?

https://github.com/Open-EO/openeo-processes?tab=readme-ov-file#process

> Processes will only be moved from proposals to the stable process specifications once there are at least two implementations and an example process in the [examples](https://github.com/Open-EO/openeo-processes/blob/master/examples) folder showing it in a use case. This doesn't require a PSC vote individually as it's not a breaking change, just an addition.

PR #514, opened 2024-06-25
Remove mention to examples folder

clausmichele:
Fixes https://github.com/Open-EO/openeo-processes/issues/513

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This repository contains a set of files formally describing the openEO Processes
 ## Process
 
 * All new processes must be added to the [`proposals`](proposals/) folder.
-* Processes will only be moved from proposals to the stable process specifications once there are at least two implementations. This doesn't require a PSC vote individually as it's not a breaking change, just an addition.
+* Processes will only be moved from proposals to the stable process specifications once there are at least two implementations and an example process in the [`openEO community examples`](https://github.com/Open-EO/openeo-community-examples/) showing it in a use case. This doesn't require a PSC vote individually as it's not a breaking change, just an addition.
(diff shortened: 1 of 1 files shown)


clausmichele:
@m-mohr can we merge this simple fix?

m-mohr:
I think we had a PSC decision on this and I hadn't had the time yet to check what the actual outcome was. 
I just want to update it directly in the right way, not merge yet another potentially wrong thing.

README.md
```diff
@@ -40,7 +40,7 @@ This repository contains a set of files formally describing the openEO Processes
 ## Process
 
 * All new processes must be added to the [`proposals`](proposals/) folder.
-* Processes will only be moved from proposals to the stable process specifications once there are at least two implementations and an example process in the [`examples`](examples/) folder showing it in a use case. This doesn't require a PSC vote individually as it's not a breaking change, just an addition.
+* Processes will only be moved from proposals to the stable process specifications once there are at least two implementations. This doesn't require a PSC vote individually as it's not a breaking change, just an addition.
```
m-mohr, line 43: We never explicitly discussed to remove this requirement (which we could discuss due to the fact that openeo-community-examples is pretty unmaintained), but we did discuss moving the examples:
https://github.com/Open-EO/PSC/issues/14

So the correct way currently is to change as follows unless another PSC discussion takes place:
```suggestion
* Processes will only be moved from proposals to the stable process specifications once there are at least two implementations and an example process in the [`openEO community examples`](https://github.com/Open-EO/openeo-community-examples/) showing it in a use case. This doesn't require a PSC vote individually as it's not a breaking change, just an addition.
```

clausmichele: Fine, thanks for the feedback

soxofaan:
not sure if intended, but I noticed this PR was applied to master instead of going through draft

PR #518, opened 2024-07-18
`apply_polygon`: rename `polygons` parameter to `geometries` (#511)

### LINKED ISSUE #511: `apply_polygon`: rename `polygons` parameter to `geometries`
`apply_polygons` has a `polygons` argument:
https://github.com/Open-EO/openeo-processes/blob/965bbaebd4d5984203a0437076c85a66a72a23e0/proposals/apply_polygon.json#L19-L20

This is not in line with comparable processes that use `geometries` for the same concept, like `aggregate_spatial`, `filter_spatial`, `filter_vector`, `vector_buffer`. Note that `mask_polygon` also doesn't use `polygons`.

It also appears to be confusing users about how MultiPolygons are supported

soxofaan:
fixes #511


PondiB (approved):
👍🏾

PR #519, opened 2024-07-26
`run_udf`: Simplified and clarified the schema for `data` #515

### LINKED ISSUE #515: run_udf: Unclear why two data types are defined
**Process ID:** run_udf

**Describe the issue:**
Run UDF has the following definition for `data`:

```json
        {
            "name": "data",
            "description": "The data to be passed to the UDF.",
            "schema": [
                {
                    "title": "Array",
                    "type": "array",
                    "minItems": 1,
                    "items": {
                        "description": "Any data type."
                    }
                },
                {
                    "title": "Single Value",
                    "description": "A single value of any data type."
                }
            ]
        }
```

Why does it explicitly define an array type? The way the second schema is defined, both apply for an array. So it's equivalent to:

```json
        {
            "name": "data",
            "description": "The data to be passed to the UDF.",
            "schema": {
                "title": "Single Value",
                "description": "A single value of any data type."
            }
        }
```

Can someone remember why this is the way it is? @soxofaan maybe?

**Proposed solution:**
Simplify the schema?

m-mohr:
Fixes #515

### DIFF
--- a/proposals/run_udf_externally.json
+++ b/proposals/run_udf_externally.json
@@ -11,21 +11,10 @@
     "parameters": [
         {
             "name": "data",
-            "description": "The data to be passed to the UDF.",
-            "schema": [
-                {
-                    "title": "Array",
-                    "type": "array",
-                    "minItems": 1,
-                    "items": {
-                        "description": "Any data type."
-                    }
-                },
-                {
-                    "title": "Single Value",
-                    "description": "A single value of any data type."
-                }
-            ]
+            "description": "The data to be passed to the UDF.\n\nThe data must be given in a way that the UDF can understand it. Usually, `run_udf` is executed as part of a data cube process such as `reduce_dimension` and in this case the process would expect an array of values as provided in the parameter `data` by `reduce_dimension`. Please refer to the documentation of the UDF runtime for details on how to provide the data.",
+            "schema": {
+                "description": "A value of any data type."
+            }
         },
         {
             "name": "url",
--- a/proposals/run_udf_externally.json
+++ b/proposals/run_udf_externally.json
@@ -1,7 +1,7 @@
 {
     "id": "run_udf_externally",
     "summary": "Run an externally hosted UDF container",
-    "description": "Runs a compatible UDF container that is either externally hosted by a service provider or running on a local machine of the user. The UDF container must follow the [openEO UDF specification](https://openeo.org/documentation/1.0/udfs.html).\n\nThe referenced UDF service can be executed in several processes such as ``aggregate_spatial()``, ``apply()``, ``apply_dimension()`` and ``reduce_dimension()``. In this case, an array is passed instead of a data cube. The user must ensure that the data is given in a way that the UDF code can make sense of it.",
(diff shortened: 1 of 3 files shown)


run_udf.json
```diff
@@ -10,21 +10,10 @@
...
-                    "items": {
-                        "description": "Any data type."
-                    }
-                },
-                {
-                    "title": "Single Value",
-                    "description": "A single value of any data type."
-                }
-            ]
+            "description": "The data to be passed to the UDF.\n\nThe data must be given in a way that the UDF can understand it. Usually, this process is executed as part of a data cube process such as `reduce_dimension` and in this case the process would expect an array of values as provided in the parameter `data` by `reduce_dimension`. Please refer to the documentation of the UDF runtime for details on how to provide the data.",
```
soxofaan, line 13: > The data must be given in a way that the UDF can understand it. 

From my UDF-usage experience this statement sounds a bit backwards. When working with UDFs in practice, the data and its form are a given and you have to tweak your UDF to match that form.
Basically in practice,  the "parent" process (e.g. reduce_dimension) dictates the form of `data`, so it might be confusing to the reader trying to understand how `run_udf` operates by stating something like "the data must be given in a way ..."

I know that it's possible to have processes before `run_udf` in  `reduce_dimension`'s reducer, which would allow for some kind of data preparation, but I don't think there is an implementation that supports that (the geopyspark one doesn't in any case).

Also note that there isn't necessarily direct correspondence between the `data` argument of `run_udf` and the actual data that is available in the UDF code. For example: with `apply`, the `data` of the "child" callback is a single pixel value, but the UDF code could still receive larger multidimensional chunks for performance reasons. So I would avoid over-specifying the link between `run_udf`'s  `data` parameter and the actual data container available in the UDF.

> Please refer to the documentation of the UDF runtime for details on how to provide the data.

Likewise: the UDF runtime docs can provide little help here, it's the parent process that dictates the form of `run_udf`'s `data` parameter

Attempt to tweak the description:
```suggestion
            "description": "The data to be passed to the UDF.\n\nUsually, `run_udf` process is used as a child process of a data cube process such as `reduce_dimension` or `apply`. These data cube processes establish the actual type of this data. For example, in case of `reduce_dimension`, `data` will be an array of values, while in case of `apply` it will be a single scalar value.",
```

proposals/run_udf_externally.json
```diff
@@ -11,21 +11,10 @@
...
-                    "items": {
-                        "description": "Any data type."
-                    }
-                },
-                {
-                    "title": "Single Value",
-                    "description": "A single value of any data type."
-                }
-            ]
+            "description": "The data to be passed to the UDF.\n\nThe data must be given in a way that the UDF can understand it. Usually, this process is executed as part of a data cube process such as `reduce_dimension` and in this case the process would expect an array of values as provided in the parameter `data` by `reduce_dimension`. Please refer to the documentation of the UDF runtime for details on how to provide the data.",
```
soxofaan, line 14: see `run_udf` comment

run_udf.json
soxofaan: I'm also fine to just leave the description as is ("The data to be passed to the UDF."), to avoid bike-shedding this too much.

m-mohr: Sounds good.

PR #526, opened 2024-12-11
Add generic aggregate process

m-mohr:
Based on aggregate_temporal.

Example:

### DIFF
--- a/proposals/aggregate.json
+++ b/proposals/aggregate.json
@@ -1,0 +1,135 @@
+{
+    "id": "aggregate",
+    "summary": "Aggregation based on general intervals",
+    "description": "Computes a aggregation based on an array of intervals.\n\nThe computed values will be projected to the labels. If no labels are specified, the lower value of the interval will be used as label for the corresponding values. In case of a conflict (i.e. the user-specified values for the lower values of the intervals are not distinct), the user-defined labels must be specified in the parameter `labels` as otherwise a `DistinctDimensionLabelsRequired` exception would be thrown. The number of user-defined labels and the number of intervals need to be equal.",
+    "categories": [
+        "cubes",
+        "aggregate"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "data",
+            "description": "A data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
+            }
+        },
+        {
+            "name": "intervals",
+            "description": "Left-closed intervals, which are allowed to overlap. Each interval in the array has exactly two elements:\n\n1. The first element is the lower value of the interval. The specified value is **included** in the interval.\n2. The second element is the upper value of the temporal interval. The specified value is **excluded** from the interval.\n\nThe second element must always be greater than the first element. Otherwise, an `ExtentEmpty` exception is thrown.",
+            "schema": {
+                "type": "array",
+                "minItems": 1,
+                "items": {
+                    "type": "array",
+                    "uniqueItems": true,
+                    "minItems": 2,
+                    "maxItems": 2,
+                    "items": {
+                        "type": "number"
+                    }
+                }
+            }
+        },
+        {
+            "name": "reducer",
+            "description": "A reducer to be applied for the values contained in each interval. A reducer is a single process such as ``mean()`` or a set of processes, which computes a single value for a list of values, see the category 'reducer' for such processes. Intervals may not contain any values, which for most reducers leads to no-data (`null`) values by default.",
+            "schema": {
+                "type": "object",
+                "subtype": "process-graph",
+                "parameters": [
+                    {
+                        "name": "data",
+                        "description": "A labeled array with elements of any type. If there's no data for the interval, the array is empty.",
+                        "schema": {
+                            "type": "array",
+                            "subtype": "labeled-array",
+                            "items": {
+                                "description": "Any data type."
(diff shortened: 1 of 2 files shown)


proposals/aggregate.json
```diff
@@ -0,0 +1,135 @@
+{
+    "id": "aggregate",
+    "summary": "Aggregation based on general intervals",
+    "description": "Computes a aggregation based on an array of intervals.\n\nThe computed values will be projected to the labels. If no labels are specified, the lower value of the interval will be used as label for the corresponding values. In case of a conflict (i.e. the user-specified values for the lower values of the intervals are not distinct), the user-defined labels must be specified in the parameter `labels` as otherwise a `DistinctDimensionLabelsRequired` exception would be thrown. The number of user-defined labels and the number of intervals need to be equal.",
```
soxofaan, line 4: > The computed values will be projected to the labels

what does "projecting" to a label mean?

proposals/aggregate.json
```diff
@@ -0,0 +1,135 @@
...
+            "name": "data",
+            "description": "A data cube.",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
+            }
+        },
+        {
+            "name": "intervals",
+            "description": "Left-closed intervals, which are allowed to overlap. Each interval in the array has exactly two elements:\n\n1. The first element is the lower value of the interval. The specified value is **included** in the interval.\n2. The second element is the upper value of the temporal interval. The specified value is **excluded** from the interval.\n\nThe second element must always be greater than the first element. Otherwise, an `ExtentEmpty` exception is thrown.",
```
soxofaan, line 21: when working with a spatial dimension (e.g. "x" or "y"): how is a users supposed to know what CRS to use to define the intervals? I don't think this is explicitly available or defined on a datacube.

proposals/aggregate.json
```diff
@@ -0,0 +1,135 @@
...
+            "schema": {
+                "type": "array",
+                "minItems": 1,
+                "items": {
+                    "type": "array",
+                    "uniqueItems": true,
+                    "minItems": 2,
+                    "maxItems": 2,
+                    "items": {
+                        "type": "number"
```
soxofaan, line 31: So intervals are only based on numbers? That implies that this process only works along a spatial dimension in practice?
Other dimensions don't have numeric labels: band dimensions have strings and temporal dimensions have date/datetime (subtype of string)

proposals/aggregate.json
```diff
@@ -0,0 +1,135 @@
...
+            "description": "Additional data to be passed to the reducer.",
+            "schema": {
+                "description": "Any data type."
+            },
+            "optional": true,
+            "default": null
+        }
+    ],
+    "returns": {
+        "description": "A new data cube with the same dimensions. The dimension properties (name, type, labels, reference system and resolution) remain unchanged, except for the resolution and dimension labels of the given temporal dimension.",
```
soxofaan, line 106: > of the given temporal dimension

is it intentional to have "temporal" there, or should this be generic for all types of dimensions?

proposals/aggregate.json
soxofaan: > The second element is the upper value of the temporal interval.

I guess that "temporal" is not intentional there

soxofaan (commented):
some quick notes

soxofaan:
>  https://gitlab.ogc.org/ogc/T20-GDC/-/issues/32#note_34824 (likely access restricted)

Indeed I don't have access

proposals/aggregate.json
m-mohr: Isn't it? The datacube extension defines the CRS and then that defines the unit and extents etc.

```suggestion
            "description": "Left-closed intervals, which are allowed to overlap. Each interval in the array has exactly two elements:\n\n1. The first element is the lower value of the interval. The specified value is **included** in the interval.\n2. The second element is the upper value of the interval. The specified value is **excluded** from the interval.\n\nThe second element must always be greater than the first element. Otherwise, an `ExtentEmpty` exception is thrown.",
```

proposals/aggregate.json
m-mohr: The order of string values are not really well-defined. As such, yes only numerical. Dimensions can have numerical values though, we have a couple of processes that just index the labels if no further information is given (e.g. apply_dimension).

proposals/aggregate.json
m-mohr:
```suggestion
        "description": "A new data cube with the same dimensions. The dimension properties (name, type, labels, reference system and resolution) remain unchanged, except for the resolution and dimension labels of the given dimension.",
```

No, just copy-pasted things poorly from aggregate_temporal ;-)

proposals/aggregate.json
m-mohr: "mapped" is probably better than "projected". Maybe it can also just be removed.

(I should also check whether this wording also exists in aggregate_temporal...)

proposals/aggregate.json
soxofaan: > The datacube extension defines the CRS and then that defines the unit and extents etc.

Yes, but nothing that is relevant to figuring out the actual labels (`reference_system`, `step`, `values` and what not) is required (in datacube extension, nor openEO API). 

Also, this might be backend-dependend, so that would undermine the reproducibility of the process graph

Issue #280, opened 2021-09-22
"Overlap-only" mode for merge_cubes


soxofaan:
`merge_cubes` currently works in "union" mode: the result covers the union of the label ranges for each dimension, which is useful for use cases where you want to stitch together several data subsets containing same kind of data. The main purpose of the overlap resolver is to have a clean seam between them. You have to be careful what you do in the overlap resolver because it might not be applied to all parts of the result (depending on the degree of overlap). If you go beyond taking the mean or first non-null value, the results might be hard to reason about.

Maybe also useful is to provide a "intersection" or "overlap-only" mode of `merge_cubes` (through an additional parameter, or as a dedicated process, I'm not sure yet)  where you only return the dimension ranges where both cubes exist. It's then easier to reason about the overlap resolver because it's guaranteed to be applied everywhere. It would make `merge_cubes` also useful for use cases where you fuse different kinds of data.

soxofaan:
Another use case: https://discuss.eodc.eu/t/blocks-after-division/425
user makes division between two data cubes
```python
s2_cube_swf = s2_cube_water_sum / s2_count
```
The python client translates this to a `merge_cubes` with an overlap resolver `(x,y) -> x / y`. 
Where there is overlap, the result is `x/y` as expected, but where there is no overlap the result one of:
-  `x` where there is only first cube data
- `y` where there is only second cube data

Obviously this is a very confusing result, especially because the user is not really aware that there is a `merge_cubes` process going on. Without an "overlap-only" merge_cubes variant the user (or client lib) is forced to do some additional masking, which is tedious and comes with a performance cost.

m-mohr:
@soxofaan I'm curious: What would happen to the x/y example once we would have two processes? Would it be changed to always use the new process (or parameter)? Otherwise, without the user explixitly stating which behavior to use, it is still unclear what happens, right? And you can't really parametrize a basic math operation... ;-)

soxofaan:
In case of (client-side) translating math expressions like `x / y`, I would always, unconditionally use the "overlap-only" variant of `merge_cubes`. The things that the original `merge_cubes` would add is only confusing, if not wrong. 
As mentioned above, the user is not really aware that  an `x / y` expression is translated to a `merge_cubes` process. The user just expects the result of `x / y`, so no need to add parts where it evaluates to `x` or `y`, I can't even imagine the user would want to opt-in to that.

[...]

PR #540, opened 2025-06-02
Combine_cubes

VictorVerhaert:
Adds a new experimental `combine_cubes` process that aims at taking away complexity from merge_cubes.

The two process can now each have their distinct usage: one for performing operations that have two cubes as input and one for merging data with a possible overlap.

Related issue: #280


m-mohr:
Looks pretty good at first look.

The combine term seems a bit too close to merge though. It would be great if we could more easily distinguish what the processes do. If we'd start from scratch, we may name them `merge_cubes_union` and `merge_cubes_intersection`, but that's too late. Maybe we can call this process `intersect_cubes` or so?

soxofaan:
> I'm wondering whether the process parameter needs to be required?

Yes I think that's the core feature of this process:  
the user wants to combine cube1 and cube2 with an explicit operation (add values, take difference, calculate ratio, ....). 

It's like `apply_dimension` or `reduce_dimension` but instead of working on values along a dimension of a single cube, you work with values from two different cubes.
And to make that work, your cubes have to be "aligned" properly, which practically means we only work on the intersection and drop all the rest.

> Is this process meant to be merge cubes but only return the overlap or is this process something else? I feel the issue was asking for the first and the process is slightly different from it ("combine").

So it's a bit the other way around: working on the intersection is the consequence of requiring an operation to combine the cubes

soxofaan:
FYI another user support issue that, after hours of digging, again turned out out to be about unexpected results from merge_cubes with an unused overlap resolver:
https://forum.dataspace.copernicus.eu/t/inconsistent-rbr-pixel-values-using-openeo-sentinel-2-data/3804

soxofaan:
@VictorVerhaert   can you rebase on or merge in latest draft branch?

we had off-line discussion about this and also converged to `intersect_cubes` as  possibly more descriptive process name

PR #547, opened 2025-07-17
`load_stac`: Clarify the `properties` parameter

### LINKED ISSUE #536: load_stac properties STAC API constraints?
From description of `properties` in `load_stac`:

https://github.com/Open-EO/openeo-processes/blob/fd26251fa55ad656bff5829c5eefd3fd7c5c2707/proposals/load_stac.json#L188-L189

ends with:

> ... The value must be a condition (user-defined process) **to be evaluated against a STAC API**. 
> This parameter is **not supported for static STAC.**

I'm not sure it's necessary to add these STAC API oriented constraints. 
An openEO backend can still do property based filtering on it's own even on static STAC documents, right?

m-mohr:
`load_stac`: Clarify handling of the `properties` parameter in the context of STAC APIs and static catalogs.

Closes #536

### DIFF
--- a/proposals/load_stac.json
+++ b/proposals/load_stac.json
@@ -186,7 +186,7 @@
         },
         {
             "name": "properties",
-            "description": "Limits the data by metadata properties to include only data in the data cube which all given conditions return `true` for (AND operation).\n\nSpecify key-value-pairs with the key being the name of the metadata property, which can be retrieved with the openEO Data Discovery for Collections. The value must be a condition (user-defined process) to be evaluated against a STAC API. This parameter is not supported for static STAC.",
+            "description": "Limits the data by metadata properties to include only data in the data cube which all given conditions return `true` for (AND operation).\n\nSpecify key-value-pairs with the key being the name of the queryable (for STAC APIs) or the metadata property name (for static STAC catalogs). Queryables can be retrieved from the Queryables endpoint of the STAC API and are often but not always aligned with the metadata property names. The value must be a condition (user-defined process) to be evaluated against a STAC API or static catalog. This parameter may not be supported for static STAC catalogs.",
             "schema": [
                 {
                     "type": "object",
--- a/proposals/load_stac.json
+++ b/proposals/load_stac.json
@@ -186,7 +186,7 @@
         },
         {
             "name": "properties",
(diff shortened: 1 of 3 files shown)


proposals/load_stac.json
```diff
@@ -186,7 +186,7 @@
         },
         {
             "name": "properties",
-            "description": "Limits the data by metadata properties to include only data in the data cube which all given conditions return `true` for (AND operation).\n\nSpecify key-value-pairs with the key being the name of the metadata property, which can be retrieved with the openEO Data Discovery for Collections. The value must be a condition (user-defined process) to be evaluated against a STAC API. This parameter is not supported for static STAC.",
+            "description": "Limits the data by metadata properties to include only data in the data cube which all given conditions return `true` for (AND operation).\n\nSpecify key-value-pairs with the key being the name of the queryable (for STAC APIs) or the metadata property name (for static STAC catalogs). Queryables can be retrieved from the Queryables endpoint of the STAC API and are often but not always aligned with the metadata property names. The value must be a condition (user-defined process) to be evaluated against a STAC API or static catalog. This parameter may not be supported for static STAC catalogs.",
```
soxofaan, line 189: some minimal tweaks:

```suggestion
            "description": "Limits the data by metadata properties to include only data in the data cube which all given conditions return `true` for (AND operation).\n\nSpecify key-value-pairs with the key being the name of the queryable (for STAC APIs) or the metadata property name (for static STAC catalogs). Queryables can be retrieved from the Queryables endpoint of the STAC API and are often but not always aligned with the metadata property names. The value must be a condition (user-defined process) to be evaluated against the value of the queryable (for STAC APIs) or metadata property (for static STAc catalogs).\n\nThis parameter may not be supported for static STAC catalogs.",
```

soxofaan: And on:

> This parameter may not be supported for static STAC catalogs

I wonder if there should be an explicit note for backend implementers to raise `FeatureUnsupported` if it's not supported, to avoid that it's ignored silently

soxofaan (approved):

proposals/load_stac.json
m-mohr: I was confused about the change and first thout on your suggestion, but it was a general issue in the PR so updated the wording correspondigly.

PR #549, opened 2025-07-17
`load_geojson`: Several improvements, especially around empty propert…

### LINKED ISSUE #448: `load_geojson` issues
We're trying to support/implement a use case that depends heavily on vector cube manipulation and some questions related to `load_geojson` came up (based on [2.0.0-rc.1](https://processes.openeo.org/2.0.0-rc.1/#load_geojson) spec):

1.  description states "Feature properties are preserved", but the process only preserves properties specified in the `properties` parameter, which is empty by default, so no properties are preserved by default actually
1. `properties` has default value empty array `[]` but it is unclear what should happen in this case: there is only a geometries dimension with labels, but no "cube data". 
    - Is it valid to have (vector) data cube e.g. with shape `(N, 0)`  (N labels along "geometries" dimension and zero labels along "properties" dimension)?
    -  Or should we in that case add a dummy label in the "properties"  dimension and give all "pixels" a no-data value, so that the cube at least has a shape `(N,1)`?
1. Iterating further on the empty cube  problem: description states "A new dimension with the name `properties` and type `other` is created if at least one property is provided.". I wonder if it wouldn't be cleaner to just always add this "properties" dimension, regardless of the number of provided properties. Then e.g.:
   - empty `properties` array  `[]` => "properties" dimension with single dummy label and no-data values
   - single property in `properties` array, e.g. `["population"]` => "properties" dimension with single label "population". Advantage compared to not adding the dimension is that the property name ("population" in example) is preserved
    - multiple properties e.g. `["id", "area", "population"]` =>  as currently specified 
1. related to empty cube problem: the description of "properties" param states "Only applies for GeoJSON Features and FeatureCollections. ". Does this mean that e.g. a Polygon GeoJSON object always results in such an empty vector data cube?
1. GeoJSON properties are typically mixed: e.g. strings and int/float. Is it valid in openEO to mix these data types within the same cube (along the "properties" dimension)?

m-mohr:
`load_geojson`:
- Specify behavior if `properties` parameter is empty
- Clarify mixed data types
- Other minor clarifications

Closes #448


proposals/load_geojson.json
```diff
@@ -18,7 +18,7 @@
...
+            "description": "A list of properties from the GeoJSON file to construct an additional dimension from. A new dimension with the name `properties` and type `other` is created. If no property is provided, a label with value `id` is created and all values are set to a no-data value. Missing values are generally set to a no-data value.\n\nDepending on the number of properties provided, the process creates the dimension differently:\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nAlthough mixing data types of values is generally allowed, certain combinations of data type may be unsupported by some implementations. It is recommended to keep the data types homogeneous.",
```
soxofaan, line 21: Initial suggestion (unfortunately the diff-view does not seem to work properly on these long lines):

```suggestion
            "description": "A list of properties from the GeoJSON data to construct an additional dimension from. A new dimension with the name `properties` and type `other` is created. Missing values are generally set to a no-data value.\n\nDepending on the number of properties provided, the process creates the dimension differently:\n\n- Empty `properties` list: A single dimension label with value `id` is created and all values are set to a no-data value.\n\n- Single property with scalar values: A single dimension label with the name of the property and a single value per geometry.\n- Single property of type array: The dimension labels correspond to the array indices. There are as many values and labels per geometry as there are for the largest array.\n- Multiple properties with scalar values: The dimension labels correspond to the property names. There are as many values and labels per geometry as there are properties provided here.\n\nMissing values are generally set to a no-data value.\n\nAlthough mixing data types of values is generally allowed, certain combinations of data type may be unsupported by some implementations. It is recommended to keep the data types homogeneous.",
```


Some additional notes:

- about the `Empty properties list: ... id is created and all values are set to a no-data value`: we could also set auto-increment values, which makes this case (which is the default by the way) quite a bit more useful I would say
- about `Single property with scalar values/of type array`: I'm a bit hesitant about involving the type of the property value in the behavior of this process (e.g. to unroll arrays into separate "columns"). I guess this is to be more user friendly, but to me it feels a bit too magic and it undermines consistency and predictability. 
- Slightly related: what happens if multiple properties are listed, but some of them have array values: would that be an error?

What would feel more strict to me is to allow the `properties` argument not only be an array of strings like it is now, but also a single string; and maybe even `null`, which could then be the default:

- `properties` arg is `null` (default) -> label `id` with no-data (or auto-increment) values
- `properties` arg is array (must be non-empty) ->  values are expected to be scalar. No automatic unrolling: to be decided if it is allowed to have arrays as pixel values or to raise error here.
- `properties` arg is string -> values are expected to be an array to be unrolled (and raise error when they are scalar)

I think that will make the API more clean and predictable, because it is more explicit about assumptions and it avoids brittle value sniffing

proposals/load_geojson.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "load_geojson",
     "summary": "Converts GeoJSON into a vector data cube",
-    "description": "Converts GeoJSON data as defined by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946.html) into a vector data cube. Feature properties are preserved.",
+    "description": "Converts GeoJSON data as defined by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946.html) into a vector data cube. The feature properties provided in the parameter `properties` are preserved.",
```
soxofaan:
```suggestion
    "description": "Converts GeoJSON data as defined by [RFC 7946](https://www.rfc-editor.org/rfc/rfc7946.html) into a vector data cube. The feature properties provided in the parameter `properties` are used as cube values.",
```

proposals/load_geojson.json
m-mohr: I think I'm fine with the proposed way. Could you propose a wording or is the initial wording already covering that?

PR #551, opened 2025-07-17
`is_nan`: Return `false` instead of `true` for non-numerical data typ…

### LINKED ISSUE #486: is_nan: Restrict to numbers and add is_numerical
**Process ID:** is_nan / is_numerical

**Describe the issue:**
is_nan is a bit weird in that it also returns true for data types that are not numbers.

**Proposed solution:**
Make is_nan be compliant with most other languages and only return True if the value is actual a float NaN.

The existing behavior could be inversed and be part of a new process is_numerical (returns true for all integers and floating point numbers except for NaN, false for NaN and all non-numerical data types).

Any opinions? @soxofaan @clausmichele

m-mohr:
Closes #486 - See this issue for context.

Please debate whether we want to merge this or not. Just created this for the sake of getting a discussion going ;-)

### DIFF
--- a/is_nan.json
+++ b/is_nan.json
@@ -1,7 +1,7 @@
 {
     "id": "is_nan",
     "summary": "Value is not a number",
-    "description": "Checks whether the specified value `x` is *not* a number. Numbers are all integers and floating-point numbers, except for the special value `NaN` as defined by the [IEEE Standard 754](https://ieeexplore.ieee.org/document/4610935).",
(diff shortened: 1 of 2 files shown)


is_nan.json
```diff
@@ -1,56 +1,56 @@
...
+            },
+            "returns": false
+        },
+        {
+            "arguments": {
+                "x": null
+            },
+            "returns": false
+        }
+    ],
```
soxofaan, line 43: there is no example where `true` is returned now, but as far as I understand the structure of the examples, it's technically not possible to construct such an example (even with the existence of the `nan` process to produce a NaN value)

m-mohr: Indeed. I may remove these examples and add them to the tests (see #468) when we merged those.

soxofaan (approved):

ValentinaHutter (approved):

PR #562, opened 2025-12-19
Add DGGS proposal / implementation guide

m-mohr:
Description how DGGS could be implemented in openEO.
Feedback would be very welcome, especially from developers that have a good understanding of DGGS. (I'm pretty new to it.)

### DIFF
--- a/meta/dggs.md
+++ b/meta/dggs.md
@@ -1,0 +1,75 @@
+# DGGS in openEO
+
+## Introduction
+
+**DGGS** (Discrete Global Grid System) is a way to divide the Earth into a set of fixed, regular, and hierarchical **grid of cells** (often hexagons or squares). Each cell has a unique ID and can be refined into smaller cells for higher resolution. The grid covers the whole Earth - no gaps, no overlaps. Instead of continuous coordinates, space is divided into **cells**. Each cell represents an area and different DGGS implementations use different shapes, e.g. Hexagons or Squares. Every point on Earth belongs to exactly one cell (at a given resolution). Well-known DGGS implementations are H3, S2 and rHEALPix.
+
+## DGGS as a dimension
+
+openEO could integrate DGGS in the following ways:
+
+1. **Mapping to raster grids (x and y)**: Expose the traditional spatial dimensions with axes x and y to the user, hiding DGGS while mapping DGGS cells to raster pixels.
+   -  This feels most natural to users that are unaware of DGGS and familiar with raster data. The datacube would be a **raster data cube**.
+   -  It would allow for use of all existing openEO processes, including neighborhood operations such as `aggregate_spatial_window` and `apply_kernel`, provided that backends implement the necessary logic.
+   -  Hiding the specifics of DGGS also hides its advantages (e.g. global consistency, hierarchical aggregation, cell-based indexing). Operations based on the DGGS cells can't be applied.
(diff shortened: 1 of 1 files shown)


meta/dggs.md
```diff
@@ -0,0 +1,75 @@
...
+   -  Raster-based such as `apply_kernel`, `aggregate_spatial`, or `resample_spatial` processes can't be applied. Operations based on DGGS cells can be applied, e.g. `apply_dimension` and `reduce_dimension` along the geometry dimension.
+3. **New dimension type**: Specify a new dimension type within openEO that specifically handles DGGS grid cells.
+   -  This approach would be DGGS-native and provide all benefits of DGGS, but it will require the addition of processes that can handle the DGGS specific operations.
+   -  It would allow the definition of new processes that directly operate on DGGS concepts, such as parent/child aggregation, k-ring neighborhood operations, or resolution changes without reprojection. It also allows that processes directly require the DGGS dimension type in process parameters and return values, which otherwise is impossible.
+   -  Requires a new version of the openEO API and STAC datacube extension that adds the new dimension type. 
+   -  A set of new and adapted processes that explicitly support DGGS semantics must be specified.
+
```
j08lue:
```suggestion

## Key usage patterns to support

1. ...
```

I think the evaluation of options could be guided by the key usage patterns - if we can identify them. Like:

1. What are the most common tasks people perform with openEO? (Or GEE, for larger user base)?
2. Do these tasks require to operate on the original data at maximum precision (e.g. for conservation of original properties such as relationship between spectral bands - thinking atmospheric correction etc.)?
3. Do these tasks involve combining datasets (of different source projection)?
4. ...

m-mohr: Indeed, that's where I need input from the service providers. I only have a rough feeling as I don't have insight into the operational services.

jdries:
I had a look as well:
- for me option 3, creating new dimension type, makes most sense.
- It should be possible to retain backwards compatibility when adding DGGS, I hope that's implicitly understood
- Many raster operations indeed need a DGGS equivalent. The popular apply_neighborhood for instance, but I don't really see that as an issue.
- Also consider multi-resolution approaches, so processes that combine data at different levels, which is typically not allowed in the current spec.

The key point should be that DGGS really delivers an improvement. Fitting it into the existing concepts might make that hard to achieve.

meta/dggs.md
```diff
@@ -0,0 +1,75 @@
...
+It could make sense for backends to offer multiple ways (e.g. 1 and 2) to load a DGGS dataset, giving all options to a user:
+
+-  `load_collection` in general should load a data cube how it's defined by the collection metadata. 
+   -  If the collection metadata for a DGGS dataset exposes the dimensions *x, y, t, bands* for example, `load_collection` must return a raster datacube with exactly those dimensions.
+
+   -  If a backend decides to expose a DGGS dataset as *geometry, t, bands*, `load_collection` must return a vector datacube with exactly those dimensions.
+
+-  An alternative process could be defined that allows to load the dataset with different dimension types.
+
+Note that both `load_collection` and `load_dggs` would have to define cell inclusion rules, e.g. which cells to include and how to handle partial cells.
```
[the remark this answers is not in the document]
m-mohr: Removed the load_dggs for now.

m-mohr:
Merging this for now, I think all feedback has been incorporated where needed. A new PR will come up with processes. General issue to track all DGGS related issues and PRs:
https://github.com/Open-EO/openeo-api/issues/596

Issue #552, opened 2025-08-13
resample_spatial: new "binning" and "geocoding" resample method


jdries:
**resample_spatial:** 

Data like Sentinel-1, Sentinel-3 and Sentinel-5p is natively provided in sensor geometry. The transition to a rectangular grid with regularly spaced coordinates requires more advanced methods. 

For instance, in sentinel-3, one method is "binning":
https://github.com/senbox-org/snap-engine/blob/1500b8ad538b6f4c97a722d25087cdea85fed504/snap-binning/src/main/java/org/esa/snap/binning/operator/BinningOp.java

Another method in sentinel-1 is referred to as "geocoding".

The most naïve approach here is to define new processes, but here I'm coining the idea of adding a custom resample method to the existing resample_spatial method.

**Proposed solution:**

Extend the list of resample_methods in resample_spatial.

If this idea is considered valid, I expect to be able to better define correct names and descriptions in the coming months. ('binning' and 'geocoding' or relatively generic terms, I'm still figuring out if we can be more specific.)

The alternative of defining new processes is possible, but I feel like `resample_spatial` is a more clear expression of what is happening, also allowing to capitalize on existing handling of this process in clients and backends. (It has a fairly specific effect on the spatial dimensions of a datacube.)

PR #565, opened 2026-03-10
added geocode method for resample spatial in the processes json

suriyahgit:
Refer: 
https://github.com/Open-EO/openeo-processes-dask/pull/361
https://github.com/Open-EO/openeo-processes/issues/552

### DIFF
--- a/resample_spatial.json
+++ b/resample_spatial.json
@@ -1,7 +1,7 @@
 {
     "id": "resample_spatial",
     "summary": "Resample and warp the spatial dimensions",
-    "description": "Resamples the spatial dimensions (x,y) of the data cube to a specified resolution and/or warps the data cube to the target projection. At least `resolution` or `projection` must be specified.\n\nRelated processes:\n\n* Use ``filter_bbox()`` to set the target spatial extent.\n* To spatially align two data cubes with each other (e.g. for merging), better use the process ``resample_cube_spatial()``.",
+    "description": "Resamples the spatial dimensions (x,y) of the data cube to a specified resolution and/or warps the data cube to the target projection. At least `resolution` or `projection` must be specified.\n\nSome backends may additionally support `method=\"geocode\"` as a backend-specific extension. In this mode, the input data cube is rectified/geocoded using 2D geolocation layers such as `lon`/`lat` or `longitude`/`latitude`. These geolocation layers must be aligned with the spatial grid and may be provided as coordinates, data variables, or as dedicated bands in band-based data cubes. For `method=\"geocode\"`, a non-zero `resolution` is required. If `projection` is omitted, the backend defaults to EPSG:4326.\n\nRelated processes:\n\n* Use ``filter_bbox()`` to set the target spatial extent.\n* To spatially align two data cubes with each other (e.g. for merging), better use the process ``resample_cube_spatial()``.",
     "categories": [
         "cubes",
         "reproject"
@@ -26,7 +26,7 @@
         },
         {
             "name": "resolution",
-            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. Doesn't change the resolution by default (`0`).",
+            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. Doesn't change the resolution by default (`0`).\n\nFor `method=\"geocode\"`, the current backend implementation requires a single non-zero numeric resolution.",
             "schema": [
                 {
                     "description": "A single number used as the resolution for both x and y.",
@@ -49,7 +49,7 @@
         },
         {
             "name": "projection",
-            "description": "Warps the data cube to the target projection, specified as [EPSG code](https://spatialreference.org/ref/epsg/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html). By default (`null`), the projection is not changed.",
+            "description": "Warps the data cube to the target projection, specified as [EPSG code](https://spatialreference.org/ref/epsg/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html). By default (`null`), the projection is not changed.\n\nFor `method=\"geocode\"`, if no projection is specified, the current backend implementation defaults to EPSG:4326.",
             "schema": [
                 {
                     "title": "EPSG Code",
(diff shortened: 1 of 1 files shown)


resample_spatial.json
```diff
@@ -1,7 +1,7 @@
...
+    "description": "Resamples the spatial dimensions (x,y) of the data cube to a specified resolution and/or warps the data cube to the target projection. At least `resolution` or `projection` must be specified.\n\nSome backends may additionally support `method=\"geocode\"` as a backend-specific extension. In this mode, the input data cube is rectified/geocoded using 2D geolocation layers such as `lon`/`lat` or `longitude`/`latitude`. These geolocation layers must be aligned with the spatial grid and may be provided as coordinates, data variables, or as dedicated bands in band-based data cubes. For `method=\"geocode\"`, a non-zero `resolution` is required. If `projection` is omitted, the backend defaults to EPSG:4326.\n\nRelated processes:\n\n* Use ``filter_bbox()`` to set the target spatial extent.\n* To spatially align two data cubes with each other (e.g. for merging), better use the process ``resample_cube_spatial()``.",
```
clausmichele:
```suggestion
    "description": "Resamples the spatial dimensions (x,y) of the data cube to a specified resolution and/or warps the data cube to the target projection. At least `resolution` or `projection` must be specified.\n\nSome back-ends may additionally support `method=\"geocode\"` as a back-end-specific extension. In this mode, the input data cube is rectified/geocoded using 2D geolocation layers, which must be provided as two bands with names `longitude` and `latitude`. These bands are dropped in the resulting cube. For `method=\"geocode\"`, a non-zero `resolution` is required. If `projection` is omitted, the backend defaults to EPSG:4326.\n\nRelated processes:\n\n* Use ``filter_bbox()`` to set the target spatial extent.\n* To spatially align two data cubes with each other (e.g. for merging), better use the process ``resample_cube_spatial()``.",
```

@suriyahgit in openEO there's no concept of coordinates or data variables, we have dimensions (spatial, temporal, bands, other). I'm not sure which name we should adopt for the special band names, since the user can set the desired output projection probably `longitude` and `latitude` are not fitting all the scenarios and we should use something more generic if possible.

resample_spatial.json
```diff
@@ -26,7 +26,7 @@
         },
         {
             "name": "resolution",
-            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. Doesn't change the resolution by default (`0`).",
+            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. Doesn't change the resolution by default (`0`).\n\nFor `method=\"geocode\"`, the current backend implementation requires a single non-zero numeric resolution.",
```
clausmichele:
```suggestion
            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. Doesn't change the resolution by default (`0`).\n\nThe `method=\"geocode\"` requires a single non-zero numeric resolution.",
```

resample_spatial.json
```diff
@@ -49,7 +49,7 @@
         },
         {
             "name": "projection",
-            "description": "Warps the data cube to the target projection, specified as [EPSG code](https://spatialreference.org/ref/epsg/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html). By default (`null`), the projection is not changed.",
+            "description": "Warps the data cube to the target projection, specified as [EPSG code](https://spatialreference.org/ref/epsg/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html). By default (`null`), the projection is not changed.\n\nFor `method=\"geocode\"`, if no projection is specified, the current backend implementation defaults to EPSG:4326.",
```
clausmichele:
```suggestion
            "description": "Warps the data cube to the target projection, specified as [EPSG code](https://spatialreference.org/ref/epsg/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html). By default (`null`), the projection is not changed.\n\nFor `method=\"geocode\"`, if no projection is specified, the default is EPSG:4326.",
```

resample_spatial.json
```diff
@@ -100,7 +101,7 @@
         },
         {
             "name": "align",
-            "description": "Specifies to which corner of the spatial extent the new resampled data is aligned to.",
+            "description": "Specifies to which corner of the spatial extent the new resampled data is aligned to.\n\nNote: the current backend implementation does not apply `align` for `method=\"geocode\"`.",
```
clausmichele:
```suggestion
            "description": "Specifies to which corner of the spatial extent the new resampled data is aligned to.\n\nIt does not apply for `method=\"geocode\"`.",
```

clausmichele (commented):
@suriyahgit I proposed some changes. Let's see what @jdries and @m-mohr think about this PR.

resample_spatial.json
m-mohr: All methods are optional, as such this special handling in the process description doesn't really make sense. The specifics of the method should be described in description of the `method` parameter. 

I think no change should be made to the process description.

resample_spatial.json
m-mohr: This restriction should only be mentioned centrally in the method parameter.

resample_spatial.json
m-mohr: This restriction should only be mentioned centrally in the method parameter.

resample_spatial.json
m-mohr: This restriction should only be mentioned centrally in the method parameter.

m-mohr (changes requested):
As the methods enum is configurable and backends just choose this, it should be simple for implementors to remove it at a central place. Thus scattering descriptions for the methods in other parameters makes this hard to maintain. All changes needed for adding geocode should only be made in the method parameter schema/description ideally.

resample_spatial.json
suriyahgit: @clausmichele I used the terms coordinates and data variables because I was initially thinking of the dataset in terms of its representation as an xarray object, and also based on how datasets appear when loaded through a local connection in the openEO ecosystem. That said, I agree these terms are not strictly necessary here and can be removed.

suriyahgit:
reviews from @m-mohr accomodated - centralised the descriptions related to method="geocode" to the method description.

resample_spatial.json
```diff
@@ -75,14 +75,15 @@
...
...ne resampling\n* `geocode`: geocodes / rectifies data with 2D geolocation layers onto a regular target grid. This mode is intended for curvilinear or swath-like inputs where per-pixel geolocation is available. The current backend implementation requires a non-zero scalar `resolution`. If `projection` is omitted, the backend defaults to EPSG:4326. The `align` parameter is currently ignored for this method.\n* `lanczos`: Lanczos windowed sinc resampling\n* `max`: maximum resampling, selects the maximum value from all valid pixels\n* `med`: median resampling, selects the median value of all valid pixels\n* `min`: minimum resampling, selects the minimum value from all valid pixels\n* `mode`: mode resampling, selects the value which appears most often of all the sampled points\n* `near`: nearest neighbour resampling (default)\n* `q1`: first quartile resampling, selects the first quartile value of all valid pixels\n* `q3`: third quartile resampling, selects the third quartile value of all valid pixels\n* `rms`: root mean square (quadratic mean) of all valid pixels\n* `sum`: compute the weighted sum of all valid pixels\n\nValid pixels are determined based on the function ``is_valid()``.",
```
m-mohr, line 78:
```suggestion
            "description": "Resampling method to use. The following options are available and are meant to align with [`gdalwarp`](https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-r), except for `geocode`, which is a backend-specific geocoding/rectification mode:\n\n* `average`: average (mean) resampling, computes the weighted average of all valid pixels\n* `bilinear`: bilinear resampling\n* `cubic`: cubic resampling\n* `cubicspline`: cubic spline resampling\n* `geocode`: geocodes / rectifies data with 2D geolocation layers onto a regular target grid. This mode is intended for curvilinear or swath-like inputs where per-pixel geolocation is available. If `projection` is set to `null` or not provided, the backend reprojects to EPSG:4326.\n* `lanczos`: Lanczos windowed sinc resampling\n* `max`: maximum resampling, selects the maximum value from all valid pixels\n* `med`: median resampling, selects the median value of all valid pixels\n* `min`: minimum resampling, selects the minimum value from all valid pixels\n* `mode`: mode resampling, selects the value which appears most often of all the sampled points\n* `near`: nearest neighbour resampling (default)\n* `q1`: first quartile resampling, selects the first quartile value of all valid pixels\n* `q3`: third quartile resampling, selects the third quartile value of all valid pixels\n* `rms`: root mean square (quadratic mean) of all valid pixels\n* `sum`: compute the weighted sum of all valid pixels\n\nValid pixels are determined based on the function ``is_valid()``.",
```

Regarding: 

> The current backend implementation requires a non-zero scalar `resolution`. 

The current state of the backend implementation should not be reflected in the specification, which should be stable over time regardless of the implementation. 

Is this an implementation detail that will be solved later or is this generally difficult and we should just say:

> Requires a non-zero scalar `resolution`. 

The same applies to 

> The `align` parameter is currently ignored for this method.

where the "currently" is problematic. If a general implementation is possible, this should not be mentioned and can be adapted in the specific implementation and/or throw a corresponding error.

m-mohr:
@suriyahgit Is this still relevant / work in progress?

PR #569, opened 2026-03-26
chore: add aspect and slope definitions

dsamaey:


dev/.words
```diff
@@ -52,3 +52,5 @@ favor
 Undeploy
 queryable
 Queryables
+_aspect
+_slope
```
m-mohr: Use inline code fences instead of ' for these words, then you don't need to list them here in this file.

tests/aspect.json5
```diff
@@ -0,0 +1,6 @@
+{
+  "id": "aspect",
+  "level": "L2",
```
m-mohr: I feel like these processes don't belong in L2, should probably be L3 or so.

proposals/aspect.json
```diff
@@ -0,0 +1,27 @@
...
+            "name": "parameters",
+            "description": "A data cube with n bands containing elevation data",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
+            }
+        }
+    ],
+    "returns": {
+        "description": "A data cube with n bands named after the n bands in the input data cube, suffixed with ``'aspect'``. Each band contains the computed aspect values (in radians, from due North).  In case the surface is flat, the NaN value is used (since there is no direction of the gradient).",
```
m-mohr:
```suggestion
        "description": "A data cube with n bands named after the n bands in the input data cube, suffixed with `aspect`. Each band contains the computed aspect values (in radians, from due North).  In case the surface is flat, the NaN value is used (since there is no direction of the gradient).",
```

proposals/slope.json
```diff
@@ -0,0 +1,27 @@
...
+            "name": "parameters",
+            "description": "A data cube with n bands containing elevation data",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
+            }
+        }
+    ],
+    "returns": {
+        "description": "A data cube with n bands named after the n bands in the input data cube, suffixed with ``'_slope'``. Each band contains the computed slope values (in radians, relative to the horizontal plane).",
```
m-mohr:
```suggestion
        "description": "A data cube with n bands named after the n bands in the input data cube, suffixed with `_slope`. Each band contains the computed slope values (in radians, relative to the horizontal plane).",
```

proposals/aspect.json
```diff
@@ -0,0 +1,27 @@
...
+        "math"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "parameters",
+            "description": "A data cube with n bands containing elevation data",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
```
m-mohr, line 16: We should probably define the required dimension types here (x,y,bands?), potentially also in the return value. See other processes for details.

proposals/slope.json
```diff
@@ -0,0 +1,27 @@
...
+        "math"
+    ],
+    "experimental": true,
+    "parameters": [
+        {
+            "name": "parameters",
+            "description": "A data cube with n bands containing elevation data",
+            "schema": {
+                "type": "object",
+                "subtype": "datacube"
```
m-mohr, line 16: We should probably define the required dimension types here (x,y,bands?), potentially also in the return value. See other processes for details.

PR #571, opened 2026-06-18
Support for DGGS (new and existing processes)

m-mohr:
- Update existing processes to cater for DGGS dimensions.
- Add new processes

I'm unsure whether we should merge mask and mask_dggs...

Revert the package.json change once the package has a new release that supports the new dimension type.

Disclaimer: I have no major experience with DGGS, this may not make sense. The processes were written based on what AI explained me about DGGS vs raster data cubes. :-)


suriyahgit:
@m-mohr See also #572 which adds a filter_dggs proposal process as a further addition to the DGGS processes set - we have this implemented within the DEDLCube project

PR #576, opened 2026-09-11
`resample_spatial`: Clarify the behavior of the `resolution=0` 

### LINKED ISSUE #567: resample_spatial: resolution=0 preserves resolution?
from https://github.com/Open-EO/openeo-processes/blob/98ae0148287b6a59beaa9b00a506fdd6f725c098/resample_spatial.json#L28-L29

>  Doesn't change the resolution by default (0).

What does that practically mean? 
When changing projection in the general case (e.g. say from UTM to lonlat), you can not "preserve" the resolution in a meaningful way because the transformation between the pixel grids is fundamentally non-linear. 

It would indeed be possible to approximately preserve resolution in a small enough local area, but the current description does not reflect all the caveats and subtleties that are in play here.

e.g. do you aim to preserve approximate resolution in the center of the area of interest? Do you aim to preserve GSD in some sense? Do you aim to avoid up-sampling or down-sampling?

soxofaan:
see #567


resample_spatial.json
```diff
@@ -26,7 +26,7 @@
         },
         {
             "name": "resolution",
-            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. Doesn't change the resolution by default (`0`).",
+            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. When the resolution is `0` (the default), the back-end tries to minimize additional up- and down-sampling in a best-effort approach.",
```
m-mohr:
```suggestion
            "description": "Resamples the data cube to the target resolution, which can be specified either as separate values for x and y or as a single value for both axes. Specified in the units of the target projection. When the resolution is `0` (the default), additional up- and down-sampling is minimized on a best-effort basis.
",
```

Trying to avoid "the back-end" terminology.

m-mohr (approved):

